PR #24, opened 2023-03-31
Fixes typo in minikube instruction

bakineugene:
Looks like svc name is incorrect
`http-proxies-lb` works for me


savnadya:
Hi, I've merged your commit into our internal repository. It will appear in Github in about an hour. Thank you!

PR #25, opened 2023-04-04
Fix broken link to helm utility in English doc

alex-teplov:


gritukan:
Hi, I've merged your commit to internal repository and it will appear here in about an hour. Thank you for your fix!

PR #44, opened 2023-06-05
refactor(sdk/go): use any instead empty interface

sashamelentyev:
Empty interface says nothing. For this case in Go1.18 have any type. It's only cosmetic change. And code with any looks cleaner than interface{}


verytable:
Could, you, please provide some motivation behind this change.

yt/go/yt/internal/params.go
```diff
@@ -2027,13 +2027,13 @@ func (p *ReadTableParams) AccessTrackingOptions() **yt.AccessTrackingOptions {
 type StartOperationParams struct {
 	verb    Verb
 	opType  yt.OperationType
-	spec    interface{}
```
verytable, line 2030: This is a generated code and the generator should be fixed as well.

One should not modify this file manually and always use generator.

sashamelentyev:
@verytable empty interface says nothing. For this case in Go1.18 have `any` type. It's only cosmetic change. And code with any looks cleaner than interface{}

verytable:
I'm just advocating the cleanliness of the description.

yt/go/yt/internal/params.go
sashamelentyev: Remove changes from this file

verytable: Why not fix generator?)

verytable: We'd prefer to have all client code converted at once.

PR #46, opened 2023-06-06
chore(sdk/go): use xerrors from library

sashamelentyev:
YTsaurus have xerrors in library - why not use this instead external experimental package?


verytable:
Could you, please, add some motivation behind this change to description.

sashamelentyev:
YTsaurus have xerrors in library - why not use this instead external experimental package?

PR #47, opened 2023-06-07
feat(sdk/go): add init attributes for create node options

sashamelentyev:
Move check map initialization to CreateNodeOptions method for reuse in option functions


yt/go/yt/interface.go
```diff
@@ -125,6 +125,12 @@ type CreateNodeOptions struct {
 	*PrerequisiteOptions
 }
 
+func (options *CreateNodeOptions) initAttributes() {
```
verytable, line 128: NIT:  Use shorter receiver name like `o`.

PR #49, opened 2023-06-07
refactor(sdk/go): use any instead empty interface

sashamelentyev:
Use any instead empty interface


verytable:
Your commit is merged to internal repository. It will appear here after some time. PR will be closed automatically.

Thank you for the contribution!

PR #53, opened 2023-06-10
refactor(sdk/go): make rpcclient as export

sashamelentyev:
Constructor `NewClient` return unexported struct (*client) - it's bad practice, because I make client struct as exported.


verytable:
Actually, we did this on purpose. Client users do not need to know the implementation details and should not be able to modify or examine client fields.

As long as there are no concrete problems, we prefer to keep the interfaces private.

PR #54, opened 2023-06-11
refactor(sdk/go): refactor check slice

sashamelentyev:
Check slice length instead nil


verytable:
afaik the slices are serialized to protobuf messages differently in case of nil and empty slice.

The server code might distinguish between these two cases. No sure that it does, but prefer to leave it the ability.

PR #70, opened 2023-07-21
Allow to use write targets allocation via followers in proxying chunk service

gritukan:

### DIFF
--- a/yt/yt/server/node/cluster_node/proxying_chunk_service.cpp
+++ b/yt/yt/server/node/cluster_node/proxying_chunk_service.cpp
@@ -49,7 +49,8 @@ public:
     TProxyingChunkService(
         TCellId cellId,
         TProxyingChunkServiceConfigPtr serviceConfig,
-        TMasterConnectionConfigPtr connectionConfig,
+        TMasterConnectionConfigPtr masterConnectionConfig,
+        NApi::NNative::TConnectionDynamicConfigPtr connectionConfig,
         IChannelFactoryPtr channelFactory,
         IAuthenticatorPtr authenticator)
         : TServiceBase(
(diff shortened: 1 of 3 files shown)


yt/yt/server/node/cluster_node/proxying_chunk_service.cpp
```diff
@@ -270,20 +273,25 @@ class TProxyingChunkService
...
+        TAllocateWriteTargetsHandler(TProxyingChunkService* owner, bool useFollowers)
             : THandlerBase(owner)
+            , UseFollowers_(useFollowers)
         { }
 
     protected:
         TChunkServiceProxy::TReqAllocateWriteTargetsPtr CreateRequest() override
         {
-            return LeaderProxy_.AllocateWriteTargets();
+            auto proxy = UseFollowers_ ? LeaderProxy_ : FollowerProxy_;
```
ifsmirnov, line 284: Seems like it should be the other way.

yt/yt/server/node/cluster_node/proxying_chunk_service.h
```diff
@@ -15,7 +15,8 @@ namespace NYT::NClusterNode {
 NRpc::IServicePtr CreateProxyingChunkService(
     NElection::TCellId cellId,
     TProxyingChunkServiceConfigPtr serviceConfig,
-    NApi::NNative::TMasterConnectionConfigPtr connectionConfig,
+    NApi::NNative::TMasterConnectionConfigPtr masterConnectionConfig,
+    NApi::NNative::TConnectionDynamicConfigPtr connectionConfig,
```
ifsmirnov, line 19: How about `nativeConnectionConfig` for better distinction?

ifsmirnov (changes requested):

yt/yt/server/node/cluster_node/proxying_chunk_service.cpp
gritukan: Oops, you are right!

yt/yt/server/node/cluster_node/proxying_chunk_service.h
gritukan: Good idea!

ifsmirnov:
Resolved: https://github.com/ytsaurus/ytsaurus/commit/c8141a87f252754efefe2cb6910d38a5177082ce

PR #76, opened 2023-07-28
Primitive support of building with clang 16

amosbird:
More information is provided in each commit message.

P.S. I'm not sure if those cmake files related to bitcode generators should be changed in such way. It seems to be auto-generated by some tool (maybe ya.make).

### DIFF
--- a/conanfile.txt
+++ b/conanfile.txt
@@ -4,7 +4,7 @@ openssl/1.1.1s
 zlib/1.2.13
 
 [tool_requires]
-bison/3.5.3
+bison/3.8.2
 m4/1.4.19
 ragel/6.10
 yasm/1.3.0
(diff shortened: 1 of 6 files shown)


Kontakter:
Thank you for the attempt to support building code with clang-16.

We unfortunately cannot accept this pull request as is by several reasons:
  * cmake files are auto-generated and we should actually patch our inner build files to apply this changes properly
  * changes in various contrib libraries cannot be done directly – we should update these libraries or apply patches by some inner rules

But I will try to apply some of the changes from this PR to simplify the task of building YTsaurus with clang-16.

And I have a good news: our devtools team has a goal to support clang-16, so I hope that it will happen in near future.

amosbird:
No problem! Feel free to close this PR if it's useless :) 

BTW, perhaps it would be beneficial to consider open sourcing the build infrastructure?

Kontakter:
We have finally support clang-16!

Thank you for this review as a reference.

PR #95, opened 2023-09-05
Added retry checker to http retrying client

kaikash:

### DIFF
--- a/yt/yt/core/http/retrying_client.cpp
+++ b/yt/yt/core/http/retrying_client.cpp
@@ -1,5 +1,6 @@
 #include "config.h"
 
+#include "library/cpp/yt/logging/logger.h"
 #include "retrying_client.h"
 #include "private.h"
 #include "util/system/compiler.h"
@@ -23,8 +24,8 @@ class TResponseCheckerBase
     : public IResponseChecker
 {
 public:
-    TResponseCheckerBase(TRetryChecker retryChecker = BIND(&DefaultRetryChecker))
-        : RetryChecker_(std::move(retryChecker))
+    TResponseCheckerBase(TRetryChecker retryChecker = {})
+        : RetryChecker_(retryChecker ? std::move(retryChecker) : BIND(&DefaultRetryChecker))
     { }
 
     virtual bool IsRetriableError(const TError& error) override 
@@ -38,9 +39,7 @@ public:
 protected:
     static bool DefaultRetryChecker(const TError& error)
     {
-        // TODO: default retry checker
-        Y_UNUSED(error);
-        return false;
+        return !error.FindMatching(NRpc::EErrorCode::InvalidCredentials).has_value();
     }
 
 private:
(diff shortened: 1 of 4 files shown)


yt/yt/core/http/retrying_client.cpp
```diff
@@ -18,14 +19,46 @@ static const auto& Logger = HttpLogger;
...
+    }
+
+    virtual TError CheckError(const IResponsePtr& response) override = 0;
+    virtual NYTree::INodePtr GetFormattedResponse() const override = 0;
+
+protected:
+    static bool DefaultRetryChecker(const TError& error)
+    {
+        // TODO: default retry checker
+        Y_UNUSED(error);
```
Kontakter, line 42: We prefer to hide name in the method declaration, i.e:
```
static bool DefaultRetryChecker(const TError& /*error*/)
```

yt/yt/core/http/retrying_client.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include "retrying_client.h"
 #include "private.h"
+#include "util/system/compiler.h"
```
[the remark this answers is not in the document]
kaikash: my IDE accidentally added this, these includes are not needed

yt/yt/core/http/retrying_client.cpp
kaikash: This is actually not method declaration. There should be a default retry checking logic. Added in new commit.

yt/yt/core/http/retrying_client.cpp
```diff
@@ -166,49 +198,51 @@ class TRetryingClient
...
-            auto now = TInstant::Now();
-            if (now > deadline) {
+            if (!shouldRetry(checkError)) {
                 break;
             }
+
+            auto now = TInstant::Now();
             TDelayedExecutor::WaitForDuration(std::min(Config_->BackoffTimeout, deadline - now));
-        }
+        } while(true);
```
Kontakter, line 245: Cosmetic thing: we usually write infinite loops as
```
while (true) {
}
```

yt/yt/core/http/retrying_client.cpp
```diff
@@ -18,14 +18,44 @@ static const auto& Logger = HttpLogger;
...
+        return RetryChecker_(error);
+    }
+
+    virtual TError CheckError(const IResponsePtr& response) override = 0;
+    virtual NYTree::INodePtr GetFormattedResponse() const override = 0;
+
+protected:
+    static bool DefaultRetryChecker(const TError& error)
+    {
+        return !error.FindMatching(NRpc::EErrorCode::InvalidCredentials).has_value();
```
Kontakter, line 40: It looks a little bit weird in the base class, I suppose that here we should just `return true`

But in oauth_service.cpp we can specify checker that make retries for all errors expect InvalidCredentials

Kontakter:
Shipped!

Issue #89, opened 2023-08-31
Support GCC in C++ SDK


aleexf:
For now, compilation of C++ SDK using GCC fails. The errors I've faced with:
- Unsupported compiler options: `-fcolor-diagnostics`, `-faligned-allocation`, `-fdebug-default-version=4`, `-fuse-init-array`, `-Wimport-preprocessor-directive-pedantic`
- `TString` fails to construct from `TStringBuf` [here](https://github.com/ytsaurus/ytsaurus/blob/e487d705c9630064ab2949036a6dcd4478207c73/yt/cpp/mapreduce/interface/logging/yt_log.cpp#L102) and [here](https://github.com/ytsaurus/ytsaurus/blob/e487d705c9630064ab2949036a6dcd4478207c73/yt/yt/core/rpc/service_detail.cpp#L2253)
- `nullptr` cannot be assigned to `atomic<uintptr_t>` [here](https://github.com/ytsaurus/ytsaurus/blob/e487d705c9630064ab2949036a6dcd4478207c73/library/cpp/yt/memory/atomic_intrusive_ptr-inl.h#L27)

After resolving the above issues, the target `cpp-mapreduce-client` compiled successfully for me.

PR #96, opened 2023-09-06
Put attribute(weak) in a GCC-compatible position

aleexf:

In addition to #89, there is one more minor fix. GCC doesn't allow putting `__attribute__((weak))` before the `extern "C"` and produces non-weak symbols. Proof: https://godbolt.org/z/h48fYsGMW


Kontakter:
Shipped!

PR #98, opened 2023-09-13
Explicitly specify proto2 syntax to avoid proto compiler warnings

chegoryu:

This commit helps avoid these warnings:
```shell
[libprotobuf WARNING /root/.conan/data/protobuf/3.17.1/_/_/build/6434614c57c36a737c095eb9c5145f8048dd86aa/src/src/google/protobuf/compiler/parser.cc:651] No syntax specified for the proto file: yt/yt_proto/yt/formats/extension.proto. Please use 'syntax = "proto2";' or 'syntax = "proto3";' to specify a syntax version. (Defaulted to proto2 syntax.)
```

If you have more internal protos, I used this script + `find . -type f -name '*.proto'`:
```python
import sys

with open(sys.argv[1]) as f:
    files = f.readlines()

for file in files:
    file = file.strip()
    with open(file, "r") as f:
        content = f.read()

    assert content.count("syntax =") + content.count("syntax=") <= 1, file
    if "syntax = " not in content and "syntax=" not in content:
        print("fixing ", file)
        with open(file, "w") as f:
            f.write("syntax = \"proto2\";\n\n" + content)
```


chegoryu:
@gritukan ping

PR #115, opened 2023-10-24
Fix parse_bash_command_line function for some corner cases.

dasfex:
Current ```parse_bash_command_line``` function doesn't work correctly for some cases. F.e. ```"YT_USE_CLIENT_PROTOBUF=0 ./cppbinary"```. As a result you will get ```"", "YT_USE_CLIENT_PROTOBUF=0 ./cppbinary", ""```. Fix that behaviour.


Kontakter:
Shipped!

PR #119, opened 2023-10-27
Fix some typings in python package

chegoryu:

### DIFF
--- a/yt/python/yt/wrapper/schema/types.py
+++ b/yt/python/yt/wrapper/schema/types.py
@@ -37,6 +37,7 @@ else:
     def dataclass_transform():
         return lambda x: x
 
+T = typing.TypeVar("T")
 
 @dataclass_transform()
 def yt_dataclass(cls):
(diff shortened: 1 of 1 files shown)


yt/python/yt/wrapper/schema/types.py
```diff
@@ -82,7 +83,7 @@ def __repr__(self):
         return "Annotation({})".format(self._ti_type)
 
 
-def create_annotated_type(py_type, ti_type, to_yt_type=None, from_yt_type=None):
+def create_annotated_type(py_type: typing.Type[T], ti_type, to_yt_type=None, from_yt_type=None) -> typing.Type[T]:
```
chegoryu, line 86: Without this change pyright complains like this:

```bash
error: Variable not allowed in type expression (reportGeneralTypeIssues)
```

About `id` field in this code:

https://github.com/ytsaurus/ytsaurus/blob/e83a3bac40b1302991efb7de50ecc86b45c9bbf8/yt/python/examples/table_schema_typed/example.py#L8

https://github.com/ytsaurus/ytsaurus/blob/e83a3bac40b1302991efb7de50ecc86b45c9bbf8/yt/python/examples/table_schema_typed/example.py#L21-L25

This typing helps him to understand that return type of this function is a type here:

https://github.com/ytsaurus/ytsaurus/blob/b18c80b3294d355b66c167d36d642da1a76e4c98/yt/python/yt/wrapper/schema/types.py#L255-L272

PR #151, opened 2023-11-12
Respect "default_role_filter" config option for HTTP redirects

zlobober:

Without this change it is impossible to separate control and data proxies as operator sets "default_role_filter" to "default" and coordinator isses redirects to a hardcoded "data" role.

### DIFF
--- a/yt/yt/server/http_proxy/helpers.cpp
+++ b/yt/yt/server/http_proxy/helpers.cpp
@@ -357,7 +357,7 @@ void ProcessDebugHeaders(const IRequestPtr& /*request*/, const IResponseWriterPt
 
 void RedirectToDataProxy(const IRequestPtr& request, const IResponseWriterPtr& response, const TCoordinatorPtr& coordinator)
 {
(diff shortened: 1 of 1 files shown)


Kontakter:
I am not sure that this parameter is solely about default role for heavy proxies, it is about filtering proxies in /hosts request. I think the better idea is to make a separate option in config about redirection.

zlobober:
I do not see a distinction between the default role for heavy proxies and the role of proxies for redirecting clients trying to perform a heavy command at a control proxy. All SDKs by default pick a heavy proxy from "/hosts" without without specifying a certain role (unless the client knows exactly which role to use), and control proxy performs its redirect with the same reasoning.

So I do not see any circumstances in which these two values must be different.

Kontakter:
Ok, it sounds reasonable

Kontakter:
Shipped!

zlobober:
Can you please cherry-pick this commit to 23.1 and 23.2?

PR #152, opened 2023-11-12
Get rid of yt/yt/library/program dependency on yt/yt/library/containers.

zlobober:
The first one is designed to be an auxillary target for bootstrapping all
yt/yt/core-level singletons. At some moment a couple of container runtime related
singletons were introduced, imposing a redunadnt dependency on yt/yt/library/containers,
which is a quite heavy dependency (including Porto and CRI clients).

After YQL plugin started configuring YT singletons, it also broke YDB Piglet whitelist
around this commit: https://github.com/ydb-platform/ydb/commit/d28c55ab25cc8cedab8a5f4736c0d66e88b3da95

This commit moves container-related singletons to yt/yt/ytlib/program which keeps them available
in YT server processes and removes them from other users of yt/yt/library/program.

### DIFF
--- a/yt/yt/library/program/CMakeLists.txt
+++ b/yt/yt/library/program/CMakeLists.txt
@@ -18,7 +18,6 @@ target_link_libraries(yt-library-program PUBLIC
   yt-yt-core
   core-service_discovery-yp
   yt-library-monitoring
-  yt-library-containers
   library-profiling-solomon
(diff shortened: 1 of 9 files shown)


psushin:
Shipped!

psushin:
Shipped!

psushin:
Shipped!

PR #171, opened 2023-11-18
Fix redundant chunk_meta.proto import.

zlobober:
This fixes the following warning:
```
yt/ytlib/chunk_client/proto/chunk_service.proto:12:1: warning: Import yt_proto/yt/client/table_chunk_format/proto/chunk_meta.proto is unused.
```


savnadya:
Shipped!

PR #172, opened 2023-11-19
Prepare for the YQL query cancellation + minor refactorings

gritukan:


yt/yql/agent/yql_agent.cpp
```diff
@@ -110,15 +110,27 @@ class TYqlAgent
...
+        YT_LOG_INFO("Starting query (QueryId: %v, ImpersonationUser: %v)",
+            queryId,
+            impersonationUser);
 
         return BIND(&TYqlAgent::DoStartQuery, MakeStrong(this), queryId, impersonationUser, request)
             .AsyncVia(ThreadPool_->GetInvoker())
             .Run();
     }
 
+    void CancelQuery(TQueryId queryId, const TError& error) override
```
zlobober, line 127: Cancel? Может Abort?

yt/yt/server/query_tracker/alert_manager.h
```diff
@@ -15,14 +15,15 @@ class TAlertManager
     : public TRefCounted
 {
 public:
-    explicit TAlertManager(IInvokerPtr controlInvoker);
+    DEFINE_SIGNAL(void(std::vector<TError>*), PopulateAlerts);
+
+public:
+    TAlertManager(IInvokerPtr controlInvoker);
```
zlobober, line 21: explicit

yt/yt/server/query_tracker/public.h
```diff
@@ -0,0 +1,37 @@
+#include <yt/yt/client/query_tracker_client/public.h>
```
zlobober, line 1: Why did you move these to public.h? They are more private than public as they are not used outside this module.

zlobober: Well, it differs from my understanding. I believe that private.h and public.h are both for lightweight declarations such as forward declarations or logger declarations. Then, you put any declaration in private.h if it is possible (i.e. it is not used in external modules), and public.h otherwise. Finally, private.h includes public.h.

In your interpretation it is hard to distinguish private.h from public.h at all.

PR #180, opened 2023-11-21
Remove using transaction_id with FakeTransaction

valinurovdenis:
FakeTransaction doesn't have transaction_id and I've encountered such error in that [row](https://github.com/valinurovdenis/ytsaurus/blob/main/yt/python/yt/wrapper/heavy_commands.py#L376). Maybe it is worth passing None in case of FakeTransaction.

### DIFF
--- a/yt/python/yt/wrapper/heavy_commands.py
+++ b/yt/python/yt/wrapper/heavy_commands.py
@@ -16,7 +16,7 @@ from .format import YtFormatReadError
 
 import yt.logger as logger
 
-from yt.common import join_exceptions
+from yt.common import join_exceptions, YT_NULL_TRANSACTION_ID as null_transaction_id
 
 try:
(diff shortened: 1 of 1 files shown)


chizhonkova:
Hi! Let's also add transaction_id attribute with [null](https://github.com/ytsaurus/ytsaurus/blob/1c14965873fa54b387b0b2c0623349c1b1f08fb5/yt/python/yt/common.py#L48) value for FakeTransaction class .

valinurovdenis:
> Hi! Let's also add transaction_id attribute with [null](https://github.com/ytsaurus/ytsaurus/blob/1c14965873fa54b387b0b2c0623349c1b1f08fb5/yt/python/yt/common.py#L48) value for FakeTransaction class .

Done

chizhonkova:
Shipped!

PR #181, opened 2023-11-22
Always revoke tablet cell peer if corresponding cluster node is removed

gritukan:


Kontakter:
@gritukan – let answer to Ruslan and try to finish this PR

PR #198, opened 2023-11-28
Added build_master_snapshots for python wrapper

kaikash:

### DIFF
--- a/yt/python/yt/wrapper/admin_commands.py
+++ b/yt/python/yt/wrapper/admin_commands.py
@@ -46,7 +46,7 @@ def master_exit_read_only(retry=None, client=None):
     if retry is not None:
         params["retry"] = retry
 
-    return make_request("exit_read_only", params=params, client=client)
+    return make_request("master_exit_read_only", params=params, client=client)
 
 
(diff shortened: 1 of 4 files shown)


Kontakter:
I think, that it is a good idea to change calling this method here https://github.com/ytsaurus/ytsaurus/blob/main/yt/python/yt/wrapper/bin/yt-admin#L18-L19

Kontakter:
Just for a case – I mean to start using new method in yt-admin

Kontakter (changes requested):
See comment above

kaikash:
Sorry, I've messed up with this PR.

When I run updated `yt wrapper` it fails with error
```
Error: type object 'YtClient' has no attribute ...
Most likely, you need to run:
$ cd yt/python/yt/wrapper/bin/generate_client_impl
$ ./generate_client_impls.sh
```
but there is no `generate_client_impls.sh` script.

Also, do we want to add these commands to `yt_binary` as well?

Kontakter:
In your case you need to build `generate_client_impl` and run `./generate_client_impl ../../client_impl.py` from `yt/python/yt/wrapper/bin/generate_client_impl/`

Kontakter (approved):

PR #206, opened 2023-11-30
Use round-robin scheduling for node replication queues

gritukan:
Without round-robin replication may stuck in case if node replication queues are filled with chunks that are impossible to replicate (for example, erasure chunks in case of too few racks)


yt/yt/server/master/chunk_server/chunk_replicator.cpp
```diff
@@ -1844,12 +1857,24 @@ void TChunkReplicator::ScheduleReplicationJobs(IJobSchedulingContext* context)
...
+
+            if (!hasSpareReplicationResources() || std::ssize(chunksToReplicate) >= maxReplicationJobCount) {
+                break;
+            }
+
+            auto it = queue.PickRandomChunk();
+            auto chunkIdWithIndex = it->first;
             auto chunkId = chunkIdWithIndex.Id;
 
+            std::vector<TChunkId> allChunkIds;
```
maxim-babenko, line 1874: Looks like garbage

maxim-babenko (changes requested):

yt/yt/server/master/chunk_server/chunk_replicator.cpp
```diff
@@ -1727,22 +1729,29 @@ void TChunkReplicator::ScheduleJobs(EJobType jobType, IJobSchedulingContext* con
 
 void TChunkReplicator::ScheduleReplicationJobs(IJobSchedulingContext* context)
 {
+    // Keep the ability to disable scheduling completely.
+    if (GetDynamicConfig()->MaxMisscheduledReplicationJobsPerHeartbeat == 0) {
```
raundn, line 1733: This is a rather obscure way of doing this, but let's leave it as legacy for now. We can always add more dynconfig parameters later.

yt/yt/server/master/chunk_server/chunk_replicator.cpp
```diff
@@ -1844,17 +1862,21 @@ void TChunkReplicator::ScheduleReplicationJobs(IJobSchedulingContext* context)
     // Gather chunks for replica fetch.
     for (int priority = 0; priority < std::ssize(node->ChunkPushReplicationQueues()); ++priority) {
         auto& queue = node->ChunkPushReplicationQueues()[priority];
-        auto it = queue.begin();
-        while (it != queue.end() && hasSpareReplicationResources() && std::ssize(chunksToReplicate) < maxReplicationJobCount) {
-            auto jt = it++;
-            auto chunkIdWithIndex = jt->first;
-            auto chunkId = chunkIdWithIndex.Id;
+        // NB: We take chunks every queue to avoid starvation.
```
raundn, line 1865: ..._from_ every queue?

raundn:
Shipped!

nadya002:
Shipped!

PR #214, opened 2023-12-04
Fix wrong naming of Parquet

zlobober:
It is `parquet`, not `parquete`. Let's fix the naming until it is too late.


nadya002:
Shipped!

savnadya:
Shipped!

PR #229, opened 2023-12-07
Add python requests into peerdir for intergration tests

koct9i:
For example, without this change this command fails:
./ya make -A -L yt/yt/tests/integration/node


Kontakter:
Shipped!

PR #230, opened 2023-12-07
Allow to pass single SortColumn to build_schema_sorted_by

chegoryu:

It is already allowed to pass single `str`. This change helps to avoid confusing error.


denvr:
Shipped!

Issue #226, opened 2023-12-07
Tests with breakpoints do not work with CRI job environment


gritukan:
Jobs with breakpoint simply finish just after the start and breakpoint wait fails with timeout.

PR #232, opened 2023-12-07
Bind base directory for fs-events into user job container

koct9i:
Fix tests with brakepoints in CRI environment.

Fix for: https://github.com/ytsaurus/ytsaurus/issues/226

### DIFF
--- a/yt/yt/tests/library/yt_env_setup.py
+++ b/yt/yt/tests/library/yt_env_setup.py
@@ -415,7 +415,7 @@ class YTEnvSetup(object):
 
     @classmethod
     def create_yt_cluster_instance(cls, index, path):
-        modify_configs_func = functools.partial(cls.apply_config_patches, cluster_index=index)
+        modify_configs_func = functools.partial(cls.apply_config_patches, cluster_index=index, cluster_path=path)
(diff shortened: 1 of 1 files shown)


gritukan:
Let's add some simple test with breakpoint

yt/yt/tests/library/yt_env_setup.py
```diff
@@ -819,6 +819,13 @@ def apply_config_patches(cls, configs, ytserver_version, cluster_index):
 
             config["ref_counted_tracker_dump_period"] = 5000
 
+            # FIXME(khlebnikov) quick hack for "with_breakpoint".
+            config["exec_node"].setdefault("root_fs_binds", []).append({
```
gritukan, line 823: What is a non-hack solution here?

koct9i:
> Let's add some simple test with breakpoint

these are simple enough
https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/tests/integration/node/test_jobs.py#L74

yt/yt/tests/library/yt_env_setup.py
gritukan: I'd suggest changing FIXME to TODO with this idea.

yalekseev:
Shipped!

PR #233, opened 2023-12-07
Fix epic UB

gritukan:
Service discovery is enabled by default (see https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/core/service_discovery/yp/config.cpp#L10), so current implementation of the dummy service discovery gets into `Y_UNREACHABLE()`.


gritukan:
To be honest, I do not see a big issue here. Internal implemnetation of SD alredy returns nullptr in case if it is disabled, so code is ready for SD to be null.

Returning nullptr in case if SD is enabled does not see a big issue for me: SD config is used for now only in `CreateServiceDiscovery` and SD is accessed only via RpcDispatcher, so I can hardly imagine the code that expects ServiceDiscovery to be present iff it is enabled in config.

In any case, Y_UNREACHABLE() seems not to be a good idea in any case: it produces UB with bizarre errors in case if reached.

gritukan:
Yes, we can, but I still don't think that current change is dangerous. In the (unlikely) case if someone write the code that relies on a fact that SD is present if `config->Enabled`, it will simply crash in every execution of open source code and will be caught by any E2E open source test.

Another variant is to change Y_UNREACHABLE to YT_ABORT in this PR and modify operator so it disables SD in all components. However, I find it a bit strange that any YT-related code (like experiments) will not work with a default configuration having SD enabled, and one should disable it explicitly everywhere.

Kontakter:
Shipped!

PR #243, opened 2023-12-11
Support SetUserPassword & IssueToken & RevokeToken commands in go client

l0kix2:


yt/go/yt/interface.go
```diff
@@ -921,6 +927,35 @@ type AdminClient interface {
 		options *RemoveMemberOptions,
 	) (err error)
 
+	// http:verb:"set_user_password"
+	// http:params:"user","new_password_sha256","current_password_sha256"
+	SetUserPassword(
```
psushin, line 932: Why did you miss ListUserTokens method? It makes sense to support all auth-related methods together.

yt/go/yt/interface.go
```diff
@@ -921,6 +927,35 @@ type AdminClient interface {
 		options *RemoveMemberOptions,
 	) (err error)
 
+	// http:verb:"set_user_password"
+	// http:params:"user","new_password_sha256","current_password_sha256"
```
psushin, line 931: Instead of using AdminClient, it is better to create a new AuthClient for auth-related methods.

yt/go/yt/internal/yt-gen-client/yt-gen-client
psushin, line 1: Are you sure that you want to commit binary files?)

yt/go/yt/integration/admin_client_test.go
```diff
@@ -47,10 +54,44 @@ func (s *Suite) TestAddRemoveMember(t *testing.T, yc yt.Client) {
 	require.False(t, s.MemberOf(t, user, group))
 }
 
+func (s *Suite) TestSetUserPassword(t *testing.T, yc yt.Client) {
+	user := "user-" + guid.New().String()
+	_ = s.CreateUser(t, user)
+
+	passwordAttr := ypath.Path("//sys/users/" + user + "/@hashed_password")
```
psushin, line 61: Please, use `.Child(user).Attr("hashed_password")` to build ypath.

yt/go/yt/integration/admin_client_test.go
```diff
@@ -47,10 +54,44 @@ func (s *Suite) TestAddRemoveMember(t *testing.T, yc yt.Client) {
...
+
+	err = yc.RevokeToken(s.Ctx, user, "", token, nil)
+	require.NoError(t, err)
+}
+
 func (s *Suite) TestAddRemoveMaintenance(t *testing.T, yc yt.Client) {
 	t.Parallel()
 
-	ctx, cancel := context.WithTimeout(s.Ctx, time.Second*30)
+	ctx, cancel := context.WithTimeout(s.Ctx, testOpsTimeout)
```
psushin, line 94: Please, let's limit this PR to auth-realated methods only.

yt/go/yt/internal/encoder.go
```diff
@@ -244,6 +246,64 @@ func (e *Encoder) RemoveMember(
...
+}
+
+func (e *Encoder) IssueToken(
+	ctx context.Context,
+	user string,
+	password string,
+	options *yt.IssueTokenOptions,
+) (token string, err error) {
+	passwordSHA256 := ""
+	if password != "" {
```
psushin, line 276: Don't you want to put this if inside `encodeSHA256` func?

yt/go/yt/internal/rpcclient/encoder.go
```diff
@@ -1155,6 +1156,35 @@ func (e *Encoder) AddMember(
...
 }
 
+func (e *Encoder) SetUserPassword(
+	ctx context.Context,
+	user string,
+	newPassword string,
+	currentPassword string,
+	options *yt.SetUserPasswordOptions,
+) (err error) {
+	return xerrors.Errorf("Unimplemented method: SetUserPassword")
```
psushin, line 1166: Do you know, if these methods are implemented in RPC proxy API on the server side?

yt/go/yt/internal/rpcclient/encoder.go
```diff
@@ -1155,6 +1156,35 @@ func (e *Encoder) AddMember(
 	return
 }
 
+func (e *Encoder) SetUserPassword(
+	ctx context.Context,
+	user string,
+	newPassword string,
+	currentPassword string,
+	options *yt.SetUserPasswordOptions,
```
psushin, line 1164: for historical reasons we name parameter `opts' in rpc proxy implementation

psushin (dismissed):

yt/go/yt/interface.go
l0kix2: Well, I've only added the ones I needed, but I can add one more, I think.

yt/go/yt/interface.go
l0kix2: done

yt/go/yt/internal/encoder.go
l0kix2: Actually I don't think it would make code better, because 1) `encodeSHA256` just encodes some string and doesn't know is it password or empty string or something 2) `encodeSHA256` is used also for tokens in `RevokeToken` and shouldn't be empty 3) checking password/token emptiness here is an area of responsibility of api method business logic, not helper, i think.

yt/go/yt/internal/rpcclient/encoder.go
l0kix2: fixed

l0kix2: I don't, I've assumed that `Unimplemented` here means `not implemented in go client`.

l0kix2: But i suppose they [don't](https://github.com/l0kix2/ytsaurus/blob/3b44ede92c2644c9f0465908b512accb87527c5f/yt/yt/client/api/rpc_proxy/client_impl.cpp#L2041-L2074).

yt/go/yt/interface.go
l0kix2: done

l0kix2:
I've addressed all raised issues, @psushin could you please reivew again.

yt/go/yt/interface.go
```diff
@@ -37,6 +37,7 @@ import (
 	"io"
 
 	"go.ytsaurus.tech/library/go/core/xerrors"
+
```
verytable, line 40: Usually we group import into 3 groups
```
import (
    stdlib // os, fmt
    
    external libs // github.com, gopkg.in
    
    local ligs // go.ytsaurus.tech
)
```
with no empty lines in each group

yt/go/yt/interface.go
```diff
@@ -1501,6 +1510,47 @@ type GenerateTimestampOptions struct{}
 
 type GetInSyncReplicasOptions struct{}
 
+type AuthClient interface {
+
```
verytable, line 1514: Let's remove this empty line.

Usually we don't have those right in the beginning of any block (function body, interface, for loop body...)

yt/go/yt/internal/encoder.go
```diff
@@ -959,3 +1037,7 @@ func (e *Encoder) AlterQuery(
 	})
 	return
 }
+
+func EncodeSHA256(input string) string {
```
verytable, line 1041: NIT: Looks a bit strange to have this function exported.

Test could have it's own copy I assume.

yt/go/yt/integration/auth_client_test.go
```diff
@@ -0,0 +1,64 @@
+package integration
```
verytable, line 1: New files should be added to ya.make

l0kix2: done

yt/go/yt/interface.go
l0kix2: Makes sense, updated goland's setting to group `go.ytsaurus.tech` and fixed.

yt/go/yt/interface.go
l0kix2: done

yt/go/yt/internal/encoder.go
l0kix2: Yes, makes sense since we don't testing this function in the test. Done

l0kix2 (commented):
@verytable i've addressed all the issues, could you please look again.

Sorry for the long response.

verytable:
Ship

PR #250, opened 2023-12-13
Set environment variables USER and LOGNAME for CRI jobs

koct9i:

### DIFF
--- a/yt/yt/server/job_proxy/environment.cpp
+++ b/yt/yt/server/job_proxy/environment.cpp
@@ -10,6 +10,7 @@
 #include <yt/yt/library/containers/public.h>
 
 #include <util/system/fs.h>
+#include <util/system/user.h>
 
 #ifdef _linux_
 #include <yt/yt/library/containers/cgroup.h>
@@ -819,6 +820,12 @@ class TCriUserJobEnvironment
     : public IUserJobEnvironment
 {
 public:
+    TCriUserJobEnvironment() {
+	TString username = ::GetUsername();
+	Environment_.push_back("USER=" + username);
+	Environment_.push_back("LOGNAME=" + username);
+    }
+
     std::optional<TDuration> GetBlockIOWatchdogPeriod() const override
     {
         // No IO watchdog for simple job environment.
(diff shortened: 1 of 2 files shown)


koct9i:
YT CPP SDK fails if cannot get username 

erminate called after throwing an instance of 'yexception'
  what():  /root/.conan2/p/b/twix-3d94e104ee873/b/ytsaurus/yt/cpp/mapreduce/interface/config.cpp:286: Cannot get user name: (Success) /root/.conan2/p/b/twix-3d94e104ee873/b/ytsaurus/util/system/user.cpp:52:  getpwuid_r failed

But it could get from environment

https://github.com/ytsaurus/ytsaurus/blob/main/util/system/user.cpp#L15

There are two more-or-less sandard unix environment variables for that:
https://github.com/util-linux/util-linux/blob/master/login-utils/login.c#L1222

So, let's set them both.

This code-path which passes environment around (exec-node -> job-proxy -> yt-exec -> user-job) is ugly as hell. Let's leave cleaing this mess for future us.

yt/yt/server/job_proxy/environment.cpp
```diff
@@ -819,6 +820,12 @@ class TCriUserJobEnvironment
     : public IUserJobEnvironment
 {
 public:
+    TCriUserJobEnvironment() {
```
gritukan, line 823: { should be on the next line

yt/yt/server/job_proxy/environment.cpp
```diff
@@ -819,6 +820,12 @@ class TCriUserJobEnvironment
...
+	TString username = ::GetUsername();
```
gritukan, line 824: auto?

yt/yt/server/node/exec_node/job_environment.cpp
```diff
@@ -1095,6 +1096,17 @@ class TCriJobEnvironment
         }
         spec->Credentials.Gid = ::getgid();
 
+        if (!spec->Environment.contains("USER")) {
+            TString username;
```
gritukan, line 1100: I'm not sure if I got it correctly. Why isn't it sufficient just to add environment variables in TCriUserJobEnvironment::TCriUserJobEnvironment() like you did and we have to do it twice?

koct9i: Here we fill environemtn for job-proxy.
TCriUserJobEnvironment fills environement for user-job.

gritukan: Oh, I see.

gritukan (approved):

koct9i:
Merge conflict is resolved. Could be merged now.

PR #253, opened 2023-12-14
Deduplicate concurrent image pulls in CRI executor

koct9i:
CRI API does not prevent concurrent pulling same docker image.

### DIFF
--- a/yt/yt/library/containers/cri/cri_executor.cpp
+++ b/yt/yt/library/containers/cri/cri_executor.cpp
@@ -530,10 +530,35 @@ public:
             return GetImageStatus(image)
                 .Apply(BIND([=, this, this_ = MakeStrong(this)] (const TCriImageApi::TRspImageStatusPtr& imageStatus) {
                     if (imageStatus->has_image()) {
-                        return MakeFuture(TCriImageDescriptor{.Image = imageStatus->image().id()});
+                        const auto& imageId = imageStatus->image().id();
+                        YT_LOG_DEBUG("Docker image found in cache (Image: %v, ImageId: %v)", image, imageId);
+                        return MakeFuture(TCriImageDescriptor{.Image = imageId});
                     }
                     return PullImage(image, /*always*/ true, authConfig, podSpec);
                 }));
+        }
+
+        auto pullComplete = NewPromise<void>();
+        {
+            auto guard = Guard(SpinLock_);
+            if (auto* pullFuture = InflightImagePulls_.FindPtr(image.Image)) {
+                YT_LOG_DEBUG("Waiting for in-flight docker image pull (Image: %v)", image);
+                return pullFuture->Apply(BIND([=, this, this_ = MakeStrong(this)] (const TError&) {
+                    // Ignore errors and retry pull after end of previous concurrent attempt.
+                    return PullImage(image, /*always*/ false, authConfig, podSpec);
+                }));
+            }
+
+            // Future for in-flight pull should not be canceled by waiters,
+            // but could become canceled when original pull is canceled.
+            auto pullFuture = pullComplete.ToFuture().ToUncancelable();
+            EmplaceOrCrash(InflightImagePulls_, image.Image, pullFuture);
+            pullFuture.Subscribe(BIND([=, this, weakThis = MakeWeak(this)] (const TError&) {
+                if (auto this_ = weakThis.Lock()) {
+                    auto guard = Guard(SpinLock_);
+                    InflightImagePulls_.erase(image.Image);
+                }
+            }));
         }
 
         auto req = ImageApi_.PullImage();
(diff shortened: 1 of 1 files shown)


yt/yt/library/containers/cri/cri_executor.cpp
```diff
@@ -544,6 +544,25 @@ class TCriExecutor
...
+            auto guard = Guard(SpinLock_);
+            if (auto* pullFuture = InflightImagePulls_.FindPtr(image.Image)) {
+                return pullFuture->Apply(BIND([=, this, this_ = MakeStrong(this)] {
+                    return PullImage(image, /*always*/ false, authConfig, podSpec);
+                }));
+            }
+
+            auto pullFuture = pullComplete.ToFuture().ToUncancelable();
+            EmplaceOrCrash(InflightImagePulls_, image.Image, pullFuture);
+            pullFuture.Subscribe(BIND([=, this, weakThis = MakeWeak(this)] (const TError &) {
```
gritukan, line 558: Here `const TError&` is not required too

yt/yt/library/containers/cri/cri_executor.cpp
```diff
@@ -552,7 +571,11 @@ class TCriExecutor
         if (podSpec) {
             FillPodSandboxConfig(req->mutable_sandbox_config(), *podSpec);
         }
-        return req->Invoke().Apply(BIND([] (const TCriImageApi::TRspPullImagePtr& rsp) -> TCriImageDescriptor {
+
+        auto pullFuture = req->Invoke();
+        pullComplete.SetFrom(pullFuture);
+
+        return pullFuture.Apply(BIND([] (const TCriImageApi::TRspPullImagePtr& rsp) -> TCriImageDescriptor {
```
gritukan, line 578: In case if this future will be cancelled, all subscribers will be cancelled too. So, it's better to do
```
auto pullFuture = req->Invoke().ToUncancellable();
```

and after that ToUncancellable() above will not be required.

yt/yt/library/containers/cri/cri_executor.cpp
```diff
@@ -544,6 +544,25 @@ class TCriExecutor
                 }));
         }
 
+        auto pullComplete = NewPromise<void>();
+        {
+            auto guard = Guard(SpinLock_);
+            if (auto* pullFuture = InflightImagePulls_.FindPtr(image.Image)) {
+                return pullFuture->Apply(BIND([=, this, this_ = MakeStrong(this)] (const TError &) {
```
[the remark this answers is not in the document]
koct9i: Actually this seems required here.
Without argument apply will short-cut error?
But this code want to retry pull in any case.

yt/yt/library/containers/cri/cri_executor.cpp
```diff
@@ -544,6 +544,25 @@ class TCriExecutor
...
+                    return PullImage(image, /*always*/ false, authConfig, podSpec);
+                }));
+            }
+
+            auto pullFuture = pullComplete.ToFuture();
```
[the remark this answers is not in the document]
koct9i: Nope. Original pull operaion must stay cancellable.
After cancel shared future will expose error and let anybody to retry pull.

yt/yt/library/containers/cri/cri_executor.cpp
koct9i: I prefer to leave it here to express that it is called even in case of error.

yt/yt/library/containers/cri/cri_executor.cpp
koct9i: This one must stay cancellable.

yt/yt/library/containers/cri/cri_executor.cpp
gritukan: Let's log the error in this case. And `const TError &` -> `const TError&`. And let's do the same for the snippet above.

yt/yt/library/containers/cri/cri_executor.cpp
gritukan: Why is it better to retry pull in case of cancellation instead of ignoring it?

koct9i: Pulls are independent. Their callers are independent. They must break indepdently.
First pull could stuck and fail becuse of network issues.

Second and following pulls may waits optimiscaly to not generate redundant workload and result for free, but they needs this image too -- for more robust process next attermp they should start pulling from scratch.

I don't see reason to optimize weird case when first operation cancels pull but there are following operaions which could inherit it. Anyway, if the are no following operations pull must be canceled.

gritukan: Olay, that's reasonable!

yt/yt/library/containers/cri/cri_executor.cpp
koct9i: This seems redundant. I've added logging of original docker pull error. Chained pulls will log their progress after that, and all messages linked by same image name.

gritukan (approved):

k-pogorelov (approved):
LGTM.

PR #259, opened 2023-12-15
Catch and wrap only ImportError exception in spyt when cannot import yt client

koct9i:


Alexvsalexvsalex:
What's a purpose?

koct9i:
> What's a purpose?

Do not confuse user with wrong diagnostics. Present code catches ALL exceptions and suggest "Please install ytsaurus-client" while error reason is not related to missing library. For example when yt client is installer but isn't happy about present config in environment variables.

Alexvsalexvsalex:
Shipped!

PR #260, opened 2023-12-15
go-sdk: add concatenate nodes call

holycheater:

### DIFF
--- a/yt/go/yt/integration/cypress_client_test.go
+++ b/yt/go/yt/integration/cypress_client_test.go
@@ -27,7 +27,7 @@ func TestCypressClient(t *testing.T) {
 		{Name: "CopyNode", Test: suite.TestCopyNode},
 		{Name: "MoveNode", Test: suite.TestMoveNode},
 		{Name: "LinkNode", Test: suite.TestLinkNode},
-		{Name: "ConcatNode", Test: suite.TestConcatNode},
+		{Name: "Concatenate", Test: suite.TestConcatenate},
 		{Name: "CreateObject", Test: suite.TestCreateObject},
 		{Name: "BinaryPath", Test: suite.TestBinaryPath},
 	})
@@ -263,7 +263,7 @@ func (s *Suite) TestLinkNode(t *testing.T, yc yt.Client) {
 	require.Equal(t, yt.NodeLink, typ)
 }
 
-func (s *Suite) TestConcatNode(t *testing.T, yc yt.Client) {
+func (s *Suite) TestConcatenate(t *testing.T, yc yt.Client) {
 	t.Parallel()
 	var err error
 
@@ -281,7 +281,7 @@ func (s *Suite) TestConcatNode(t *testing.T, yc yt.Client) {
 	_, err = yc.CreateNode(ctx, src2, yt.NodeTable, nil)
 	require.NoError(t, err)
 
-	err = yc.ConcatNode(ctx, []ypath.YPath{src1, src2}, dst, nil)
+	err = yc.Concatenate(ctx, []ypath.YPath{src1, src2}, dst, nil)
 	require.NoError(t, err)
 }
 
--- a/yt/go/yt/integration/cypress_client_test.go
+++ b/yt/go/yt/integration/cypress_client_test.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/require"
+
 	"go.ytsaurus.tech/yt/go/guid"
 	"go.ytsaurus.tech/yt/go/ypath"
 	"go.ytsaurus.tech/yt/go/yt"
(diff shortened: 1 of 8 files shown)


yt/go/yt/integration/cypress_client_test.go
```diff
@@ -261,6 +263,28 @@ func (s *Suite) TestLinkNode(t *testing.T, yc yt.Client) {
...
+	dst := tmpPath()
+
+	_, err = yc.CreateNode(ctx, dst, yt.NodeTable, nil)
+	require.NoError(t, err)
+	_, err = yc.CreateNode(ctx, src1, yt.NodeTable, nil)
+	require.NoError(t, err)
+	_, err = yc.CreateNode(ctx, src2, yt.NodeTable, nil)
+	require.NoError(t, err)
+
+	err = yc.ConcatNode(ctx, []ypath.YPath{src1, src2}, dst, nil)
```
achulkov2, line 284: I think it is best to keep the name consistent across different sdks, it is called Concatenate in C++ and yt concatenate in cli/python, so probably `yc.Concatenate` is better

yt/go/yt/integration/cypress_client_test.go
```diff
@@ -261,6 +263,28 @@ func (s *Suite) TestLinkNode(t *testing.T, yc yt.Client) {
...
+	require.NoError(t, err)
```
achulkov2, line 285: Maybe the test should have a few rows in the source tables and also check that the contents match?

yt/go/yt/integration/cypress_client_test.go
holycheater: fixed method name

yt/go/yt/integration/cypress_client_test.go
holycheater: I tried, but it does not seem possible atm: integration test calls both rpc-client and http-client. rpc-client does not implement WriteTable :|

holycheater: okay, seems like it's possible to skip RPC tests, added one

yt/go/yt/internal/rpcclient/encoder.go
```diff
@@ -442,6 +442,30 @@ func (e *Encoder) LinkNode(
...
+	options *yt.ConcatenateOptions,
+) (err error) {
+	srcs := make([]string, 0, len(src))
+	for _, s := range src {
+		srcs = append(srcs, s.YPath().String())
+	}
+	req := &rpc_proxy.TReqConcatenateNodes{
+		SrcPaths:             srcs,
+		DstPath:              ptr.String(dst.YPath().String()),
+		Fetcher:              nil,
```
psushin, line 458: Why did you decide to ignore `Fetcher` parameter?

yt/go/yt/internal/rpcclient/encoder.go
```diff
@@ -442,6 +442,30 @@ func (e *Encoder) LinkNode(
...
+		TransactionalOptions: convertTransactionOptions(options.TransactionOptions),
```
psushin, line 459: You should anticipate, that the options could be nil.

yt/go/yt/internal/rpcclient/request.go
```diff
@@ -2,6 +2,7 @@ package rpcclient
 
 import (
 	"go.ytsaurus.tech/library/go/core/log"
+
```
psushin, line 5: Let's remove it.

yt/go/yt/internal/rpcclient/request.go
```diff
@@ -367,6 +368,33 @@ func (r *LinkNodeRequest) SetRetry(retry bool) {
...
+	*rpc_proxy.TReqConcatenateNodes
+}
+
+func NewConcatenateRequest(r *rpc_proxy.TReqConcatenateNodes) *ConcatenateRequest {
+	return &ConcatenateRequest{TReqConcatenateNodes: r}
+}
+
+func (r *ConcatenateRequest) Log() []log.Field {
+	return []log.Field{
+		log.String("target", r.GetDstPath()),
```
psushin, line 381: target -> destination
source -> sources

yt/go/yt/integration/cypress_client_test.go
psushin: you could use `yttest.Env.YT` from `Suite` - this is generally available http client
with it you can always do `write_table` via http, and not skip tests for RPC implementation

yt/go/yt/integration/cypress_test.go
```diff
@@ -213,4 +213,21 @@ func TestCypress(t *testing.T) {
 			require.Contains(t, err.Error(), "привет")
 		})
 	})
+
+	t.Run("Concatenate", func(t *testing.T) {
```
psushin, line 217: What does this test actually test?

psushin (changes requested):

Issue #280, opened 2023-12-21
Race in makedirs on python test cleanup that causes the test to fail (looks like arcadia_interop.py does not handle parallel tests)


chegoryu:
When I try to run all tests from `yt/python/yt/wrapper/tests/py3_only` (just `ya make -A .`) some of tests fail with this error:

```python
yt/python/yt/wrapper/tests/py3_only <py3test> [size:large] [tags: ya:external, ya:fat, ya:force_sandbox, ya:full_logs, ya:noretries, ya:yt, ya:ytexec]
------ sole chunk ran 181 tests (total:415.97s - test:415.93s)
[fail] test_parquet.py::TestParquet::test_dump_parquet[native_v4] (nadya02) [default-linux-x86_64-debug] (1.09s)
setup failed:
yt/python/yt/wrapper/testlib/conftest_helpers.py:69: in <lambda>
    request.addfinalizer(lambda: environment.cleanup())
yt/python/yt/testlib/test_environment.py:280: in cleanup
    self.save_sandbox()
yt/python/yt/testlib/test_environment.py:284: in save_sandbox
    arcadia_interop.save_sandbox(self.sandbox_dir, self.uniq_dir_name)
yt/python/yt/environment/arcadia_interop/arcadia_interop.py:310: in save_sandbox
    os.makedirs(output_path_dir)
contrib/tools/python3/src/Lib/os.py:225: in makedirs
    mkdir(name, mode)
E   FileExistsError: [Errno 17] File exists: '/home/echunaev/git/ytsaurus_v2/yt/python/yt/wrapper/tests/py3_only/test-results/py3test/testing_out_stuff/TestYtWrapperV4'
Log: /home/echunaev/git/ytsaurus_v2/yt/python/yt/wrapper/tests/py3_only/test-results/py3test/testing_out_stuff/test_parquet.py.TestParquet.test_dump_parquet.native_v4.(nadya02).log
Logsdir: /home/echunaev/git/ytsaurus_v2/yt/python/yt/wrapper/tests/py3_only/test-results/py3test/testing_out_stuff
[fail] test_arrow.py::TestArrow::test_arrow_reader[v4] (nadya02) [default-linux-x86_64-debug] (1.07s)
setup failed:
yt/python/yt/wrapper/testlib/conftest_helpers.py:69: in <lambda>
    request.addfinalizer(lambda: environment.cleanup())
yt/python/yt/testlib/test_environment.py:280: in cleanup
[...]

PR #281, opened 2023-12-21
Fix mkdirs race on test cleanup

chegoryu:

Fixes: https://github.com/ytsaurus/ytsaurus/issues/280

With this fix `ya make -A .` didn't fail on any test in `yt/python/yt/wrapper/tests/py3_only` (I checked it twice).


Kontakter:
Shipped!

PR #288, opened 2023-12-27
Better logging for CRI process operations

koct9i:


Kontakter:
I think it is better to create `Logger` field inside TCriProcess with `Pod` and `Container` tags

Kontakter:
@koct9i – there is some conflicts, resolve it, please and after that I will import this pull request to Arcadia

Issue #248, opened 2023-12-13
Expose cluster connection through an authentication-free http handler


zlobober:
We have various HTTP handlers in HTTP proxy; the most important of them are in /api/... tree and they are auto-generated basing on YT driver commands which has a declarative description of YT API commands. All these commands require credentials  of some sort (i.e. token in "Authorization: OAuth ..." scheme or cookies).

The rest of handlers are registered around here: 
https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/server/http_proxy/bootstrap.cpp#L431-L465

These handlers mostly do not require authentication.

We'd like to get a new http handler `/cluster_connection`, which, being invoked, returns the so-called cluster connection. After this change, it becomes easy to identify a remote cluster by simply providing its HTTP endpoint (a balancer above cluster endpoints, to be precise), and this will be convenient e.g. in k8s operator for defining remote node groups (todo: link with corresponding issue).

Cluster connection is a YSON document describe the configuration of a native client to YT. You may have a look on real-life cluster connection by invoking `yt get //sys/@cluster_connection`. The handler must perform exactly this get request via master cache, and produce YSON result (take "/service" handler as an example).

You may then write an integration test around test_http_proxy.py which makes a direct HTTP call to "/cluster_connection" and compares the result to the value of "//sys/@cluster_connection".

zlobober:
Consider this issue assigned to @Krock21rus (invitation to our organization is pending).

PR #291, opened 2023-12-28
Add /cluster_connection unauthenticated API #248

Krock21rus:
#248 


yt/yt/server/http_proxy/coordinator.cpp
```diff
@@ -522,6 +522,34 @@ void THostsHandler::HandleRequest(
...
+{
+    auto path = req->GetUrl().Path;
+    if (path != "/cluster_connection" && path != "/cluster_connection/") {
+        rsp->SetStatus(EStatusCode::NotFound);
+        WaitFor(rsp->Close())
+            .ThrowOnError();
+        return;
+    }
+
+    auto nodeYson = WaitFor(Client_->GetNode("//sys/@cluster_connection"))
```
gritukan, line 544: Let's add here an option to read from master cache. By default, it will be a follower.

yt/yt/server/http_proxy/coordinator.cpp
```diff
@@ -522,6 +522,34 @@ void THostsHandler::HandleRequest(
...
+
+TClusterConnectionHandler::TClusterConnectionHandler(NApi::IClientPtr client)
+    : Client_(std::move(client))
+{ }
+
+void TClusterConnectionHandler::HandleRequest(
+    const NHttp::IRequestPtr& req,
+    const NHttp::IResponseWriterPtr& rsp)
+{
+    auto path = req->GetUrl().Path;
```
gritukan, line 536: const auto&?

yt/yt/tests/integration/proxies/test_http_proxy.py
```diff
@@ -153,6 +153,20 @@ def check_role_updated():
...
 
+    @authors("aleksandr.gaev")
+    def test_cluster_connection(self):
+        def get_cluster_connection(path):
+            return requests.get(self._get_proxy_address() + path)
+
+        url = "{}/api/v3/get?path=//sys/@cluster_connection".format(self._get_proxy_address())
+        cluster_connection = requests.get(url).content
+        assert len(cluster_connection) > 0
+
```
gritukan, line 164: Can we also do `get('//sys/@cluster_connection')` to check that new handler actually returns cluster connection and not something strange?

Krock21rus:
We discussed and agreed on text json format(Just like `/api/v3/get?path=//sys/@cluster_connection` behaves). yson or binary formats can be added later

gritukan (approved):
LGTM

yt/yt/server/http_proxy/coordinator.cpp
```diff
@@ -522,6 +522,36 @@ void THostsHandler::HandleRequest(
...
+    const auto& path = req->GetUrl().Path;
+    if (path != "/cluster_connection" && path != "/cluster_connection/") {
+        rsp->SetStatus(EStatusCode::NotFound);
+        WaitFor(rsp->Close())
+            .ThrowOnError();
+        return;
+    }
+
+    TGetNodeOptions options;
+    options.ReadFrom = EMasterChannelKind::Cache;
```
Kontakter, line 545: It is a little bit preferrable to write:
```
auto options = TGetNodeOptions{.ReadFrom = EMasterChannelKind::Cache};
```

Krock21rus: This code does not compile :( 

```
/home/ytsaurus/yt/yt/server/http_proxy/coordinator.cpp:544:37: error: field designator 'ReadFrom' does not refer to any field in type 'TGetNodeOptions'
    auto options = TGetNodeOptions{.ReadFrom = EMasterChannelKind::Cache};
```

Looks like some C++ limitation

We have
```
struct TGetNodeOptions : public TMasterReadOptions

struct TMasterReadOptions {
    EMasterChannelKind ReadFrom;
}
```

And only
```
TMasterReadOptions{.ReadFrom = EMasterChannelKind::Cache}
```
works. But not in this case

Kontakter:
Shipped!

PR #292, opened 2023-12-28
Set unique attempt for each cri contianer create

koct9i:
- Set unique attempt for each CRI container create

### DIFF
--- a/yt/yt/library/containers/cri/cri_executor.cpp
+++ b/yt/yt/library/containers/cri/cri_executor.cpp
@@ -207,6 +207,7 @@ public:
         : Config_(std::move(config))
         , RuntimeApi_(CreateRetryingChannel(Config_, channelFactory->CreateChannel(Config_->RuntimeEndpoint)))
         , ImageApi_(CreateRetryingChannel(Config_, channelFactory->CreateChannel(Config_->ImageEndpoint)))
+        , Attempt_(RandomNumber<ui32>())
     { }
 
     TString GetPodCgroup(TString podName) const override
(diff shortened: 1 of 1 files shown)


yt/yt/library/containers/cri/cri_executor.cpp
```diff
@@ -351,6 +352,9 @@ class TCriExecutor
         {
             auto* metadata = config->mutable_metadata();
             metadata->set_name(ctSpec->Name);
+
+            // Set unique attempt for each newly created container.
+            metadata->set_attempt(Attempt_++);
```
Kontakter, line 357: What this attempt field actually mean in CRI?

For me attempt notion is usually used in retries. But here we have CriExecutor that is responsible for all requests to CRI

koct9i: "Attempt" is designed to be used for retries for sure.
But it's just a number appened to container name.

https://github.com/containerd/containerd/blob/main/pkg/cri/server/helpers.go#L128

Unique key for create is (namespace, pod, name, attempt)
it returns container id and all futher operations are done by id.

We could add randomness to container name with same result.
At the end of each job CRI executor list sand removes all contianers in the pod thus garbage should be collected.

gritukan:
We usually start commit name from a capital letter, so "set" -> "Set".

Besides this, LGTM.

gritukan (approved):

yalekseev:
Shipped!

yt/yt/library/containers/cri/cri_executor.cpp
Kontakter: I would prefer to add randomness to the name, because the current solution is misleading.

PR #293, opened 2023-12-28
Fix race between TCriProcess spawn and kill

koct9i:

### DIFF
--- a/yt/yt/library/containers/cri/cri_executor.cpp
+++ b/yt/yt/library/containers/cri/cri_executor.cpp
@@ -97,6 +97,15 @@ public:
 
     void Kill(int /*signal*/) override
     {
+        if (Finished_) {
+            return;
+        }
+        Finished_ = true;
+
+        if (!Started_) {
+            THROW_ERROR_EXCEPTION("Process is not started yet");
+        }
+
         YT_LOG_DEBUG("Killing process");
         WaitFor(Executor_->StopContainer(ContainerDescriptor_))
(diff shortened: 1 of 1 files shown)


yt/yt/library/containers/cri/cri_executor.cpp
```diff
@@ -96,6 +96,15 @@ class TCriProcess
 
     void Kill(int /*signal*/) override
     {
+        if (Finished_) {
```
Kontakter, line 99: I don't see this field in this class.

Are you sure, that this patch is working?

koct9i: That protected field in parrent class TProcessBase

https://github.com/ytsaurus/ytsaurus/blob/2744cdef14773f33eb7db1de465c56a99bd1eae6/yt/yt/library/process/process.h#L57

gritukan, line 99: nip: this can be done shorter via std::exchange

yt/yt/library/containers/cri/cri_executor.cpp
```diff
@@ -149,6 +158,13 @@ class TCriProcess
             ContainerDescriptor_,
             ContainerSpec_->Environment);
 
+        YT_VERIFY(!Started_);
```
gritukan, line 161: And here

gritukan (approved):

Kontakter:
Shipped!

PR #296, opened 2023-12-29
Retry CRI requests when containerd is not initialized

koct9i:
Containerd does not report correct error code in this case yet.
Thus we have to check error message.

Link: https://github.com/containerd/containerd/pull/9565

### DIFF
--- a/yt/yt/library/containers/cri/cri_executor.cpp
+++ b/yt/yt/library/containers/cri/cri_executor.cpp
@@ -221,8 +221,8 @@ public:
         TCriExecutorConfigPtr config,
         IChannelFactoryPtr channelFactory)
         : Config_(std::move(config))
-        , RuntimeApi_(CreateRetryingChannel(Config_, channelFactory->CreateChannel(Config_->RuntimeEndpoint)))
-        , ImageApi_(CreateRetryingChannel(Config_, channelFactory->CreateChannel(Config_->ImageEndpoint)))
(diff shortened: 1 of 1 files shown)


yt/yt/library/containers/cri/cri_executor.cpp
```diff
@@ -704,6 +704,16 @@ class TCriExecutor
             auth->set_registry_token(authConfig.RegistryToken);
         }
     }
+
+    static TRetryChecker GetRetryChecker()
+    {
+        static const auto Result = BIND_NO_PROPAGATE([] (const TError& error) {
+            return IsRetriableError(error) ||
+                   (error.GetCode() == NYT::EErrorCode::Generic &&
+                    error.GetMessage() == "server is not initialized yet");
```
gritukan, line 713: Isn't there any GRPC code for this error since error message hard code does not look like reliable?

koct9i: Exactly. This is a bug in containerd. It returns Generic error instead of Unavailable. I've fixed it - see link in commit message.

gritukan: I see. Let's add TODO to rewrite the condition after your fix is merged and deployed.

koct9i: > Broken containerd stay in old distributives forever. Conditions for this TODO will never met. 😒

gritukan (approved):

savnadya:
Shipped!

savnadya:
Shipped!

PR #297, opened 2023-12-29
Do not discard range attr (and other attrs) in input tables for CppJob

chegoryu:

Unfortunately, I cannot test this change, since the OS version does not have the cython parts, in particular the `CppJob`.

But I need this patch because I copied the code from this commit (this code has been removed from the OS repository): https://github.com/ytsaurus/ytsaurus/commit/4bc99c326b3e8ac8d04d9fa55c594ac6d345c34c

### DIFF
--- a/yt/cpp/mapreduce/client/py_helpers.cpp
+++ b/yt/cpp/mapreduce/client/py_helpers.cpp
@@ -48,7 +48,9 @@ TStructuredJobTableList NodeToStructuredTablePaths(const TNode& node, const TOpe
         if (inputNode.IsNull()) {
             ++intermediateTableCount;
         } else {
-            paths.emplace_back(inputNode.AsString());
+            NYT::TRichYPath path;
+            NYT::Deserialize(path, inputNode);
+            paths.emplace_back(std::move(path));
         }
     }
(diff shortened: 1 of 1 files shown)


dim-an:
Shipped!

chegoryu:
```python
   new_table_path.attributes.update(old_table_path.attributes)
```

Oh, I see...
For me it doesn't work because `cpp_wrapper.pyx` is not OS yet and I used https://github.com/ytsaurus/ytsaurus/commit/4bc99c326b3e8ac8d04d9fa55c594ac6d345c34c revision as a reference.

Btw maybe it is still better to forward this attrs to mapper/reducer and remove this "restore" part from cython code.

As example - you have mapper that forwards some columns by default (let's call them `main_columns`) and you want sometimes forward more columns (`extra_columns`). 

You can do it via `job_constructor_args` and some parsing or you can just try to specify `columns=extra_columns` for `input_table` in python code, but this info will be lost after `new_table_path.attributes.update(old_table_path.attributes)` because mapper will set `columns=main_columns` even when you do something like `preparer.GetInputColumnFilters() + main_columns` in mapper code.

PR #304, opened 2024-01-04
Properly handle chunks without replicas in chunk media statistics computation

gritukan:


maxim-babenko:
Shipped!

savnadya:
Shipped!

Issue #227, opened 2023-12-07
Better security workflow for obtaining first password/token


zlobober:
Currently there is no way to enforce a new user to change his password or token, so if secrets are being passed via a non-secure channel (like in messenger), user might continue using them for indefinitely long. This increases the chance of security breach in case of this non-secure channel compromisation (i.e. hacking a messenger account).

There are two obvious improvements for this workflow:
1. Provide the possibility to set a temporary password, which forces user to reset a password after logging in.
2. Provide the possibility to issue or revoke a token from the web interface, allowing user to safely issue token by himself after first login.

As a result, it becomes possible to not pass (long-lived) password and token from system administrator to user via non-secure channels.

PR #311, opened 2024-01-08
Add password_is_temporary attribute to a user. Allow it to be specified in SetUserPassword API

Krock21:
Issue: #227 

UI issue: https://github.com/ytsaurus/ytsaurus-ui/issues/240

SetUserPassword action now may add `@password_is_temporary` property to a user

This property will be examined by UI in order to force a user on /change_password page

### DIFF
--- a/yt/yt/client/api/security_client.h
+++ b/yt/yt/client/api/security_client.h
@@ -70,7 +70,9 @@ struct TCheckPermissionByAclResult
 
 struct TSetUserPasswordOptions
     : public TTimeoutOptions
-{ };
+{
+    bool PasswordIsTemporary = false;
+};
 
 struct TIssueTokenOptions
     : public TTimeoutOptions
--- a/yt/yt/client/driver/authentication_commands.cpp
+++ b/yt/yt/client/driver/authentication_commands.cpp
@@ -16,6 +16,12 @@ void TSetUserPasswordCommand::Register(TRegistrar registrar)
     registrar.Parameter("current_password_sha256", &TThis::CurrentPasswordSha256_)
         .Default();
     registrar.Parameter("new_password_sha256", &TThis::NewPasswordSha256_);
+    registrar.ParameterWithUniversalAccessor<bool>(
+        "password_is_temporary",
+        [] (TThis* command) -> bool& {
+            return command->Options.PasswordIsTemporary;
+        })
+        .Default(false);
 }
 
 void TSetUserPasswordCommand::DoExecute(ICommandContextPtr context)
--- a/yt/yt/tests/integration/master/test_users.py
+++ b/yt/yt/tests/integration/master/test_users.py
@@ -728,6 +728,41 @@ class TestUsers(YTEnvSetup):
         set("//sys/users/u/@name", "v", authenticated_user="u")
         wait(lambda: self._get_last_seen_time("v") > last_seen, timeout=2)
 
+    @authors("krock21")
+    def test_password_temporary(self):
+        if self.DRIVER_BACKEND == "rpc":
+            return
+
+        create_user("u")
+        assert not exists("//sys/users/u/@password_is_temporary")
+
+        set_user_password("u", "p1")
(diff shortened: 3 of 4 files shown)


yt/yt/server/master/security_server/user_proxy.cpp
```diff
@@ -106,6 +106,11 @@ class TUserProxy
             .SetWritable(true)
             .SetRemovable(true)
             .SetOpaque(true));
+        descriptors->push_back(TAttributeDescriptor(EInternedAttributeKey::PasswordIsTemporary)
+            .SetPresent(static_cast<bool>(user->GetPasswordIsTemporary()))
```
zlobober, line 110: Why do we need static_cast here?

yt/yt/server/master/security_server/user.h
```diff
@@ -213,6 +213,8 @@ class TUser
     DEFINE_BYREF_RO_PROPERTY(std::optional<TString>, HashedPassword);
     //! Salt used for password encryption.
     DEFINE_BYREF_RO_PROPERTY(std::optional<TString>, PasswordSalt);
+    //! A flag that indicates that password is temporary and should be changed.
+    DEFINE_BYVAL_RW_PROPERTY(bool, PasswordIsTemporary);
```
zlobober, line 217: Either this attribute should be system and be a part of a master state (in which case it should be persisted), or it should be user attribute (in which case it should not be stored as a separate field). I think the second option is better here, as this attribute does not have any custom behavior on get/set/remove.

cc @gritukan

zlobober (changes requested):

yt/yt/server/master/security_server/user.h
```diff
@@ -213,6 +213,8 @@ class TUser
     DEFINE_BYREF_RO_PROPERTY(std::optional<TString>, HashedPassword);
     //! Salt used for password encryption.
     DEFINE_BYREF_RO_PROPERTY(std::optional<TString>, PasswordSalt);
+    //! A flag that indicates that password is temporary and should be changed.
```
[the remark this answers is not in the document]
Krock21: Removed this code

yt/yt/server/master/security_server/user_proxy.cpp
Krock21: Copypasted from surroundings, removed

yt/yt/server/master/security_server/user.h
Krock21: Discussed with @gritukan, made it a user attribute

Krock21:
Fixed comments

gritukan (approved):
LGTM

DimasKovas (approved):

DimasKovas (changes requested):
flake8 output:
```
yt/yt/tests/integration/master/test_users.py:740: [E712] comparison to False should be 'if cond is False:' or 'if not cond:'
yt/yt/tests/integration/master/test_users.py:743: [E712] comparison to True should be 'if cond is True:' or 'if cond:'
yt/yt/tests/integration/master/test_users.py:749: [E712] comparison to True should be 'if cond is True:' or 'if cond:'
yt/yt/tests/integration/master/test_users.py:752: [E712] comparison to False should be 'if cond is False:' or 'if not cond:'
yt/yt/tests/integration/master/test_users.py:755: [E712] comparison to False should be 'if cond is False:' or 'if not cond:'
yt/yt/tests/integration/master/test_users.py:758: [E712] comparison to True should be 'if cond is True:' or 'if cond:'
```

DimasKovas (approved):

DimasKovas:
FYI: The last change broke the test:
```
yt/yt/tests/integration/master/test_users.py:740: in test_password_temporary
    assert get("//sys/users/u/@password_is_temporary") is False
E   AssertionError: assert false is False
E    +  where false = get('//sys/users/u/@password_is_temporary')
```

It's because `get` returns `yson.YsonBoolean` type, not `bool`.

I fixed it in the imported PR. It will be merged as soon as all the checks are completed

PR #319, opened 2024-01-11
Replace type=bool argparse usages with intended behavior

achulkov2:
Some argument parsers have historically been broken, since type=bool doesn't actually do what one might expect (convert false/true to the corresponding boolean value). In reality, it calls `bool(argument_value)`, so the only way of actually passing `False` is by specifying an empty string.


Kontakter:
Shipped!

PR #320, opened 2024-01-12
[queues] Take cdw and timestamp from consumer meta in queue agent

savnadya:


yt/yt/server/queue_agent/consumer_controller.cpp
```diff
@@ -388,9 +403,8 @@ class TConsumerSnapshotBuildSession final
 
             const auto& partitionRowInfo = partitionInfo.begin()->second;
 
-            if (partitionRowInfo.CumulativeDataWeight) {
+            if (!consumerPartitionSnapshot->CumulativeDataWeight && partitionRowInfo.CumulativeDataWeight) {
```
achulkov2, line 406: We do not really need to check the second condition here, assigning the optional altogether is fine.

yt/yt/server/queue_agent/consumer_controller.cpp
```diff
@@ -222,6 +222,15 @@ class TConsumerSnapshotBuildSession final
 
                 subConsumerPartitionSnapshot->NextRowIndex = consumerPartitionInfo.NextRowIndex;
                 subConsumerPartitionSnapshot->LastConsumeTime = consumerPartitionInfo.LastConsumeTime;
+
+                if (consumerPartitionInfo.ConsumerMeta) {
+                    if (auto cumulativeDataWeight = consumerPartitionInfo.ConsumerMeta->CumulativeDataWeight) {
```
achulkov2, line 227: It is not necessary to check for null, since this should be our first time setting this value.

achulkov2 (commented):
We should parametrize some tests with the consumers' schema, so that we don't break something for the old consumer schema accidentally.

achulkov2:
LGTM modulo small comments above.

PR #322, opened 2024-01-12
Fix chunk spec fetcher usage with ordered dynamic tables

achulkov2:
Fix critical bug in queue agent static exports.

Using TMasterChunkSpecFetcher with ordered dynamic tables has been broken for a while now: due to a bug in pagination, you cannot fetch chunks past index 100k (in ordered dynamic tables the first chunk index is not 0 in case of trimmed chunks). One needs to specify the chunk count to be -1 as a workaround.

### DIFF
--- a/yt/yt/server/queue_agent/queue_static_table_exporter.cpp
+++ b/yt/yt/server/queue_agent/queue_static_table_exporter.cpp
@@ -446,7 +446,8 @@ private:
         chunkSpecFetcher->Add(
             QueueObject_.ObjectId,
             QueueObject_.ExternalCellTag,
-            QueueObject_.ChunkCount);
+            // XXX(achulkov2, gritukan): YT-11825
+            /*chunkCount*/ -1);
 
         WaitFor(chunkSpecFetcher->Fetch())
             .ThrowOnError();


yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -446,13 +446,16 @@ class TQueueExportTask
...
+            // XXX(achulkov2, gritukan): YT-11825
+            /*chunkCount*/ -1);
 
         WaitFor(chunkSpecFetcher->Fetch())
             .ThrowOnError();
 
         ChunkSpecs_ = chunkSpecFetcher->GetChunkSpecsOrderedNaturally();
         YT_LOG_DEBUG("Finished fetching chunk specs (Count: %v)", ChunkSpecs_.size());
+
+        YT_LOG_FATAL_UNLESS(std::ssize(ChunkSpecs_) == QueueObject_.ChunkCount, "Fetched chunk spec count mismatch: expected %v, received %v", QueueObject_.ChunkCount, ChunkSpecs_.size());
```
gritukan, line 458: Are you sure that dynamic stores will not make these values different?

savnadya:
Shipped!

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -446,13 +446,16 @@ class TQueueExportTask
         chunkSpecFetcher->Add(
             QueueObject_.ObjectId,
             QueueObject_.ExternalCellTag,
-            QueueObject_.ChunkCount);
+            // XXX(achulkov2, gritukan): YT-11825
```
gritukan, line 449: Why me?)

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
achulkov2: Sadly, you are right in case of @enable_dynamic_store_read = %true. I don't see a way to make this check work, so I removed it.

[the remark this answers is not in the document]
achulkov2: NA.

DimasKovas (approved):

PR #343, opened 2024-01-22
Fix race in multiplexing value consumers

gritukan:

### DIFF
--- a/yt/yt/server/job_proxy/user_job_write_controller.cpp
+++ b/yt/yt/server/job_proxy/user_job_write_controller.cpp
@@ -73,7 +73,7 @@ public:
     std::unique_ptr<IFlushableValueConsumer> CreateConsumer(int index, TTableSchemaPtr schema)
     {
         YT_VERIFY(!Closed_);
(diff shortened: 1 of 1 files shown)


yt/yt/server/job_proxy/user_job_write_controller.cpp
```diff
@@ -126,6 +126,14 @@ class TMultiplexingWriter
 
             return
                 BIND([writer = Owner_->Underlying_, rowBuffer = RowBuffer_, rows = std::move(Rows_)] {
+                    // NB: In multiplexing writer, there are multiple value consumers pointing to the same writer.
```
gritukan, line 129: There is a similar code https://github.com/ytsaurus/ytsaurus/blame/49e9d43be87c9973667714c25722258e48d94af6/yt/yt/client/table_client/value_consumer.cpp#L335 for the regular value consumers

gritukan:
This fix is also needed in 23.1, 23.2 stable branches.

dim-an:
Shipped

PR #350, opened 2024-01-25
Use master transactions as query leases

gritukan:

### DIFF
--- a/yt/python/yt/environment/init_query_tracker_state.py
+++ b/yt/python/yt/environment/init_query_tracker_state.py
@@ -393,6 +393,89 @@ TRANSFORMS[5] = [
     ),
 ]
 
+TRANSFORMS[6] = [
+    Conversion(
+        "active_queries",
+        table_info=TableInfo(
+            [
+                ("query_id", "string"),
+            ],
+            [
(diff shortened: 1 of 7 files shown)


yt/python/yt/environment/init_query_tracker_state.py
```diff
@@ -393,6 +393,89 @@
...
+            optimize_for="lookup",
+            attributes={
+                "tablet_cell_bundle": SYS_BUNDLE_NAME,
+            },
+        ),
+        use_default_mapper=True,
+    ),
+    Conversion(
+        "finished_queries",
+        table_info=TableInfo(
```
Krisha11, line 432: What changes in this and the following conversions?

gritukan: It adds a new field lease_transaction_id and removes field ping_time

Krisha11:
LGTM - ready to merge after fixing conflict/codestyle

yt/python/yt/environment/init_query_tracker_state.py
gritukan: Indeed, fixed.

Krisha11 (approved):

PR #359, opened 2024-01-30
Add option for setting memory limit for user job containers

koct9i:
This adds controller agent operation options:
"set_container_memory_limit" and "container_memory_overhead".

### DIFF
--- a/yt/yt/server/lib/job_proxy/config.h
+++ b/yt/yt/server/lib/job_proxy/config.h
@@ -249,6 +249,8 @@ public:
 
     std::optional<double> ContainerCpuLimit;
 
+    std::optional<i64> SlotContainerMemoryLimit;
+
     NYT::NRpcProxy::TApiServiceConfigPtr ApiService;
 
     std::optional<int> StatisticsOutputTableCountLimit;
--- a/yt/yt/server/node/exec_node/job.cpp
+++ b/yt/yt/server/node/exec_node/job.cpp
@@ -2488,6 +2488,10 @@ TJobProxyInternalConfigPtr TJob::CreateConfig()
         }
     }
 
+    if (UserJobSpec_ && UserJobSpec_->slot_container_memory_limit()) {
+        proxyConfig->SlotContainerMemoryLimit = UserJobSpec_->slot_container_memory_limit();
+    }
+
     proxyConfig->MemoryTracker->MemoryStatisticsCachePeriod = proxyConfig->MemoryTracker->UseSMapsMemoryTracker
         ? CommonConfig_->SMapsMemoryTrackerCachePeriod
         : CommonConfig_->MemoryTrackerCachePeriod;
--- a/yt/yt/server/node/exec_node/job_environment.cpp
+++ b/yt/yt/server/node/exec_node/job_environment.cpp
@@ -1182,6 +1182,7 @@ private:
         }
 
         spec->Resources.CpuLimit = config->ContainerCpuLimit;
+        spec->Resources.MemoryLimit = config->SlotContainerMemoryLimit;
 
         const auto& cpusetCpu = SlotCpusetCpus_[slotIndex];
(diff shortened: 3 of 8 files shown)


yt/yt/server/node/exec_node/job_environment.cpp
```diff
@@ -1186,6 +1186,11 @@ class TCriJobEnvironment
 
         spec->Resources.CpuLimit = config->ContainerCpuLimit;
 
+        if (config->ContainerMemoryLimit) {
+            // Sum up limits because job-proxy lives in the same container.
```
gritukan, line 1190: job proxy

yt/yt/server/node/exec_node/job.cpp
```diff
@@ -2394,6 +2394,11 @@ TJobProxyInternalConfigPtr TJob::CreateConfig()
         }
     }
 
+    if (UserJobSpec_) {
+        proxyConfig->ContainerMemoryLimit = UserJobSpec_->container_memory_limit();
+        proxyConfig->JobProxyMemoryLimit = UserJobSpec_->job_proxy_memory_limit();
```
gritukan, line 2399: Job proxy memory limit is useful even if there is no user job

yt/yt/server/lib/job_proxy/config.h
```diff
@@ -249,6 +249,10 @@ class TJobProxyInternalConfig
 
     std::optional<double> ContainerCpuLimit;
 
+    i64 ContainerMemoryLimit = 0;
```
gritukan, line 252: Do we need initialization here?

yt/yt/ytlib/controller_agent/proto/job.proto
```diff
@@ -254,6 +254,8 @@ message TUserJobSpec
 
     optional double container_cpu_limit = 53 [default = 0];
 
+    optional int64 container_memory_limit = 72;
```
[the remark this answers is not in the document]
koct9i: After second thought, this should include job proxy memory limit.
Enforcing memory limit for them together makes sense even if they are isolated.

yt/yt/server/node/exec_node/job.cpp
koct9i: If we limit them both together then job proxy without user job could be limited too.

yt/yt/server/lib/job_proxy/config.h
koct9i: This means unlimted by default. Probably it's better wrap it with std::optional.

yt/yt/ytlib/controller_agent/proto/job.proto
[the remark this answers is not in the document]
koct9i: Maybe rename it into slot_memory_limit?

yt/yt/server/lib/job_proxy/config.h
gritukan: Yes

yt/yt/ytlib/controller_agent/proto/job.proto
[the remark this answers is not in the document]
gritukan: Seems like a good idea

k-pogorelov (changes requested):

yt/yt/ytlib/controller_agent/proto/job.proto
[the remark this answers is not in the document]
koct9i: I'm talking about enforcing overall limit for job-proxy and user-job together.
Including all initializaion stuff and job shell.

Name "slot_contianer_memory_limit" should be fine.

k-pogorelov (approved):

k-pogorelov:
Rebase this PR, please.
There is build error.

koct9i:
@k-pogorelov done

Issue #223, opened 2023-12-06
Support for Validate & Explain in Queries


zlobober:
Subj.

Must look like a yql-specific setting which enables a special way of query running.

zlobober:
Can we close this one?

Krock21:
Yes. Supported from [Query Tracker 0.0.5](https://github.com/ytsaurus/ytsaurus/releases/tag/docker%2Fquery-tracker%2F0.0.5)

PR #362, opened 2024-01-31
Add execution mode to query tracker. Support validate and explain execution modes. Promote yql plugin version

Krock21:
Issue: #223

Commit in YDB repository, adjusting yql plugin for yt: https://github.com/ydb-platform/ydb/commit/0c915e616b63e7047189c91e523a47e6476bd3df

### DIFF
--- a/.github/workflows/cpp.yaml
+++ b/.github/workflows/cpp.yaml
@@ -46,7 +46,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: ydb-platform/ydb
-          ref: 341849f5d448ac6e7362551e08f40533bbfb0ffb
+          ref: 31adbe01b564f452d8c0e21b8c3331bc8a324855
           path: ydb
 
   build:
--- a/yt/yql/plugin/bridge/interface.h
+++ b/yt/yql/plugin/bridge/interface.h
@@ -96,8 +96,10 @@ using TFuncBridgeRun = TBridgeQueryResult*(
     const char* impersonationUser,
     const char* queryText,
     const char* settings,
+    int settingsLength,
     const TBridgeQueryFile* files,
-    int fileCount);
+    int fileCount,
+    int executeMode);
 using TFuncBridgeGetProgress = TBridgeQueryResult*(TBridgeYqlPlugin* plugin, const char* queryId);
 using TFuncBridgeAbort = TBridgeAbortResult*(TBridgeYqlPlugin* plugin, const char* queryId);
 using TFuncBridgeFreeAbortResult = void(TBridgeAbortResult* result);
--- a/yt/yt/server/query_tracker/yql_engine.cpp
+++ b/yt/yt/server/query_tracker/yql_engine.cpp
@@ -9,6 +9,7 @@
 #include <yt/yt/ytlib/api/native/connection.h>
 
 #include <yt/yt/ytlib/yql_client/yql_service_proxy.h>
+#include <yt/yt/ytlib/yql_client/public.h>
 
 #include <yt/yt/core/ytree/convert.h>
 #include <yt/yt/core/ytree/attributes.h>
@@ -40,6 +41,7 @@ class TYqlSettings
 {
 public:
     std::optional<TString> Stage;
+    EExecuteMode ExecuteMode;
 
     REGISTER_YSON_STRUCT(TYqlSettings);
 
@@ -47,6 +49,8 @@ public:
     {
         registrar.Parameter("stage", &TThis::Stage)
             .Optional();
+        registrar.Parameter("execution_mode", &TThis::ExecuteMode)
+            .Default(EExecuteMode::Run);
     }
 };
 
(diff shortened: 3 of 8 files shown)


yt/yt/server/query_tracker/yql_engine.cpp
```diff
@@ -40,13 +40,16 @@ class TYqlSettings
 {
 public:
     std::optional<TString> Stage;
+    std::optional<int> ExecutionMode;
```
gritukan, line 43: Can we use enum here?

zlobober: I agree with Grigory. In YT code we tend to use enums wherever possible, even if the underlying system operates with int. An example from CHYT world: https://github.com/ytsaurus/ytsaurus/blob/main/yt/chyt/server/private.h#L127-L138. 

This interface is the last mile before passing values to essentially a third-party code via C ABI, so having ENUM here and upwards from here is perfectly fine. Also, we would not want a strange integer parameter in our own API (e.g. in HTTP API it should be enum, which is serialized/deserialized as a string literal).

.github/workflows/cpp.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: ydb-platform/ydb
-          ref: 341849f5d448ac6e7362551e08f40533bbfb0ffb
+          ref: 0c915e616b63e7047189c91e523a47e6476bd3df # TODO UPDATE WITH WORKING COMMIT
```
Krisha11, line 49: Let's do this before moving the link? In the current situation, new nightly images will be broken(

Krock21: Yes, I meant "DO NOT MERGE UNTIL A WORKING COMMIT IS CREATED" :)

yt/yt/server/query_tracker/yql_engine.cpp
Krock21: Changed to enum

Seems that registrar can handle enums well, didn't expect that

yt/yt/server/query_tracker/yql_engine.cpp
```diff
@@ -133,6 +143,7 @@ class TYqlQueryHandler
     const NApi::NNative::IConnectionPtr Connection_;
     const TYqlSettingsPtr Settings_;
     const TString Stage_;
+    const EExecuteMode ExecutionMode_;
```
zlobober, line 146: It should be either execute mode or execution mode both in type name and in variable.

yt/yt/server/query_tracker/yql_engine.cpp
```diff
@@ -95,6 +104,7 @@ class TYqlQueryHandler
         ToProto(req->mutable_query_id(), QueryId_);
         yqlRequest->set_query(Query_);
         yqlRequest->set_settings(ConvertToYsonString(SettingsNode_).ToString());
+        yqlRequest->set_mode(static_cast<int>(ExecuteMode_));
```
gritukan, line 107: ToProto<int>

yt/yql/plugin/bridge/interface.h
```diff
@@ -96,8 +96,10 @@ using TFuncBridgeRun = TBridgeQueryResult*(
     const char* impersonationUser,
     const char* queryText,
     const char* settings,
+    int settingsLength,
     const TBridgeQueryFile* files,
-    int fileCount);
+    int fileCount,
+    int executeMode);
```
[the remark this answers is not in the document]
Krock21: It can't. We have 2 enums:
1. [NYT::NYqlClient::EExecuteMode](https://github.com/ytsaurus/ytsaurus/blob/798da40d10e94e2530e1413978c328b40746a47f/yt/yt/ytlib/yql_client/public.h#L10). Lives in ytsaurus and passed to yql_client as int
2. [NYql::NEmbedded::EExecuteMode](https://github.com/ydb-platform/ydb/blob/9c92253aebf1829c27ce2e7337b635f73d92beb8/ydb/library/yql/public/embedded/yql_embedded.h#L30). Lives in ydb and doesn't have an int representation

Logically we should use a first one, but it can't be used since interface lives in ydb repository

Probably we can use the second one, with some additional work
I see that we use `const char*` as parameters to this interface, and it says "This is a plain C version of an interface described in yt/yql/plugin/plugin.h." in the file, so we may have troubles including 2nd enum

Krock21:
Rebased, fixed comments

.github/workflows/cpp.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: ydb-platform/ydb
-          ref: 341849f5d448ac6e7362551e08f40533bbfb0ffb
+          ref: 0c915e616b63e7047189c91e523a47e6476bd3df # DO NOT MERGE. WAIT FOR A WORKING COMMIT AND MERGE WITH IT
```
Krisha11, line 49: Now you can change it to
31adbe01b564f452d8c0e21b8c3331bc8a324855

Krisha11 (approved):

gritukan (approved):

yt/yt/server/query_tracker/yql_engine.cpp
```diff
@@ -40,13 +41,16 @@ class TYqlSettings
...
     std::optional<TString> Stage;
+    EExecuteMode ExecuteMode;
 
     REGISTER_YSON_STRUCT(TYqlSettings);
 
     static void Register(TRegistrar registrar)
     {
         registrar.Parameter("stage", &TThis::Stage)
             .Optional();
+        registrar.Parameter("execution_mode", &TThis::ExecuteMode)
```
maxim-babenko, line 52: So is it execution_mode or execute_mode?

Krock21: execution_mode in the API
ExecuteMode in variables in code

PR #379, opened 2024-02-09
Add QT ya make build in CI

Krisha11:
This branch was used primarily for testing - and contains a large number of such commits.

The closest to the current version can be considered a successful build from
https://github.com/ytsaurus/ytsaurus/actions/runs/7843895758/job/21405173965
Which differs from the current version only in style corrections

### DIFF
--- a/.github/workflows/build-jobs.yaml
+++ b/.github/workflows/build-jobs.yaml
@@ -195,6 +195,9 @@ jobs:
           yt-ya-make ../ytsaurus/ytsaurus/yt/yt/server/all
           yt-ya-make ../ytsaurus/ytsaurus/yt/yql/agent/bin
 
+          yt-ya-make ../ytsaurus/ytsaurus/yt/yt/python/yson_shared
+          yt-ya-make ../ytsaurus/ytsaurus/yt/yt/python/driver/native_shared
+
           alias ydb-ya-make="../ytsaurus/ydb/ya make -T --ignore-recurses --no-src-links --build=release  --output=."
 
           ydb-ya-make ../ytsaurus/ydb/ydb/library/yql/tools/mrjob
@@ -223,7 +226,6 @@ jobs:
           ydb-ya-make ${udfs_prefix}/unicode_base
           ydb-ya-make ${udfs_prefix}/url_base
           ydb-ya-make ${udfs_prefix}/yson2
-          ydb-ya-make ${udfs_prefix}/dsv
 
       - name: Build Python packages
         if: ${{ inputs.build-python }}
--- a/.github/workflows/build-jobs.yaml
+++ b/.github/workflows/build-jobs.yaml
@@ -127,11 +127,6 @@ jobs:
           cd ../build_relwithdebinfo
           ninja ytserver-all
 
-      - name: Build ytsaurus (ya make, relwithdebinfo)
-        if: ${{ inputs.build-query-tracker-ya-make }}
-        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/build-jobs.yaml
```diff
@@ -173,10 +186,55 @@ jobs:
...
           ninja ytserver-yql-agent
 
+      - name: Build query tracker (ya make, relwithdebinfo)
+        if: ${{ inputs.build-query-tracker-ya-make }}
+        run: |
+          mkdir -p ../build_qt_ya_make
+          cd ../build_qt_ya_make
+          rm -rf *
+
+          ../ytsaurus/ytsaurus/ya make -T --build=relwithdebinfo  --output=../ya_build ../ytsaurus/ytsaurus/yt/yql/agent/bin
```
Krisha11, line 196: It takes about 3 hours to complete a single build. So, in case of nightly builds, I suggest to build only the RelWithDebInfo version

.github/workflows/build-jobs.yaml
```diff
@@ -123,6 +127,15 @@ jobs:
...
           ninja ytserver-all
 
+      - name: Build ytsaurus (ya make, relwithdebinfo)
+        if: ${{ inputs.build-query-tracker-ya-make }}
+        run: |
+          mkdir -p ../ya_build
+          cd ../ya_build
+          rm -rf *
+
+          ../ytsaurus/ytsaurus/ya make -T --build=relwithdebinfo  --output=../ya_build ../ytsaurus/ytsaurus/yt/yt
```
Krisha11, line 137: We build all targets in yt/yt here to save the previous logic (checking that our code can be build by ya make)
https://github.com/ytsaurus/ytsaurus/pull/379/files#diff-0e63f0336811c0261bddff6fc554b7bee0e469a72fd20a1a3bacec95179f7036L52

It also doesn't change the build time too much.

yt/python/packages/build_ytsaurus_packages.sh
```diff
@@ -8,13 +8,15 @@ script_name=$0
...
 ytsaurus_package_name=""
+not_prepare_bindings_libraries=false
 
 print_usage() {
     cat << EOF
 Usage: $script_name [-h|--help]
                     [--ytsaurus-source-path /path/to/ytsaurus.repo (default: $ytsaurus_source_path)]
                     [--ytsaurus-build-path /path/to/ytsaurus.build (default: $ytsaurus_build_path)]
                     [--ytsaurus-package-name some-ytsaurus-package-name (default: all packages will be build) (values: ytsaurus-client, ytsaurus-yson, ytsaurus-local, ytsaurus-native-driver)]
+                    [--not-prepare-bindings-libraries]
```
gritukan, line 19: Maybe `binding-libraries`?

yt/python/packages/build_ytsaurus_packages.sh
```diff
@@ -8,13 +8,15 @@ script_name=$0
 ytsaurus_source_path="."
 ytsaurus_build_path="."
 ytsaurus_package_name=""
+not_prepare_bindings_libraries=false
```
gritukan, line 11: Let's make a variable without not with true default

yt/python/packages/build_ytsaurus_packages.sh
Krisha11: In this case it will be inconsistent with this and other places
https://github.com/ytsaurus/ytsaurus/pull/379/files#diff-d7e1077c377a7d9234aeabf23282b2eb66f3c06286a1029e83636bdfc5078007R71

Also changing it in all places in this PR seems strange too

Krisha11:
Shipped!

PR #384, opened 2024-02-14
Automatic update for HTTPS proxy TLS certificates

koct9i:
- Add methods for SSL context update
  This is required for TLS certificate rotation without downtime.
  
- Automatically update HTTPS server TLS certificates
  
- yt/environmnt: keep TLS certificates in files
  
- Add test for http proxy certificate update

### DIFF
--- a/yt/yt/core/crypto/tls.h
+++ b/yt/yt/core/crypto/tls.h
@@ -20,6 +20,10 @@ class TSslContext
 public:
     TSslContext();
 
+    void Reset();
+    void Commit(TInstant time = TInstant::Zero());
+    TInstant GetCommitTime();
+
     void UseBuiltinOpenSslX509Store();
 
     void SetCipherList(const TString& list);
@@ -48,6 +52,7 @@ public:
 
 private:
     const TIntrusivePtr<TSslContextImpl> Impl_;
+    TInstant CommitTime_;
 };
 
 DEFINE_REFCOUNTED_TYPE(TSslContext)
--- a/yt/yt/core/https/server.cpp
+++ b/yt/yt/core/https/server.cpp
@@ -84,6 +84,7 @@ IServerPtr CreateServer(
     } else {
         YT_ABORT();
     }
+    sslContext->Commit();
 
     auto address = TNetworkAddress::CreateIPv6Any(config->Port);
     auto tlsListener = sslContext->CreateListener(address, poller, acceptor);
--- a/yt/yt/core/https/server.cpp
+++ b/yt/yt/core/https/server.cpp
@@ -2,14 +2,22 @@
 #include "config.h"
 
 #include <yt/yt/core/http/server.h>
+#include <yt/yt/core/http/private.h>
 
 #include <yt/yt/core/crypto/tls.h>
+
+#include <yt/yt/core/logging/log.h>
+
+#include <yt/yt/core/misc/fs.h>
 
 #include <yt/yt/core/net/address.h>
 
 #include <yt/yt/core/concurrency/poller.h>
+#include <yt/yt/core/concurrency/periodic_executor.h>
 
 namespace NYT::NHttps {
+
+static const auto& Logger = NHttp::HttpLogger;
 
 using namespace NNet;
 using namespace NHttp;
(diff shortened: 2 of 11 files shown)


yt/yt/tests/integration/proxies/test_http_proxy.py
```diff
@@ -1154,3 +1177,52 @@ def test_ping_https_verify_false(self):
...
+    def test_certificate_broken_update(self):
+        proxy_cert, proxy_cert_key = self._get_proxy_cert_path()
+
+        with open(proxy_cert, 'rb') as f:
+            cert_content = f.read()
+
+        with open(proxy_cert, 'w') as f:
+            f.write('hello')
+
+        time.sleep(3)
```
Kontakter, line 1221: It is better to use `wait` function

yt/yt/tests/integration/proxies/test_http_proxy.py
```diff
@@ -1154,3 +1177,52 @@ def test_ping_https_verify_false(self):
...
+
+    @authors("khlebnikov")
+    def test_certificate_broken_update(self):
+        proxy_cert, proxy_cert_key = self._get_proxy_cert_path()
+
+        with open(proxy_cert, 'rb') as f:
+            cert_content = f.read()
+
+        with open(proxy_cert, 'w') as f:
+            f.write('hello')
```
Kontakter, line 1219: Please use double quotes

yt/yt/tests/integration/proxies/test_http_proxy.py
koct9i: I dont'n see how to do that without adding structured logging events for that case. This 
seems too overengineed.

I've removed this test case. It's not so important anyway.

yalekseev:
Shipped!

yt/yt/core/https/server.cpp
```diff
@@ -22,8 +30,9 @@ class TServer
     : public IServer
 {
 public:
-    explicit TServer(IServerPtr underlying)
+    explicit TServer(IServerPtr underlying, TPeriodicExecutorPtr certificateUpdater)
```
maxim-babenko, line 33: explicit is now redundant

yt/yt/core/crypto/tls.h
```diff
@@ -48,6 +52,7 @@ class TSslContext
 
 private:
     const TIntrusivePtr<TSslContextImpl> Impl_;
+    TInstant CommitTime_;
```
maxim-babenko, line 55: Why placing it here instead of hiding inside pimpl?

yt/yt/core/https/server.cpp
```diff
@@ -62,28 +77,66 @@ class TServer
...
+    TPeriodicExecutorPtr certificateUpdater;
+    if (sslConfig->UpdatePeriod &&
+        sslConfig->CertChain->FileName &&
+        sslConfig->PrivateKey->FileName)
+    {
+        certificateUpdater = New<TPeriodicExecutor>(
+            poller->GetInvoker(),
+            BIND([=, serverName = config->ServerName] {
+                try {
+                    auto modificationTime = Max(
```
maxim-babenko, line 120: Now you have IO in poller thread, seems like a problem

yt/yt/core/https/server.cpp
```diff
@@ -62,28 +77,66 @@ class TServer
...
+                        NFS::GetPathStatistics(*sslConfig->CertChain->FileName).ModificationTime,
+                        NFS::GetPathStatistics(*sslConfig->PrivateKey->FileName).ModificationTime);
+
+                    // Detect fresh and stable updates.
+                    if (modificationTime > sslContext->GetCommitTime() &&
+                        modificationTime + sslConfig->UpdatePeriod <= TInstant::Now())
+                    {
+                        YT_LOG_INFO("Updating TLS certificates (ServerName: %v, ModificationTime: %v)", serverName, modificationTime);
```
maxim-babenko, line 128: Let's place the arguments on separate lines

yt/yt/core/https/server.cpp
```diff
@@ -62,28 +77,66 @@ class TServer
...
+
+    auto sslConfig = config->Credentials;
+    TPeriodicExecutorPtr certificateUpdater;
+    if (sslConfig->UpdatePeriod &&
+        sslConfig->CertChain->FileName &&
+        sslConfig->PrivateKey->FileName)
+    {
+        certificateUpdater = New<TPeriodicExecutor>(
+            poller->GetInvoker(),
+            BIND([=, serverName = config->ServerName] {
```
maxim-babenko, line 118: Why not just =? Capturing `config` is no big deal.

Issue #386, opened 2024-02-15
Remove port when validating yt proxy in chyt controller


chegoryu:
When I have `YT_PROXY` env with port, the following command fails:

```shell
$ CHYT_CTL_ADDRESS=chyt.my.cluster.yt YT_PROXY=my.cluster.yt:80 yt clickhouse ctl status ch_public
yt clickhouse ctl status: error: bad cluster proxy choice: my.cluster.yt:80
controller http://chyt.my.cluster.yt serves only following clusters: ['http-proxies-lb.my.cluster.svc.cluster.local', 'my.cluster.yt']
```

I beleive that it is better to cut port from yt proxy for this validation.

PR #387, opened 2024-02-15
Normalize yt proxy before comparing in strawberry parser

chegoryu:
Fixes: https://github.com/ytsaurus/ytsaurus/issues/386

### DIFF
--- a/yt/python/yt/cli/strawberry_parser.py
+++ b/yt/python/yt/cli/strawberry_parser.py
@@ -16,7 +16,7 @@ def _normalize_proxy(proxy):
     if ":" in proxy:
         proxy = proxy.rsplit(":", 1)[0]
 
-    return proxy
+    return proxy.lower()
 
 
 def _strawberry_ctl_handler(family):
--- a/yt/python/yt/cli/strawberry_parser.py
+++ b/yt/python/yt/cli/strawberry_parser.py
@@ -8,6 +8,17 @@ import yt.yson as yson
 import os
 
 
+def _normalize_proxy(proxy):
+    for prefix in ["http://", "https://"]:
+        if proxy.startswith(prefix):
+            proxy = proxy[len(prefix):]
+
+    if ":" in proxy:
+        proxy = proxy.rsplit(":", 1)[0]
+
+    return proxy
+
+
 def _strawberry_ctl_handler(family):
     def handler(**kwargs):
         address = kwargs.pop("address")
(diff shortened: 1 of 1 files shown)


yt/python/yt/cli/strawberry_parser.py
```diff
@@ -8,6 +8,17 @@
 import os
 
 
+def _normalize_proxy(proxy):
+    for prefix in ["http://", "https://"]:
+        if proxy.startswith(prefix):
+            proxy = proxy[len(prefix):]
```
chegoryu, line 14: I can't you [removeprefix](https://docs.python.org/3/library/stdtypes.html#str.removeprefix) here due to python 3.7 :(

yt/python/yt/cli/strawberry_parser.py
```diff
@@ -17,7 +28,7 @@ def handler(**kwargs):
         cluster_proxy = yt.config["proxy"]["url"]
         params = kwargs
 
-        if cluster_proxy not in proxy_choices:
+        if _normalize_proxy(cluster_proxy) not in map(_normalize_proxy, proxy_choices):
```
DimasKovas, line 31: Normalization in this check is not sufficient. You will pass through a `bad cluster proxy choice` error, but you will get another less clear error (404 not found or something like this), because it is used to form a url to send queries (`https://{controller_address}/{proxy}`). So, if you want to normalize `cluster_proxy`, you should actually change it and pass on to strawberry.make_request_generator. And that's also the reason why it's wrong to normalize `proxy_choices` here (you can normalize them manually in strawberry controller config, though)

yt/python/yt/cli/strawberry_parser.py
```diff
@@ -8,6 +8,17 @@
 import os
 
 
+def _normalize_proxy(proxy):
+    for prefix in ["http://", "https://"]:
+        if proxy.startswith(prefix):
+            proxy = proxy[len(prefix):]
+
+    if ":" in proxy:
```
DimasKovas, line 16: Not sure that truncating the port is a good idea, at least when it is not a "well known" port (80/443)

DimasKovas (changes requested):
Well, the current schema to specify a cluster in strawberry controller is doing well in our internal YTsaurus installations (because we use short names like `hahn` to refer to a cluster, not a real address), but isn't suitable for Open Source installations. We have a plan to fix this (in short, the plan is to get the full strawberry controller url from a cluster-specific cli config instead of crafting it from a ctl address and a cluster proxy), but it will take a while to implement. For now it's ok to try normalizing a proxy if it improves usability in certain cases, but need to fix mentioned issues, because now it dosn't seem to be working

zlobober:
Also, without a test such behavior will definitely break in the future when somebody decides to refactor some code around.

yt/python/yt/cli/strawberry_parser.py
chegoryu: Yeah, it's harder than I expected, so I'll wait for the issue to be resolved then.

Issue #385, opened 2024-02-14
Pass PATH variable to user job in CRI job environment


gritukan:
I have the following environment in job:

```
"ENV: environ({'SHELL': '/bin/bash', 'YT_JOB_PROXY_SOCKET_PATH': '/slot/pipes/yt-node-9029-8-job-proxy-4', 'YT_START_ROW_INDEX': '0', 'YT_JOB_ID': 'c38925c0-128c6842-13440384-357', 'PWD': '/slot/sandbox', 'LOGNAME': 'root', 'YT_OPERATION_ID': '17f9155b-52002d0a-134403e8-b18b3911', 'YT_TOKEN': 'haa0gLMUKNp0FfgsSnPN', 'YT_JOB_INDEX': '0', 'USER': 'root', 'SHLVL': '0', 'LD_LIBRARY_PATH': './modules/_shared:./tmpfs/modules/_shared', 'YT_JOB_COOKIE': '0', 'YT_PROXY': 'planck.yt.nemax.nebiuscloud.net', 'YT_TASK_JOB_INDEX': '0', 'YT_ALLOW_HTTP_REQUESTS_TO_YT_FROM_JOB': '1', 'YT_FORBID_REQUESTS_FROM_JOB': '1', '_': '/opt/conda/bin/python3.11', 'LC_CTYPE': 'C.UTF-8'})"
```

PATH seems to be missing.

PR #390, opened 2024-02-16
Add job proxy option to forward all environment variables to jobs

koct9i:
This is required for forwarding environment variables set in docker image into user job environment.

Issue #385


yt/yt/server/job_proxy/user_job.cpp
```diff
@@ -1178,40 +1179,70 @@ class TUserJob
...
+            if (index) {
+                Environment_[*index] = variable;
+            } else {
+                Environment_.push_back(variable);
+                EnvironmentIndex_[name] = Environment_.size()-1;
+            }
+        } else {
+            // Without "=" works as unset.
+            auto index = EnvironmentIndex_.FindPtr(variable);
+            if (index) {
```
gritukan, line 1197: `if (auto index = `

yt/yt/server/job_proxy/user_job.cpp
```diff
@@ -1178,40 +1179,70 @@ class TUserJob
         YT_LOG_INFO("Pipes initialized");
     }
 
+    void SetEnvironment(const TString& variable)
+    {
+        auto sep = variable.find('=');
+        if (sep != TString::npos) {
+            const auto name = variable.substr(0, sep-1);
+            auto index = EnvironmentIndex_.FindPtr(name);
+            if (index) {
```
gritukan, line 1188: `if (auto index = ...`

yt/yt/server/lib/job_proxy/config.h
```diff
@@ -228,6 +228,9 @@ class TJobProxyInternalConfig
     //! For testing purposes only.
     bool DoNotSetUserId;
 
+    // Forward variables from job proxy environment to user job.
```
gritukan, line 231: //!

yt/yt/server/lib/exec_node/config.h
```diff
@@ -697,6 +697,9 @@ class TJobProxyConfig
     //! Enables job abort on violated memory reserve.
     bool AlwaysAbortOnMemoryReserveOverdraft;
 
+    // Forward variables from job proxy environment to user job.
```
gritukan, line 700: //!

yt/yt/server/job_proxy/user_job.cpp
```diff
@@ -1178,40 +1179,70 @@ class TUserJob
         YT_LOG_INFO("Pipes initialized");
     }
 
+    void SetEnvironment(const TString& variable)
```
k-pogorelov, line 1182: Let's overload this method and do not serialize/deserialize strings

yt/yt/library/process/process.cpp
```diff
@@ -144,6 +144,15 @@ TErrorOr<TString> ResolveBinaryPath(const TString& binary)
     return failure();
 }
 
+std::vector<TString> GetEnviron()
+{
+    std::vector<TString> env;
```
k-pogorelov, line 149: reserve?

k-pogorelov (changes requested):

yt/yt/server/lib/job_proxy/config.h
```diff
@@ -228,6 +228,9 @@ class TJobProxyInternalConfig
...
+    bool ForwardAllEnvironmentVariables;
```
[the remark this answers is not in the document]
koct9i: And why exactly it should be in dynamic config? =)

Effect of this option depends on chosen job-environment which part of static config.

yt/yt/server/job_proxy/user_job.cpp
koct9i: What do you mean? I don't see reason to do any more complicated optimizations here. It's a few kb at most.

yt/yt/server/lib/job_proxy/config.h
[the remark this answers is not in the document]
k-pogorelov: All dynamizable options should be in dynamic config.
But if there are no dynamic configs for different environments, ok, but we should implement it later.

yt/yt/server/job_proxy/user_job.cpp
k-pogorelov: I mean that it would be better to implement 2 more variable setters than Format(...) strings just to pass arguments to the function.

Yes, it's cold code, but I'm worried about style.

yt/yt/server/lib/job_proxy/config.h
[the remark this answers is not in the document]
gritukan: I agree with Konstantin here, since for CRI this option can be dynamic and I see no reason to keep it static, it won't make code more complicated in any sence.

yt/yt/server/job_proxy/user_job.cpp
gritukan: To be honest, I'm not sure that this is _that_ important. From my perspective, function with two arguments is a little bit nicer, but it's really a small thing.

yt/yt/library/process/process.cpp
k-pogorelov: This is marked resolved but was not fixed.

k-pogorelov (changes requested):

yt/yt/library/process/process.cpp
gritukan: It's fixed below https://github.com/ytsaurus/ytsaurus/pull/390/files#diff-b254d392f69ccc1e18b0111c6416b3faaf71ce3dcf83c3aff9a3a8a4adccec47R154

k-pogorelov (approved):

yt/yt/server/job_proxy/user_job.cpp
```diff
@@ -1192,34 +1193,59 @@ class TUserJob
...
+    {
+        if (auto sep = variable.find('='); sep != TString::npos) {
+            const auto name = variable.substr(0, sep - 1);
```
koct9i, line 1199: off-by-one =)

k-pogorelov (approved):

PR #391, opened 2024-02-16
Wrap and attribute origin of error loaded from lock file in disabled location

koct9i:
Currently there is no clear difference between fresh error and
error recovered from previous run.


gritukan (approved):

Kontakter:
Shipped!

PR #396, opened 2024-02-20
Support SortColumn in run_sort method

chegoryu:
Without this change it is impossible to sort table in descending order without using raw specs:

```python
yt.yson.common.YsonError: Value <yt.wrapper.schema.table_schema.SortColumn object at 0x7f9f258cb450> cannot be serialized to YSON since it has unsupported type "<class 'yt.wrapper.schema.table_schema.SortColumn'>"
```

### DIFF
--- a/yt/python/yt/wrapper/tests/test_operations.py
+++ b/yt/python/yt/wrapper/tests/test_operations.py
@@ -185,8 +185,8 @@ class TestOperations(object):
 
         assert yt.is_sorted(table)
 
-        yt.run_sort(table, sort_by=[SortColumn("x", sort_order="descending")])
-        assert list(reversed([{"x": x, "y": y} for x, y in sorted(columns, key=lambda c: c[0])])) == list(yt.read_table(table))
+        yt.run_sort(table, sort_by=[SortColumn("x", sort_order=SortColumn.DESCENDING)])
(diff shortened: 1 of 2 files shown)


yt/python/yt/wrapper/tests/test_operations.py
```diff
@@ -185,6 +185,11 @@ def test_sort(self):
 
         assert yt.is_sorted(table)
 
+        yt.run_sort(table, sort_by=[SortColumn("x", sort_order="descending")])
```
denvr, line 188: `sort_order=SortColumn.DESCENDING` more correctly

yt/python/yt/wrapper/tests/test_operations.py
```diff
@@ -185,6 +185,11 @@ def test_sort(self):
...
+        assert list(reversed([{"x": x, "y": y} for x, y in sorted(columns, key=lambda c: c[0])])) == list(yt.read_table(table))
```
denvr, line 189: `reversed` over `sorted` - a little bit complicated.
may be `sorted(.. reverse=True)`?

chegoryu: Thanks, now it looks better

Kontakter (approved):

PR #400, opened 2024-02-21
Cherrypicking Query Tracker ACO commits to 23.2

Krock21:
Cherrypicking Query Tracker ACO commits to 23.2
I had to manually fix conflicts in first and last commit(marked with [23.2]) because they were created on top of some refactorings

### DIFF
--- a/yt/yt/ytlib/api/native/client_queries.cpp
+++ b/yt/yt/ytlib/api/native/client_queries.cpp
@@ -173,7 +173,10 @@ NSecurityClient::ESecurityAction CheckAccessControl(
         if (userSubjects.contains(NSecurityClient::SuperusersGroupName)) {
             return NSecurityClient::ESecurityAction::Allow;
         }
-        THROW_ERROR_EXCEPTION("Error while fetching access control object %Qv", actualAccessControlObject)
+        THROW_ERROR_EXCEPTION(
+            "Error while fetching access control object queries/%v. "
+            "Please make sure it exists",
+            actualAccessControlObject)
             << aclOrError;
     }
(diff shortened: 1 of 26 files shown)


yt/yt/ytlib/api/native/client_queries.cpp
```diff
@@ -38,6 +38,319 @@ using namespace NYTree;
...
+    lookupOptions.EnablePartialResult = true;
+    std::vector keys{
+        TActiveQueryKey{.QueryId = queryId}.ToKey(rowBuffer),
+    };
+    auto asyncLookupResult = client->LookupRows(
+        tablePath,
+        TRecordDescriptor::Get()->GetNameTable(),
+        MakeSharedRange(std::move(keys), rowBuffer),
+        lookupOptions);
+    auto asyncRecord = asyncLookupResult.Apply(BIND([=] (const IUnversionedRowsetPtr& rowset) {
```
Krock21, line 127: Original commit uses `TUnversionedLookupRowsResult` from [this refactoring commit](https://github.com/ytsaurus/ytsaurus/commit/01bcd9b04eecd2aa43b311fa9ce7882a73806146). We don't bring that commit to 23.2 and manually use old version `IUnversionedRowsetPtr`

Krisha11:
Import test failed with 1 return code.To reproduce run 'ya make -t --test-type import_test' or 'ya make' and 'Y_PYTHON_ENTRY_POINT=library.python.testing.import_test.import_test:main yt-yt-tests-integration-size_l'
Found difference in signatures between YtClient and client_api.py
Error: Difference in signature for start_query

Most likely, you need to run:
```
$ cd yt/python/yt/wrapper/bin/generate_client_impl
$ ./generate_client_impls.sh
```

PR #410, opened 2024-03-01
Support for temporary tokens in yql_agent

Krisha11:
In this pull request, I am changing the authorization method in yql QT requests. Temporary tokens will be used instead of impersonation.

The algorithm for generating temporary tokens is:
1) We get all used clusters from the query (in yql_plugin)
2) Generating a temporary token on all clusters (in yql_agent)
3) Run the query (in yql_agent)

yql_plugin PR: https://github.com/ydb-platform/ydb/pull/2356

### DIFF
--- a/yt/yql/agent/yql_agent.cpp
+++ b/yt/yql/agent/yql_agent.cpp
@@ -5,6 +5,7 @@
 
 #include <library/cpp/yt/logging/backends/arcadia/backend.h>
 
+#include <yt/yt/ytlib/api/native/client.h>
 #include <yt/yt/ytlib/hive/cluster_directory.h>
 
 #include <yt/yt/client/security_client/public.h>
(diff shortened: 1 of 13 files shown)


yt/yql/plugin/bridge/plugin.cpp
```diff
@@ -169,6 +172,46 @@ class TYqlPlugin
...
+                .ContentLength = file.Content.size(),
+                .Type = file.Type,
+            });
+        }
+
+        auto* bridgeClustersResult = BridgeGetUsedClusters(
+            BridgePlugin_,
+            queryText.data(),
+            settingsString.data(),
+            settingsString.length(),
```
Alexvsalexvsalex, line 198: Why do we need pass `settings` string length here? Query text doesn't require that.

Krisha11: The value 0 of one of the enum settings fields here can be serialized as '\0'. In order not to lose half of the string, we remember its length.

https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/server/query_tracker/yql_engine.cpp#L52

It was probably possible to just use the settings conversion into EYsonFormat::Text instead of EYsonFormat::Binary in all places. But historically it was done this way

In the case of the query text, we assume that the correct query would not have any '\0' character, so we don't worry about its length

yt/yql/agent/yql_agent.cpp
```diff
@@ -32,6 +33,70 @@ const auto& Logger = YqlAgentLogger;
...
+        }
+        YT_LOG_DEBUG("Token received (User: %v, Cluster: %v)", user, cluster);
+    }
+
+    return token;
+}
+
+static TString IssueToken(const TQueryId queryId, const TString& user, const std::vector<TString>& clusters, THashMap<TString, NApi::NNative::IClientPtr>& queryClients, const TDuration& expirationTimeout)
+{
+    while (true) {
```
Krock21, line 74: Let's do it 10 times, for example, and throw an error on failure to issue token. It will be much easier to debug in case some issue happens with token generation.

Krock21 (commented):
One small comment and LGTM

Krisha11:
Shipped!

PR #418, opened 2024-03-05
HTTP proxy TLS certificate update - follow-up

koct9i:
- Move TLS context commit time into impl
  
- Update TLS certificates in http proxy control invoker

### DIFF
--- a/yt/yt/core/https/server.cpp
+++ b/yt/yt/core/https/server.cpp
@@ -30,7 +30,7 @@ class TServer
     : public IServer
 {
 public:
-    explicit TServer(IServerPtr underlying, TPeriodicExecutorPtr certificateUpdater)
+    TServer(IServerPtr underlying, TPeriodicExecutorPtr certificateUpdater)
         : Underlying_(std::move(underlying))
         , CertificateUpdater_(certificateUpdater)
     { }
@@ -101,7 +101,8 @@ static void ApplySslConfig(const TSslContextPtr&  sslContext, const TServerCrede
 IServerPtr CreateServer(
     const TServerConfigPtr& config,
     const IPollerPtr& poller,
-    const IPollerPtr& acceptor)
+    const IPollerPtr& acceptor,
+    const IInvokerPtr& controlInvoker)
 {
     auto sslContext =  New<TSslContext>();
     ApplySslConfig(sslContext, config->Credentials);
@@ -113,9 +114,10 @@ IServerPtr CreateServer(
         sslConfig->CertChain->FileName &&
         sslConfig->PrivateKey->FileName)
     {
+        YT_VERIFY(controlInvoker);
         certificateUpdater = New<TPeriodicExecutor>(
-            poller->GetInvoker(),
-            BIND([=, serverName = config->ServerName] {
+            controlInvoker,
+            BIND([=] {
                 try {
                     auto modificationTime = Max(
                         NFS::GetPathStatistics(*sslConfig->CertChain->FileName).ModificationTime,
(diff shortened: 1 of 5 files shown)


koct9i:
Fixes for https://github.com/ytsaurus/ytsaurus/pull/384

yt/yt/core/https/server.cpp
```diff
@@ -125,14 +126,18 @@ IServerPtr CreateServer(
...
                         sslContext->Reset();
                         ApplySslConfig(sslContext, sslConfig);
                         sslContext->Commit(modificationTime);
-                        YT_LOG_INFO("TLS certificates updated (ServerName: %v)", serverName);
+                        YT_LOG_INFO("TLS certificates updated (ServerName: %v)",
+                            config->ServerName);
                     }
                 } catch (const std::exception& ex) {
-                    YT_LOG_WARNING(ex, "Unexpected exception while updating TLS certificates (ServerName: %v)", serverName);
+                    YT_LOG_WARNING(ex, "Unexpected exception while updating TLS certificates (ServerName: %v)",
```
Kontakter, line 139: Let place second argument on the separate line

yt/yt/core/https/server.cpp
```diff
@@ -150,7 +156,9 @@ IServerPtr CreateServer(
 
 IServerPtr CreateServer(const TServerConfigPtr& config, const IPollerPtr& poller)
 {
-    return CreateServer(config, poller, poller);
+    YT_LOG_WARNING_IF(config->Credentials->UpdatePeriod,
+        "TLS certificate update runs in poller invoker");
```
Kontakter, line 160: Why is it WARNING?

koct9i: https://github.com/ytsaurus/ytsaurus/pull/384#discussion_r1510335021
Currently this function is used only in tests and does not enable updates.
Do you prefer assert?

yt/yt/core/https/server.cpp
```diff
@@ -150,7 +157,7 @@ IServerPtr CreateServer(
 
 IServerPtr CreateServer(const TServerConfigPtr& config, const IPollerPtr& poller)
 {
-    return CreateServer(config, poller, poller);
+    return CreateServer(config, poller, poller, nullptr);
```
gritukan, line 160: nip: I'd prefer `/*paramName*/` before nullptr

yt/yt/core/https/server.cpp
```diff
@@ -101,7 +101,8 @@ static void ApplySslConfig(const TSslContextPtr&  sslContext, const TServerCrede
 IServerPtr CreateServer(
     const TServerConfigPtr& config,
     const IPollerPtr& poller,
-    const IPollerPtr& acceptor)
+    const IPollerPtr& acceptor,
+    const IInvokerPtr& controlInvoker)
```
gritukan, line 105: Why is it called control here?

yt/yt/server/http_proxy/bootstrap.cpp
```diff
@@ -267,7 +267,7 @@ TBootstrap::TBootstrap(TProxyConfigPtr config, INodePtr configNode)
 
     if (Config_->HttpsServer) {
         Config_->HttpsServer->ServerName = "HttpsApi";
-        ApiHttpsServer_ = NHttps::CreateServer(Config_->HttpsServer, Poller_, Acceptor_);
+        ApiHttpsServer_ = NHttps::CreateServer(Config_->HttpsServer, Poller_, Acceptor_, GetControlInvoker());
```
gritukan, line 270: Typically control invoker in YT process is used for some lightweight actions that should be properly serialized with this invoker (for example, in node, complex communication with master is performed in this thread). Sometimes this actions are latency-critical, so we expect this thread not to be loaded and all actions be quite fast.

Assuming this I'm not sure that control invoker is the best choice for a thread to work with filesystem, since disk may be overloaded. Maybe something like RpcHeavy suites better for that?

yt/yt/core/https/server.cpp
koct9i: Because it is executes control flow for https server.

yt/yt/server/http_proxy/bootstrap.cpp
koct9i: HTTP-proxy executes disk-change-checker (whatever it is) in the same control-invoker.

If you want to offload all disk I/O into isolated thread then rpc-heavy seems wrong choice too.

Anyway most times certificate update ends with calling stat for couple files which metadata almost certainly will be in in cache or even in tmpfs.

gritukan: > HTTP-proxy executes disk-change-checker (whatever it is) in the same control-invoker.

Okay, in this case it's perferctly fine to leave control invoker.

yalekseev:
Shipped!

PR #419, opened 2024-03-05
[styleguide] Fix link to property.h

artsiukhou:
### What

Link to `property.h` in the C++ style guide doc is relative, so it's resolved to `https://github.com/ytsaurus/ytsaurus/blob/main/yt/styleguide/library/cpp/yt/misc/property.h` that returns 404:
image

### How

Making the link absolute, so it's resolved to `https://github.com/ytsaurus/ytsaurus/blob/main/library/cpp/yt/misc/property.h`. 

### How tested

Looking at "rich diff" and clicking at the link

### Other

### DIFF
--- a/yt/styleguide/cpp.md
+++ b/yt/styleguide/cpp.md
@@ -182,7 +182,7 @@ public:
 };
 ```
 
-See more complex property macros in [property.h](library/cpp/yt/misc/property.h).
+See more complex property macros in [property.h](/library/cpp/yt/misc/property.h).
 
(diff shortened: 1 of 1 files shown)


zlobober (approved):

artsiukhou:
@zlobober, thanks for approving the PR. I don't have write permissions to the repo, so would appreciate if you could merge it!

zlobober:
@savnadya, would you mind merging? Seems like this PR got lost.

Kontakter (approved):

PR #424, opened 2024-03-08
Fix docs language in an article for YQL syntax

rBunin:
Wrong language in Eng docs.


Kontakter (approved):
Thank you for this change!

PR #427, opened 2024-03-08
Fix QT compatibility with old proxies

gritukan:


Krock21:
Let's do it ASAP while we have not so many clusters with broken table

What actions are needed for clusters with broken tables?(6th TRANSFORMS version before this PR)?

Kontakter:
Could you give some evidence: where this field is presented in server code?

And I think we should add some test for this fix

gritukan:
> Let's do it ASAP while we have not so many clusters with broken table

Migration with this issue affects only nightly builds, it did not affect any releases, so I do not expect that there are many users with this bug. However, I can add migration 7 that will add this field (if it is missing) and do nothing (if it already exists).

> Could you give some evidence: where this field is presented in server code?

https://github.com/ytsaurus/ytsaurus/blob/stable/23.2/yt/yt/ytlib/api/native/client_queries.cpp#L457

It was present before and I removed this field from server code together with schema change. It breaks old proxies with new QT compatibility.

gritukan:
Regarding tests, the only way to test it are compat-tests, but they are not open sourced. Probably, they should be implemented internally to detect such problems in future.

cc: @Krisha11

Krisha11 (approved):

PR #438, opened 2024-03-10
Remove dead code

chegoryu:
I believe this code is dead: https://github.com/ytsaurus/ytsaurus/commit/8965c79c57ff3336293ebe92be01c124bceef85c#diff-08d4872f549f39d80f5c4f4bde5f3b78f817cbb1ccdb4f462bdb628257557bd8R165

### DIFF
--- a/yt/yt/client/ypath/parser_detail.cpp
+++ b/yt/yt/client/ypath/parser_detail.cpp
@@ -163,30 +163,6 @@ TString ParseCluster(TString str, const IAttributeDictionaryPtr& attributes)
 
     attributes->Set("cluster", clusterName);
     return remainingString;
-
-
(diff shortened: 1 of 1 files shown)


gritukan:
cc: @achulkov2 

This code became dead after your commit 8965c79c57ff3336293ebe92be01c124bceef85c. Should it be removed or moved into another place?

chegoryu:
By the way, I'm surprised that the warning for unreachable code is disabled.

As I remember it is possible to enable this warning (error with `-Werror` flag) somewhere near devtools directories.

achulkov2:
It is indeed an old code path that I rewrote into the code before the return statement and forgot to delete it after using it for debug purposes. 🤦 
Safe to remove, LGTM.

chizhonkova (approved):

PR #446, opened 2024-03-12
New README

ogorbacheva:
Editing README for the public documentation project.


yt/docs/README.md
```diff
@@ -1,208 +1,434 @@
...
+
+The `dyn-query-language.md` article also reuses content, same as `query-tracker.md`. So, as you can see in the code above:
+- The "YT QL" link refers to the article located in the `en/user-guide/dynamic-tables/` folder, but not the `en/_includes/user-guide/dynamic-tables/`.
+- This link is set relativelly to the `_includes` folder.
 
-Пользователи и группы располагаются в одном пространстве имен, то есть имена пользователей и групп не могут совпадать.
+### Headers
 
-{% endnote %}
+Headers should always have anchors:
```
Krock21, line 358: That is some strict rule that may be unnecessary. Current documentation doesn't follow it

I believe anchors may be generated automatically for all headers(implicit anchors), based on text. I saw it in some systems

Explicit anchors are worth for important parts of the document that will be linked from another document

yt/docs/README.md
```diff
@@ -1,208 +1,434 @@
...
+
+```
+## Users, Groups { #users-groups }
+```
+
+
+### Lists
+
+Every item in an ordered list starts with a capital letter and ends with a period, while items in an unordered list start with a lowercase letter and end with a semicolon; the last item in the list always ends with a period.
+
```
Krock21, line 368: Worth adding when we use ordered lists, and when we use unordered lists.
(Probably when order matters we use ordered, and unordered otherwise. Rule with semicolon is strange in this case)

yt/docs/README.md
```diff
@@ -1,208 +1,434 @@
...
+````
+
+### Tables
+
+When working with tables, keep the following in mind:
 
-## Таблицы
+- There must be a reference to the table, for example: "Table 1 shows...".
+- It is mandatory to add a table title: `<small>Table N — Title of the table`. It should start with a capital letter, without a period at the end. 
+- Periods are not needed in the table header, as well as at the end of the sentence in the right column.
```
Krock21, line 406: Does it mean periods are mandatory at the end of the sentence in the non-right column?
The example doesn't answer it since `name` or `string` are not sentences

yt/docs/README.md
```diff
@@ -211,3 +437,47 @@ yt check-permission yql write //tmp
...
+
+### Documentation Philosophy
+
+- When writing text, determine your target audience, whom you are writing for. Take into account that YTsaurus users may have different backgrounds and may be from different areas: analysts, engineers, administrators, and developers. The target audience influences the level of detail of your text and the terminology you use.
+- Your text should not disrupt the integrity of the entire document or its narrative logic. When writing content, consider the context of the section where this content is intended to be added.
+- Try to avoid jargon, slang, Anglicisms, and colloquial language. Use commonly accepted terms. Strive to follow the academic style.
+- Accompany your text with examples. Content without examples is perceived as abstract and is hard to assimilate.
+- Use cross-links between the sections.
+- Aim to write concisely and clearly. Do not make sentences too long — up to 15 words per sentence. Split long paragraphs into short ones — up to 5 sentences in one paragraph.
+- If you need any help with content writing, don't hesitate to ask questions in YTsaurus [community chats](#need-help).
```
Krock21, line 452: 1. Seems like an important paragraph that we may consider moving upwards
2. It is interesting that we break rules that we define in the same document :)
   + Maybe we should change/remove the rule?
   + Same for unordered lists above
> items in an unordered list start with a lowercase letter and end with a semicolon;

yt/docs/README.md
Krock21: One more thing is that we don't define what is a header. I thought this applies to `# ## ### #### ##### ######`, but seems like it applies to `#` and `##`, as we see in this document

yt/docs/README.md
```diff
@@ -211,3 +437,47 @@ yt check-permission yql write //tmp
...
+
+------
+
+## How to Build Locally
```
Krock21, line 456: Lets later unify it with automatic CI that should appear on github soon

yt/docs/README.md
```diff
@@ -211,3 +437,47 @@ yt check-permission yql write //tmp
...
 ```
+````
+
+------
+
+### Documentation Philosophy
+
+- When writing text, determine your target audience, whom you are writing for. Take into account that YTsaurus users may have different backgrounds and may be from different areas: analysts, engineers, administrators, and developers. The target audience influences the level of detail of your text and the terminology you use.
+- Your text should not disrupt the integrity of the entire document or its narrative logic. When writing content, consider the context of the section where this content is intended to be added.
+- Try to avoid jargon, slang, Anglicisms, and colloquial language. Use commonly accepted terms. Strive to follow the academic style.
```
achulkov2, line 448: Lets specify that Anglicisms are for Russian. Also it's a debatable point, I think we are already quite a few and have no desire to stop, since it is closer to the english-sourced naming inside YT (e.g. chunks/tablets/etc.).

yt/docs/README.md
```diff
@@ -211,3 +437,47 @@ yt check-permission yql write //tmp
...
+   ```
+4. Clone YTsaurus repository in case you haven't done it yet:
+   ```bash
+   $ git clone https://github.com/ytsaurus
+   ```
+5. Run the build:
+   ```bash
+   $ yfm -i ytsaurus/yt/docs/ -o ytsaurus-doc-output
+   ```
+6. Navigate to the `ytsaurus-doc-output` folder and then go to the `ru/` or `en/` directory, depending on the documentation language you want to check. When opening the `index.html` file, the main page of the documentation will be shown.
```
achulkov2, line 475: Screenshot 2024-03-19 at 14 41 58
When opening index.html it looks like this :(
This is very different from the rendering within Yandex and on ytsaurus.tech/docs.
How can we improve this?

yt/docs/README.md
```diff
@@ -1,208 +1,434 @@
...
+     3. In the `en/_includes/user-guide/new-article.md` file, write the text of your article. For example:
+
+         ```
+         # Deployment in a Kubernetes Cluster
+
+         This section describes the installation of YTsaurus in a Kubernetes cluster ...
+         ```
+        If your text contains links to another articles, see how to insert these links correctly in the [Links](#links) section.
+
+     > **Why this structure?** <br/> It is related to the specifics of the YTsaurus documentation build. Documentation is represented in two versions: public version (the one you are reading), and the internal one which is available only for the team members. Content of the public documentation can be reused in its internal version. Therefore, articles are basically placed in the `_includes` folder.
```
achulkov2, line 221: Basically is probably not what you meant? I think you wanted to say something like "by default"/"as a rule" etc.

yt/docs/README.md
```diff
@@ -1,208 +1,434 @@
...
-   ```
+-------
 
-   Либо делаем pull request:
+### Reusing Content
 
-   ```bash
-   ya pr create -m 'My new page'
-   ```
+The documentation is written using the Single Source approach. This means the common content that is reused across different parts of the documentation is stored in separate files and then added in the necessary articles. In YTsaurus, the shared content is stored in the `ru/_includes` folder for the Russian content, and in `en/_includes` for the English one.
```
achulkov2, line 117: At this moment in the article the purpose of the `_includes` directory is kind of unclear. Maybe we should move the explanation from "How to Add a New Article" here?

yt/docs/README.md
```diff
@@ -1,208 +1,434 @@
-{% note info %}
+# About YTsaurus Documentation
```
achulkov2, line 1: Maybe just "YTsaurus documentation"?

yt/docs/README.md
```diff
@@ -1,208 +1,434 @@
...
-   tar -xzf docs.tar.gz -C docs
-   python3 -m http.server 8888 -d docs
-   или
-   cd docs
-   python ../dev-server.py
-   ```
+-------
 
-   Либо делаем pull request:
+### Reusing Content
```
achulkov2, line 115: Reused content? Shared content?
This is not a section on how to reuse content, but rather on how our documentation is structured with a single shared source.

yt/docs/README.md
achulkov2: Or explain it in both places, since it might be somewhat confusing to an outside committer, and it won't hurt to repeat the reasoning.

achulkov2 (changes requested):

yt/docs/README.md
ogorbacheva: I would use "About" because this section is not YTsaurus documentation itself, but it is "documentation about documentation". So this title more accurately conveys the essence of the section.

yt/docs/README.md
```diff
@@ -1,208 +1,434 @@
...
+        |-- _includes/      # Common documents that are reused in different parts of the Russian documentation.
+            |-- user-guide/ # For example, here are stored common articles that are used in the user guide section.
+            |-- admin-guide/
+    |-- en/                 # Content in English.
+        |-- toc.yaml        # Table of contents of the English documentation.
+        |-- user-guide/           
+        |-- _includes/      # Common documents that are reused in different parts of the English documentation.
+            |-- user-guide/
+            |-- ...
+    |-- images/             # Folder with images.
```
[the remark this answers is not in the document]
ogorbacheva: I have added some recommendations about images at the end of the article.

savnadya:
Shipped!

PR #449, opened 2024-03-13
add debug tools to ytsaurus-odin docker image 

kruftik:


yt/odin/requirements.txt
```diff
@@ -2,6 +2,7 @@ python-dateutil==2.8.2
 Flask==2.0.3
 Werkzeug==2.2.2
 cheroot==10.0.0
+python-prctl==1.8.1
```
Kontakter, line 5: Я вроде избавлялся от необходимости зависеть от prctl, она тут точно нужна?

Просто у этой библиотеки плохая лицензия

kruftik: > как я понимаю, она же в Арке используется в завендоренном виде из контриба?

как я понимаю, она же в Арке используется в завендоренном виде из контриба?

zlobober: (Use English, please)

zlobober: cc @koct9i

Kontakter: Maybe, but it should be checked and fixed in this case. Because prctl is under GPL and I prefer to avoid using this module even as dependency installed in Docker image.

zlobober: > Maybe, but it should be checked and fixed in this case.

Can you elaborate, what do you mean here? Setting correct prctl is exactly the way of correct orphaned child reaping.

> Because prctl is under GPL and I prefer to avoid using this module even as dependency installed in Docker image.

Why so? We do include GPL dependencies in our docker images, YTsaurus one depends on libiconv, for example. Having a binary distribution of GPL dependency in form of a Docker image dependency does not violate any form of GPL, as far as we researched in 2022.

Kontakter: > Can you elaborate, what do you mean here? 
Sure, I mean slightly improve usages of prctl to avoid using this library. One way is writing  small cython-bindings for setting pdeathsig (the simplest way I think), another way is to figure out cases in which processes can leak and fixing these cases.

> Why so? We do include GPL dependencies in our docker images,
libiconv library distributed under LGPL license, it allows to use it as a dynamically linked dependency, but python-prctl distributed under GPL and as far as I know this license does not allow to use library even as a dependency.

koct9i:
Fixed here https://github.com/ytsaurus/ytsaurus/pull/496

PR #461, opened 2024-03-15
Skip tests which requires porto if it is not available

koct9i:


yt/yt/tests/library/yt_env_setup.py
```diff
@@ -491,6 +492,9 @@ def create_yt_cluster_instance(cls, index, path):
             enable_tls=cls.ENABLE_TLS,
         )
 
+        if yt_config.jobs_environment_type == "porto" and not porto_available():
+            pytest.skip("Porto not available")
```
gritukan, line 496: is not avaiable?

Kontakter (approved):

PR #466, opened 2024-03-17
Support building yt/cpp and yt/yt/core with vanilla protobuf

gritukan:
After this PR yt/cpp and yt/yt/core are possible to be built both with Arcadia protobuf (that uses TString as a string) and vanilla protobuf (that uses std::string as a string). To achieve so, a couple of interoperability primitives are introduced.

* `TProtobufString` is an alias to protobuf string type, i.e. it can be `TString` or `std::string` depending on the protobuf implementation.
* `IsVanillaProtobuf` and `IsArcadiaProtobuf` are the constexpr boolean values that allow to check protobuf implementation both in the compile time and runtime.

The most challenging interoperability issue solved here is a string copy between protobuf message and C++ code that has a form of `TString str = msg.str()`. This code works perfect with Arcadia protobuf but does not work with vanilla protobuf. To solve it, a previously introduced primitive `FromProto<TString>` is used. This expression makes the most efficient cast possible between protobuf string and C++ string. Internally, it is just a copy in both cases. Since TString is CoW by default, this expression is almost zero-cost (actually it's just one atomic operation), so no degradation is expected for YTsaurus server builds. The most hot code is handled differently to avoid even atomic operations (see `GetRequestTargetYPath`). In case of vanilla protobuf string is copied, however there are no places in C++ SDK where it might be a problem. If such issues would appear, performance-critial code can be rewritten in `GetRequestTargetYPath`-style.

### DIFF
--- a/yt/yt/core/misc/protobuf_helpers.h
+++ b/yt/yt/core/misc/protobuf_helpers.h
@@ -20,6 +20,7 @@
 #include <google/protobuf/repeated_field.h>
 
 #include <google/protobuf/io/zero_copy_stream_impl_lite.h>
+
 namespace NYT {
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/yt/yt/core/misc/protobuf_helpers.h
+++ b/yt/yt/core/misc/protobuf_helpers.h
@@ -6,6 +6,7 @@
 #include "range.h"
 #include "serialize.h"
 
+#include <google/protobuf/io/zero_copy_stream_impl_lite.h>
 #include <yt/yt/core/compression/public.h>
 
 #include <yt/yt_proto/yt/core/misc/proto/guid.pb.h>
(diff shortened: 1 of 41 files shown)


savnadya:
Shipped!

savnadya:
Shipped!

yt/yt/core/misc/protobuf_helpers.h
```diff
@@ -6,6 +6,7 @@
 #include "range.h"
 #include "serialize.h"
 
+#include <google/protobuf/io/zero_copy_stream_impl_lite.h>
```
maxim-babenko, line 9: Let's move it down and place an empty line between directories

maxim-babenko (approved):

savnadya:
Shipped!

savnadya:
Shipped!

PR #470, opened 2024-03-19
Use ya make in QT release

Krisha11:
Tested in:
https://github.com/ytsaurus/ytsaurus/actions/runs/8347041812


Krisha11:
Shipped!

PR #471, opened 2024-03-19
Fix python packages in QT build

Krisha11:


Krisha11:
Tested in 
https://github.com/ytsaurus/ytsaurus/actions/runs/8345278287

Krisha11:
Shipped!

PR #497, opened 2024-04-01
Del ya make strip in relwithdebinfo

Krisha11:
Tested in https://github.com/ytsaurus/ytsaurus/actions/runs/8485472853


Krisha11:
Shipped!

PR #505, opened 2024-04-04
Fix resolving queue path in AdvanceConsumer

savnadya:


yt/yt/client/queue_client/consumer_client.cpp
```diff
@@ -537,12 +537,19 @@ class TGenericConsumerClient
...
 
-        auto partitionRowInfos = CollectPartitionRowInfos(
-            QueueRef_->Path,
-            QueueClusterClient_,
-            std::move(tabletAndRowIndices),
-            params,
-            Logger);
+        THashMap<int, THashMap<i64, TPartitionRowInfo>> partitionRowInfos;
+        try {
+            partitionRowInfos = CollectPartitionRowInfos(
```
achulkov2, line 542: I think that we generally try to avoid try-except clauses like this in our control flow. Let's make `CollectPartitionRowInfos` return a future to its result. It should be fairly easy to rewrite the logic inside into SelectRows + Apply.

yt/yt/ytlib/queue_client/registration_manager.cpp
```diff
@@ -398,7 +424,10 @@ void TQueueConsumerRegistrationManager::GuardedRefreshReplicationTableMappingCac
 
     for (const auto& replicatedTableInfo : replicatedTableMapping) {
         for (const auto& replica : replicatedTableInfo.GetReplicas()) {
-            ReplicaToReplicatedTable_[replica] = replicatedTableInfo.Ref;
+            ReplicaToReplicatedTable_[replica] = TResolvedObject{
+                .Path = replicatedTableInfo.Ref,
+                .ObjectType = replicatedTableInfo.ObjectType,
```
achulkov2, line 429: If we implement it like this, replicas will have a type of `[Chaos]ReplicatedTable`, while the `[chaos]_replicated_table` objects themselves will have a `std::nullopt` type.

Ideally, we would want to separate these types, but that all depends on how this will be used in the end.

achulkov2 (changes requested):
I think we should separate the `ObjectType` related changes into a separate PR in which we actually use them.

yt/yt/ytlib/api/native/helpers.cpp
```diff
@@ -207,4 +215,25 @@ TError MakeRevivalError(
 
 ////////////////////////////////////////////////////////////////////////////////
 
+void CheckConsumerPermission(
```
achulkov2, line 218: Let's expose the context switch happening inside by returning a future? We can even chain this with `.Apply`.

yt/yt/ytlib/api/native/transaction.cpp
```diff
@@ -352,17 +353,21 @@ class TTransaction
...
-        auto tableMountCache = GetClient()->GetTableMountCache();
-        auto queuePhysicalPath = queuePath;
-        auto queueTableInfoOrError = WaitFor(tableMountCache->GetTableInfo(queuePath.GetPath()));
-        if (!queueTableInfoOrError.IsOK()) {
-            THROW_ERROR_EXCEPTION("Failed to resolve queue path")
-                << queueTableInfoOrError;
+        CheckConsumerPermission(Client_->GetNativeConnection(), consumerPath, Client_->GetOptions());
+
+        auto registrationCheckResult = Client_->GetNativeConnection()->GetQueueConsumerRegistrationManager()->GetRegistration(queuePath, consumerPath);
+        if (!registrationCheckResult.Registration) {
```
achulkov2, line 359: Checking registrations should also go inside the helper, IMO, since it is also exactly the same.

yt/yt/tests/integration/queues/test_consumer_registrations.py
```diff
@@ -817,6 +816,8 @@ def test_multicluster_symlink_registrations(self, create_registration_table):
 class TestDataApi(TestQueueConsumerApiBase, ReplicatedObjectBase, TestQueueAgentBase):
     NUM_TEST_PARTITIONS = 2
 
+    NUM_REMOTE_CLUSTERS = 2
```
achulkov2, line 819: Let's make a separate suite for multi-cluster API tests. The multi-cluster set-up/tear-down takes a while and would make all the other tests in the suite longer.

achulkov2 (changes requested):

achulkov2 (approved):

savnadya:
It was merged here: https://github.com/ytsaurus/ytsaurus/commit/cdc867f5fd050a1752c0a98090e7dddea5d924d6

Issue #287, opened 2023-12-26
Improve input data size estimation for operations with input ranges


gritukan:
For now, input statistics (row count, data weight, (un)compressed data size) for operations are calculated as a sum of corresponding statistics for input chunks. This is precise for the operations that take whole tables as an input and works pretty well in case if tables and ranges are quite big. However, this approach is very inaccurate for the case of big input chunks and small ranges, since for every range size of the whole chunk will be accounted.

The idea of this proposal is to add a new type of chunk fetcher called range fetcher (feel free to suggest a better naming) that improves estimates for the input data slice statistics by analogy with columnar statistics fetcher. For each chunk that is on the border of some range it fetches block meta from node and calculates the statistics of the blocks that will actually be fetched by a reader. Note, that currently block meta contains row count and uncompressed data size only, so one can add data weight and compressed data size to the chunk meta or just make an approximation `range_data_weight = chunk_data_weight * (range_row_count / chunk_row_count)` that is already used in some other parts of controllers.

chegoryu:
Very nice feature!
I’m now manually selecting the data weight coefficient so that actual input data size for jobs is reasonable.

gritukan:
High-level plan:

1) Introduce an RPC call `GetDataSliceStatistics` on the data node similar to `GetColumnarStatistics`. Its input is a set of chunks, columns and ranges. It returns approximate statistics of what will be read (total size of blocks in the projection of the chunk to provided columns and keys).
2) Implement TDataSliceStatisticsFetcher similar to columnar statistics fetcher that fetches statistics from nodes for given set of chunks.
3) Use fetcher in controller to make data size and row count estimations more precise. This feature should be opt-in (at least for the first time)

zlobober:
Rpc call should probably be named GetChunkSliceStatistics instead of GetDataSliceStatistics, because there is no notion of data slice in data node (data slice if a logical union of a number of overlapping chunk slices of different chunks). Otherwise, I'm ok with the plan

gritukan:
Yeah, you are right. It's chunk slices in relevant places of code.

gritukan:
Also, there is a room for possible optimization: if chunk is completely inside some range we call do a fallback to columnar statistics fetching which can be done from masters. However, I think it is not required for the first version.

PR #512, opened 2024-04-08
Add use_chunk_slice_statistics to implement #287

achulkov2:
Uses and improves the existing `TChunkSliceSizeFetcher` implemented for dynamic table resharding. 
The fetcher can now be supplied with a column filter for chunk slices, which will be used to only consider blocks that contain the relevant columns.

If `use_chunk_slice_statistics` is set to true for an operation, we compute approximate data weights for all slices with non-trivial limits based on block key/row bounds.

Caveats:
- Turning this feature on disables fetching columnar statistics to avoid squaring the columnar coefficient. Currently this is not very good for non-columnar tables, since currently we are accounting whole blocks and not columns. **Using this feature (for now) effectively disables columnar approximations for non-columnar tables**.
- This feature does not work well with `TChunkSliceFetcher`, since it does its own similar data weight approximation. For this reason this feature cannot be enabled for operations which use the sorted controller and is ignored for dynamic tables.
- Small slices that do not cross the boundary of any block will have an approximated data weight of **zero**. This could lead to some surprising effects and might be changed in the future by adding an additional flag to `TChunkSliceSizeFetcher` to force it to round up to the data weight of at least one block.
- `TChunkSliceSizeFetcher` approximates data weights based on block uncompressed size, which might be inaccurate for heavily-encoded columns of columnar tables. This can be improved later by storing data weights in block meta.

### DIFF
--- a/yt/yt/server/controller_agent/controllers/operation_controller_detail.cpp
+++ b/yt/yt/server/controller_agent/controllers/operation_controller_detail.cpp
@@ -70,6 +70,7 @@
 
 #include <yt/yt/ytlib/table_client/chunk_meta_extensions.h>
 #include <yt/yt/ytlib/table_client/chunk_slice_fetcher.h>
+#include <yt/yt/ytlib/table_client/chunk_slice_size_fetcher.h>
 #include <yt/yt/ytlib/table_client/columnar_statistics_fetcher.h>
 #include <yt/yt/ytlib/table_client/helpers.h>
 #include <yt/yt/ytlib/table_client/schema.h>
@@ -7343,6 +7344,11 @@ void TOperationControllerBase::InferInputRanges()
 TError TOperationControllerBase::GetAutoMergeError() const
 {
     return TError("Automatic output merge is not supported for %lv operations", OperationType);
+}
+
+TError TOperationControllerBase::GetUseChunkSliceStatisticsError() const
+{
+    return TError("Fetching chunk slice statistics is not supported for %lv operations", OperationType);
 }
 
 void TOperationControllerBase::FillPrepareResult(TOperationControllerPrepareResult* result)
--- a/yt/yt/tests/integration/controller/test_merge_operation.py
+++ b/yt/yt/tests/integration/controller/test_merge_operation.py
@@ -2376,12 +2376,17 @@ class TestSchedulerMergeCommands(YTEnvSetup):
                 "data_size_per_job": 5000,
                 "force_transform": True,
                 "use_chunk_slice_statistics": True,
-                # NB: In sorted mode this behaviour is facilitated by TChunkSliceFetcher and not by the option above.
+                # NB: In sorted mode this behavior is facilitated by TChunkSliceFetcher and not by the option above.
                 "mode": merge_mode,
             }
         )
 
         op.track()
+
+        if merge_mode == "sorted":
+            assert "use_chunk_slice_statistics_disabled" in op.get_alerts()
+        else:
+            assert "use_chunk_slice_statistics_disabled" not in op.get_alerts()
 
         assert get("//tmp/d/@chunk_count") < 5
 
@@ -2415,7 +2420,7 @@ class TestSchedulerMergeCommands(YTEnvSetup):
                 "data_size_per_job": 400,
                 "force_transform": True,
                 "use_chunk_slice_statistics": True,
-                # NB: In sorted mode this behaviour is facilitated by TChunkSliceFetcher and not by the option above.
+                # NB: In sorted mode this behavior is facilitated by TChunkSliceFetcher and not by the option above.
                 "mode": merge_mode,
             }
         )
@@ -2459,6 +2464,15 @@ class TestSchedulerMergeCommands(YTEnvSetup):
         op.track()
 
         assert get("//tmp/d/@chunk_count") == 1
+
+    @authors("achulkov2")
+    def test_plain_run_produces_no_alerts(self):
+        self._prepare_tables()
+
+        op = merge(mode="unordered", in_=[self.t1, self.t2], out="//tmp/t_out")
+        op.track()
+
+        assert not op.get_alerts()
 
 ##################################################################
 
--- a/yt/yt/tests/integration/controller/test_merge_operation.py
+++ b/yt/yt/tests/integration/controller/test_merge_operation.py
@@ -11,7 +11,7 @@ from yt_commands import (
 from yt_type_helpers import (
     make_schema, normalize_schema, normalize_schema_v3, optional_type, list_type)
 
-from yt_helpers import skip_if_no_descending, skip_if_renaming_disabled
+from yt_helpers import skip_if_no_descending, skip_if_old, skip_if_renaming_disabled
 
 from yt.environment.helpers import assert_items_equal
 from yt.common import YtError
@@ -2353,6 +2353,8 @@ class TestSchedulerMergeCommands(YTEnvSetup):
     @authors("achulkov2")
     @pytest.mark.parametrize("merge_mode", ["unordered", "ordered", "sorted"])
     def test_chunk_slice_statistics(self, merge_mode):
+        skip_if_old(self.Env, (24, 2), "use_chunk_slice_statistics is not supported in older versions")
+
         create("table", "//tmp/t", attributes={
             "optimize_for": "lookup",
             "schema": [
@@ -2393,6 +2395,8 @@ class TestSchedulerMergeCommands(YTEnvSetup):
     @authors("achulkov2")
     @pytest.mark.parametrize("merge_mode", ["unordered", "ordered", "sorted"])
     def test_chunk_slice_statistics_underestimation(self, merge_mode):
+        skip_if_old(self.Env, (24, 2), "use_chunk_slice_statistics is not supported in older versions")
+
         create("table", "//tmp/t", attributes={
             "optimize_for": "lookup",
             "schema": [
@@ -2434,6 +2438,8 @@ class TestSchedulerMergeCommands(YTEnvSetup):
     # TODO(achulkov2): Add sorted mode to check once columns are supported by TChunkSliceFetcher.
     @pytest.mark.parametrize("merge_mode", ["unordered", "ordered"])
     def test_chunk_slice_statistics_work_as_columnar_statistics(self, merge_mode):
+        skip_if_old(self.Env, (24, 2), "use_chunk_slice_statistics is not supported in older versions")
+
         create("table", "//tmp/t", attributes={
             "optimize_for": "scan",
             "schema": [
--- a/yt/yt/tests/integration/controller/test_merge_operation.py
+++ b/yt/yt/tests/integration/controller/test_merge_operation.py
@@ -18,6 +18,8 @@ from yt.common import YtError
 import yt.yson as yson
 
 import pytest
+import string
+import random
 
 from time import sleep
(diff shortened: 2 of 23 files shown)


yt/yt/tests/integration/controller/test_merge_operation.py
```diff
@@ -2272,6 +2272,40 @@ def test_mr_any_value_with_hunks(self, optimize_for):
 
         assert read_table("//tmp/t") == rows
 
+    @authors("achulkov2")
+    def test_chunk_slice_statistics(self):
```
achulkov2, line 2276: Let's test that we still throw the same error when trying to provide key bounds on non-sorted tables.

yt/yt/ytlib/chunk_client/input_chunk_slice.cpp
```diff
@@ -464,7 +464,7 @@ TInputChunkSlice::TInputChunkSlice(
 
     if (protoChunkSlice.has_row_count_override() || protoChunkSlice.has_data_weight_override()) {
         YT_VERIFY(protoChunkSlice.has_row_count_override() && protoChunkSlice.has_data_weight_override());
-        OverrideSize(protoChunkSlice.row_count_override(), std::max<i64>(1, protoChunkSlice.data_weight_override() * chunkSlice.GetInputChunk()->GetColumnSelectivityFactor()));
+        OverrideSize(protoChunkSlice.row_count_override(), std::max<i64>(1, protoChunkSlice.data_weight_override() * chunkSlice.GetInputChunk()->GetSelectivityFactor()));
```
achulkov2, line 467: I think using both factors here might break [this](https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/ytlib/table_client/chunk_slice_fetcher.cpp#L326) usage within `TChunkSliceFetcher`. 
The chunk slice fetcher computes the override data weight as `slice_row_count * chunk_data_weight_per_row`, which already approximates the real number of rows inside the slice.

achulkov2: Disabled chunk slice statistics for operations and tables that use `TChunkSliceFetcher`.

yt/yt/ytlib/chunk_client/input_chunk.h
```diff
@@ -67,7 +67,6 @@ class TInputChunkBase
     DEFINE_BYVAL_RW_PROPERTY(double, ColumnSelectivityFactor, 1.0);
 
     DEFINE_BYVAL_RO_PROPERTY(bool, StripedErasure, false);
-
```
gritukan, line 70: Let's keep it

yt/yt/ytlib/chunk_client/input_chunk.h
```diff
@@ -140,10 +144,12 @@ class TInputChunk
     i64 GetUncompressedDataSize() const;
     i64 GetCompressedDataSize() const;
 
+    double GetSelectivityFactor() const;
```
gritukan, line 147: Let's add some comments about what is GetRowCount(), GetDataWeight(), etc (when is it exact and when approximate). Also, let's comment what is selectivity factor and write somewhere difference between block selectivity factor, column selectivity factor and row selectivity factor.

yt/yt/ytlib/scheduler/config.h
```diff
@@ -1012,6 +1012,11 @@ class TOperationSpecBase
     //! Note that turning this on may significantly affect workload partitioning for existing operations.
     bool UseColumnarStatistics;
 
+    //! If true, enables more accurate job size estimation for operations on tables with key-bounds.
```
gritukan, line 1015: key bounds

yt/yt/ytlib/table_client/chunk_slice.cpp
```diff
@@ -406,15 +406,59 @@ void ToProto(
...
+            nameTable = FromProto<TNameTablePtr>(*nameTableExt);
+        } else if (auto schemaExt = FindProtoExtension<TTableSchemaExt>(chunkMeta.extensions())) {
+            nameTable = TNameTable::FromSchemaStable(FromProto<TTableSchema>(*schemaExt));
+        } else {
+            return {};
+        }
+
+        THashSet<int> blockFilter;
+
+        auto columnMetaExt = FindProtoExtension<TColumnMetaExt>(chunkMeta.extensions());
```
gritukan, line 429: What happens if chunk is not columnar here?

yt/yt/ytlib/table_client/chunk_slice_size_fetcher.cpp
```diff
@@ -63,6 +71,9 @@ TFuture<void> TChunkSliceSizeFetcher::DoFetchFromNode(
     req->SetResponseHeavy(true);
     req->SetMultiplexingBand(EMultiplexingBand::Heavy);
 
+    // We use this name table to replace all column names with their ids across the whole rpc request message.
+    TNameTablePtr nameTable = New<TNameTable>();
```
gritukan, line 75: auto?

yt/yt/server/controller_agent/controllers/operation_controller_detail.cpp
```diff
@@ -6174,6 +6214,17 @@ void TOperationControllerBase::FetchInputTables()
         columnarStatisticsFetcher->ApplyColumnSelectivityFactors();
     }
 
+    if (chunkSliceSizeFetcher && chunkSliceSizeFetcher->GetChunkCount() > 0) {
+        YT_LOG_INFO("Fetching input chunk slice statistics for input tables (ChunkCount: %v)",
+            chunkSliceSizeFetcher->GetChunkCount());
+        // TODO(achulkov2): Do we need a cancellation stage similar to columnar statistics fetch? It is a testing option.
```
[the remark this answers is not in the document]
achulkov2: This one: https://github.com/ytsaurus/ytsaurus/blob/c65a28cd940b2b0759fba14ae98fff996c9a34b9/yt/yt/tests/integration/controller/test_columnar_statistics.py#L700.
But I will just drop the comment. No test — no option.

yt/yt/ytlib/chunk_client/input_chunk.h
achulkov2: Added a bunch of comments.

yt/yt/ytlib/table_client/chunk_slice.cpp
achulkov2: Nothing good :)
I think I intended to have an if after `FindProtoExtension` here. Added it.

Meanwhile, I realized another caveat: the way we are accounting for columns is not exactly the same as with columnar statistics. If I have a block that contains the columns `a`, `b` and `c` in any format and I am only reading column `a`, the columnar statistics factor will give me an approximation on the weight of the single column `a`, whereas the current logic will give me the size of the whole block. The worst case scenario here happens for regular non-columnar tables, where by using the new option option we will lose the benefits we had with columnar statistics.

The annoying part is that for columnar tables we either have to rely on columnar statistics completely (which might be quite inexact), without matching block by columns, or perform additional logic which would approximate only the relevant parts of each block by also looking into columnar statistics. 
For non-columnar tables we could fallback and **use** columnar statistics alongside chunk slice size statistics, but it seems very inconsistent with explicitly **not** using it for columnar tables. 

Thus, I decided to solidify the fact that block selectivity factors override column selectivity factors. See the new comments and logic in `GetSelectivityFactor` and `ApplySelectivityFactors`.
In the future we can improve `TChunkSliceSizeFetcher` to handle column selection better. Maybe it can be generalized across different fetchers and used inside the regular columnar statistics fetcher and inside  `TChunkSliceFetcher` as well.

But it should be noted that in some cases `use_chunk_slice_statistics = true` for non-columnar tables might produce different ("worse") job partitioning than `use_columnar_statistics = true`.

The curious part of all of this is that the jobs do actually read blocks as a whole :)

yt/yt/ytlib/chunk_client/input_chunk.h
```diff
@@ -135,15 +141,33 @@ class TInputChunk
...
+    //! in lower/upper limits and the input chunks selectivity factors (see next paragraph).
+    //! Columnar factors are considered for all formats, i.e. we approximate the logical data weight bounded by all selectors.
+    //!
+    //! The data weight reduction logic is performed as follows:
+    //!   1. If `BlockSelectivityFactor` is set, we assume that it single-handedly accounts for the data weight that will be read.
+    //!   2. Otherwise, we compute the row selectivity factor as the proportion of rows bounded by limits out of all the rows in the chunk
+    //!      and use it alongisde `ColumnSelectivityFactor`.
     i64 GetDataWeight() const;
+    //! Same as above, but for uncompressed data size.
+    //! Columnar factors are only considered for columnar formats, i.e. we approximate the uncompressed data size that needs to be read from disk.
```
sergey-v-galtsev, line 157: IMHO
`i.e. we approximate the uncompressed data size that needs to be read from disk.` -> `i.e. we approximate the uncompressed size of data that needs to be read from disk.`
Or else it looks like we read uncompressed data from disk.

yt/yt/ytlib/chunk_client/input_chunk.h
```diff
@@ -135,15 +141,33 @@ class TInputChunk
...
     i64 GetUncompressedDataSize() const;
+    //! Same as above, but for uncompressed data size.
+    //! Columnar factors are only considered for columnar formats, i.e. we approximate the compressed data size that needs to be read from disk.
```
sergey-v-galtsev, line 160: `i.e. we approximate the compressed data size that needs to be read from disk.` -> `i.e. we approximate the compressed data size that is needed to be read from disk.`

yt/yt/ytlib/chunk_client/proto/data_node_service.proto
```diff
@@ -423,10 +423,14 @@ message TReqGetChunkSliceDataWeights
         required NYT.NProto.TGuid chunk_id = 1;
         optional TReadLimit lower_limit = 2;
         optional TReadLimit upper_limit = 3;
+        optional NTableClient.NProto.TColumnFilter column_filter = 4;
     }
 
     repeated TChunkSlice chunk_requests = 1;
     required NYT.NProto.TWorkloadDescriptor workload_descriptor = 4;
+    // Name table for mapping column ids to column names across the whole request.
+    // Required if any column filters are set.
```
sergey-v-galtsev, line 432: `Required if any column filters are set.` -> `Required if any column filter is set.`

yt/yt/ytlib/table_client/chunk_slice.cpp
```diff
@@ -3,12 +3,12 @@
 #include "private.h"
 #include "chunk_meta_extensions.h"
 #include "key_set.h"
-
```
sergey-v-galtsev, line 6: Style, IMHO: I would leave the empty line here.

yt/yt/ytlib/chunk_client/input_chunk.h
sergey-v-galtsev: Also `data that needs to be read` -> `data that is needed to be read`.

sergey-v-galtsev:
Shipped!

sergey-v-galtsev:
Shipped!

sergey-v-galtsev:
Shipped!

sergey-v-galtsev:
Shipped!

sergey-v-galtsev:
Shipped!

PR #532, opened 2024-04-15
Removing and adding a chunk to an incremental master heartbeat for an offline node.

don-dron:
At the moment, there is a problem of leakage of chunk changes in the incremental heartbeat when changing chunks while sending the full heartbeat.

See OnChunkRemoved, OnChunkAdded and OnChunkMediumChanged methods of MasterConnector of data node.

### DIFF
--- a/yt/yt/server/node/data_node/config.cpp
+++ b/yt/yt/server/node/data_node/config.cpp
@@ -419,6 +419,8 @@ void TMasterConnectorDynamicConfig::Register(TRegistrar registrar)
         .Default(1000000);
     registrar.Parameter("enable_profiling", &TThis::EnableProfiling)
         .Default(false);
+    registrar.Parameter("test_location_disable_during_full_heartbeat", &TThis::TestLocationDisableDuringFullheartbeat)
+        .Default();
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/yt/yt/server/node/data_node/master_connector.cpp
+++ b/yt/yt/server/node/data_node/master_connector.cpp
@@ -334,9 +334,42 @@ public:
         YT_VERIFY(delta->State == EMasterConnectorState::Registered);
 
         delta->State = EMasterConnectorState::Online;
-        YT_VERIFY(delta->AddedSinceLastSuccess.empty());
-        YT_VERIFY(delta->RemovedSinceLastSuccess.empty());
-        YT_VERIFY(delta->ChangedMediumSinceLastSuccess.empty());
+
+        {
+            YT_VERIFY(delta->AddedSinceLastSuccess.empty());
+            YT_VERIFY(delta->RemovedSinceLastSuccess.empty());
+            YT_VERIFY(delta->ChangedMediumSinceLastSuccess.empty());
+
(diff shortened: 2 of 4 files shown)


yt/yt/server/node/data_node/master_connector.cpp
```diff
@@ -334,9 +334,42 @@ class TMasterConnector
         YT_VERIFY(delta->State == EMasterConnectorState::Registered);
 
         delta->State = EMasterConnectorState::Online;
-        YT_VERIFY(delta->AddedSinceLastSuccess.empty());
-        YT_VERIFY(delta->RemovedSinceLastSuccess.empty());
-        YT_VERIFY(delta->ChangedMediumSinceLastSuccess.empty());
+
+        {
```
gritukan, line 338: Why do we need this ident?

yt/yt/server/node/data_node/config.cpp
```diff
@@ -419,6 +419,8 @@ void TMasterConnectorDynamicConfig::Register(TRegistrar registrar)
         .Default(1000000);
     registrar.Parameter("enable_profiling", &TThis::EnableProfiling)
         .Default(false);
+    registrar.Parameter("test_location_disable_during_full_heartbeat", &TThis::TestLocationDisableDuringFullheartbeat)
```
gritukan, line 422: location_id_to_disable_during_full_heartbeat?

yt/yt/server/node/data_node/master_connector.cpp
```diff
@@ -334,9 +334,42 @@ class TMasterConnector
...
+
+            delta->AddedSinceLastSuccess = delta->AddedSinceLastFullHeartbeatTry;
+            delta->RemovedSinceLastSuccess = delta->RemovedSinceLastFullHeartbeatTry;
+            delta->ChangedMediumSinceLastSuccess = delta->ChangedMediumSinceLastFullHeartbeatTry;
+
+            if (!delta->AddedSinceLastSuccess.empty() ||
+                !delta->RemovedSinceLastSuccess.empty() ||
+                !delta->ChangedMediumSinceLastSuccess.empty())
+            {
+                YT_LOG_INFO(
```
gritukan, line 351: Won't it be too many log messages in case of location disable?

yt/yt/server/node/data_node/master_connector.cpp
```diff
@@ -926,7 +982,19 @@ class TMasterConnector
         YT_LOG_INFO("Sending full data node heartbeat to master (CellTag: %v)",
             cellTag);
 
+        auto dynamicConfig = GetDynamicConfig();
```
gritukan, line 985: This variable is used only once, no need to store it

yt/yt/server/node/data_node/master_connector.cpp
don-dron: It shouldn't, it's only when sending full heartbeats. Maximum number of chunk IDs in the log is 3&

yt/yt/server/node/data_node/master_connector.cpp
don-dron: For readability only.

PR #535, opened 2024-04-16
Fix maven assembly plugin configuration

mamogaaa:


alextokarew:
Partly implemented in another commits, partly declined because of removal of Spark fork

PR #540, opened 2024-04-19
Add ytsaurus core image ya package configuration

achulkov2:
Adds two configurations for `ya package`:
- build everything from source, including python binaries with local python.
- build only server binaries and override an existing base image.

This allows producing `ya`-based ytsaurus core images. See usage examples in `README.md` files.

Tested these images using e2e tests from our k8s operator.

### DIFF
--- a/yt/docker/ya-build/README.md
+++ b/yt/docker/ya-build/README.md
@@ -1,0 +1,5 @@
+# Building ytsaurus images with ya make
+
+This directory contains configurations for use with the `ya package` command.
+The `ya` tool is special build instrument which can be found in the root of the repository.
+See individual README files for the specifics of each configuration and `ya package` usage instructions.
--- a/yt/docker/ya-build/ya.make
+++ b/yt/docker/ya-build/ya.make
@@ -1,0 +1,4 @@
+RECURSE(
+    ytsaurus
+    ytsaurus-server-override
+)
--- a/yt/docker/ya-build/ytsaurus-server-override/Dockerfile
+++ b/yt/docker/ya-build/ytsaurus-server-override/Dockerfile
@@ -1,0 +1,10 @@
+ARG BASE_IMAGE="dev"
+FROM ytsaurus/ytsaurus:${BASE_IMAGE}
+
+USER root
+
+# Override binaries built from source.
+COPY ./ytserver-all /usr/bin/ytserver-all
+COPY ./init_queue_agent_state.py /usr/bin/init_queue_agent_state
+COPY ./init_operations_archive.py /usr/bin/init_operations_archive
+COPY ./chyt-controller /usr/bin/chyt-controller
--- a/yt/docker/ya-build/ytsaurus-server-override/README.md
+++ b/yt/docker/ya-build/ytsaurus-server-override/README.md
@@ -1,0 +1,12 @@
+# Building the ytsaurus core image by overriding server binaries
+
+This configuration allows overriding the server binaries in an existing `ytsaurus` core image by local ones built from source.
+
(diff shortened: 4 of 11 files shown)


yt/opensource-packages/ytsaurus/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+    ln -s /usr/bin/ytserver-all /usr/bin/ytserver-http-proxy && \
+    ln -s /usr/bin/ytserver-all /usr/bin/ytserver-tools && \
+    ln -s /usr/bin/ytserver-all /usr/bin/ytserver-scheduler && \
+    ln -s /usr/bin/ytserver-all /usr/bin/ytserver-controller-agent && \
+    ln -s /usr/bin/ytserver-all /usr/bin/ytserver-query-tracker && \
+    ln -s /usr/bin/ytserver-all /usr/bin/ytserver-queue-agent && \
+    ln -s /usr/bin/ytserver-all /usr/bin/ytserver-tcp-proxy
+
+COPY ./init_queue_agent_state.py /usr/bin/init_queue_agent_state
+COPY ./init_operations_archive.py /usr/bin/init_operations_archive
```
Krisha11, line 74: Why is there no `init_query_tracker_state` here?

achulkov2: I didn't add it because it isn't present in the main image either: https://github.com/ytsaurus/ytsaurus/blob/main/yt/docker/ytsaurus/Dockerfile#L29. 
I gathered that's because we have a separate query tracker image.

Surprisingly, we do create a `ytserver-query-tracker` symlink, both in the main image and in this PR. Seems like something redundant.

yt/opensource-packages/ytsaurus/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+    && unzip protoc.zip -d /usr/local \
+    && rm protoc.zip
+
+RUN mkdir -p ${SOURCE_ROOT}
+COPY ./yt/python ${SOURCE_ROOT}/yt/python
+COPY ./yt/yt/python ${SOURCE_ROOT}/yt/yt/python
+COPY ./certs/cacert.pem ${SOURCE_ROOT}/certs/cacert.pem
+
+RUN mkdir -p ${BUILD_ROOT}
+COPY ./artefacts/libyson_lib.so ${BUILD_ROOT}/yt/yt/python/yson_shared/libyson_lib.so
```
Kontakter, line 39: I like British English, but artifact looks more commonly used by programmers.

yt/opensource-packages/ytsaurus/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+COPY ./artefacts/libdriver_lib.so ${BUILD_ROOT}/yt/yt/python/driver/native_shared/libdriver_lib.so
+COPY ./artefacts/libdriver_rpc_lib.so ${BUILD_ROOT}/yt/yt/python/driver/rpc_shared/libdriver_rpc_lib.so
+
+RUN python3.11 -m venv package-env
+RUN . package-env/bin/activate; \
```
Kontakter, line 44: Use `source` instead of `.` – it gives same effect, but more verbose.

Kontakter, line 44: I am not familiar with Dockerfile semantics. Do these changes of environment are visible in latter RUN commands?

achulkov2: They aren't, so we have to do all of this within one RUN command.

Kontakter (approved):

Kontakter (changes requested):

Kontakter (approved):

PR #546, opened 2024-04-24
Fix maven assembly #2

mamogaaa:


alextokarew (changes requested):
TarArchiverPlugin willbe removed soon, so we don't accept any modifications to it.

PR #550, opened 2024-04-25
Always retry chained docker image pull

koct9i:
Without unused "error" argument retry does not happens.

Fixes: c437f3bb48e ("Cosmetics")


koct9i:
Please apply into stable/23.2 stable/24.1

Kontakter (approved):

PR #556, opened 2024-04-29
Add configuration for retryable errors for CRI executor

koct9i:
Retry after error "create containerd task" which should be transient and
more likely caused by well-known race in containerd.

Link: https://github.com/containerd/containerd/issues/9160

### DIFF
--- a/yt/yt/library/containers/cri/config.cpp
+++ b/yt/yt/library/containers/cri/config.cpp
@@ -24,6 +24,14 @@ void TCriExecutorConfig::Register(TRegistrar registrar)
 
     registrar.Parameter("cpu_period", &TThis::CpuPeriod)
         .Default(TDuration::MilliSeconds(100));
+
+    registrar.Parameter("retry_error_prefixes", &TThis::RetryErrorPrefixes)
+        .Default({
+            // https://github.com/containerd/containerd/pull/9565
+            "server is not initialized yet",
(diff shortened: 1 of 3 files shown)


yt/yt/library/containers/cri/config.cpp
```diff
@@ -24,6 +24,12 @@ void TCriExecutorConfig::Register(TRegistrar registrar)
 
     registrar.Parameter("cpu_period", &TThis::CpuPeriod)
         .Default(TDuration::MilliSeconds(100));
+
+    registrar.Parameter("retry_error_prefixes", &TThis::RetryErrorPrefixes)
+        .Default({
+            "server is not initialized yet",
+            "failed to create containerd task",
```
Kontakter, line 31: Let put in the comment a link that points to concrete message from CRI codebase that we suppose to retry

For example: https://github.com/containerd/containerd/blob/53160fb4b69ea0a82fa9a050138f8b94a0927942/internal/cri/server/container_start.go#L123

Kontakter (changes requested):
See my comment

yt/yt/library/containers/cri/config.cpp
koct9i: Ok, but links to issue or pr is more useful.

Actually there should be matching against more detailed error including nested errors.

Kontakter (approved):

PR #566, opened 2024-05-02
Fix typo: comitted, commited -> committed

chegoryu:
I left only these typos:

```cpp

# Build.
build/scripts/link_sbom.py
61:        res["properties"].append({'name': 'has_uncommited_changes', 'value': True})

# Contrib.
contrib/clickhouse/src/Coordination/KeeperLogStore.h
19:    /// Read log storage from filesystem starting from last_commited_log_index
20:    void init(uint64_t last_commited_log_index, uint64_t logs_to_keep);

contrib/clickhouse/src/Coordination/KeeperStateManager.h
36:    void loadLogStore(uint64_t last_commited_index, uint64_t logs_to_keep);

contrib/clickhouse/src/Coordination/Changelog.h
100:    void readChangelogAndInitWriter(uint64_t last_commited_log_index, uint64_t logs_to_keep);

contrib/clickhouse/src/Databases/DatabaseReplicatedSettings.h
13:    M(UInt64, wait_entry_commited_timeout_sec, 3600, "Replicas will try to cancel query if timeout exceed, but initiator host has not executed it yet", 0) \

contrib/clickhouse/src/Databases/DatabaseReplicatedWorker.cpp
337:        size_t max_iterations = database->db_settings.wait_entry_commited_timeout_sec;

contrib/python/pytest-benchmark/pytest_benchmark/utils.py
77:        parts.append("uncommited-changes")

contrib/libs/poco/Data/include/Poco/Data/Transaction.h
57:             /// commited automatically. If no error occurs, rollback is disabled and does
85:             /// Rolls back the current database transaction if it has not been commited

contrib/clickhouse/src/Storages/StorageMergeTree.cpp
2061:            ///  and we should be able to rollback already added (Precomitted) parts


# Kinda contrib.
yt/spark/spark/sql/core/src/test/scala/org/apache/spark/sql/execution/streaming/state/StateStoreSuite.scala
1048:    // Reload the store from the commited version and repeat the above test.

# Metrics tag :(
yt/yt/server/master/security_server/security_manager.cpp
595:                    TWithTagGuard guard(&buffer, "status", "commited");
599:                    TWithTagGuard guard(&buffer, "status", "uncommited");

```

The only backwards incompatible place is:

https://github.com/ytsaurus/ytsaurus/blob/091bcf82fe4fc8d9a967eb36beddf5767e92e659/yt/python/yt/wrapper/dynamic_table_commands.py#L26-L27


Kontakter (approved):
Approve for running tests in internal CI

PR #569, opened 2024-05-03
Add YYYY-MM-DD commit date into "nightly" docker image tag

koct9i:
Date of last commit is much more useful than obscure count of commits.
And it does not require full depth checkout.


savnadya:
Shipped!

PR #570, opened 2024-05-03
[stable/23.2] Add YYYY-MM-DD commit date into docker image tags

koct9i:


.github/workflows/ytsaurus_release.yaml
```diff
@@ -75,6 +75,7 @@ jobs:
       runs-on-value: ${{ needs.start-vm.outputs.label }}
       publish-ytsaurus: true
       image-tag: ${{ needs.tag.outputs.image-tag }}
+      nightly-image-tag: ${{ needs.tag.outputs.image-tag }}
```
savnadya, line 78: Let's make it optional, and don't pass it here (only in cpp.yaml)

Issue #558, opened 2024-04-29
Pickling: respawning


zlobober:
Assume the user code is designed to work with some fixed docker image, e.g. ubuntu-22.04-py3. In order to use that, currently one needs to run smth like `yt.run_map(..., spec={"mapper": {"docker_image": "ubuntu-22.04-py3"}})`.

This does not ensure that the host that runs this run_map is actually using the same Python at the same version of Ubuntu; failure to use similar environment would result in failure during unpickling.

Let's introduce a helper, e.g. `yt.use_docker_image("ubuntu-22.04-py3")`, which is to be invoked immediately at the beginning of a python file. As a result, it must do following things:
- parent process spawns a child docker container (run on the local machine) with self as an entrypoint and exposing the full file system + host network + some env variable a-la YT_RESPAWNED_IN_CONTAINER=1
- the child process runs past the yt.use_docker_image because YT_RESPAWNED_IN_CONTAINER=1 and executes the rest of the script
- all operation invocations like yt.run_map automatically assume the specified docker image

NB: there is no pickling between the parent and the child processes, since the respawning should happen right at the beginning of a script. If user runs "use_docker_image" later, we do not guarantee the correct behavior.

PR #572, opened 2024-05-03
Pickling: respawning

thenno:
resolves https://github.com/ytsaurus/ytsaurus/issues/558


yt/python/yt/wrapper/py_wrapper.py
```diff
@@ -756,3 +759,184 @@ def with_formats(func, input_format=None, output_format=None):
...
+            *docker_mount_args,
+            self._image,
+            "python3", self._main_script_path,
+        ]
+        return command
+
+    def run(self):
+        command = self.make_command()
+        logger.info("Start respawn in docker")
+        logger.info("Command: " + " ".join(command))
```
Kontakter, line 873: Let mege this message with previous one:
```
logger.info("Respawn in docker (command: ...)")
```

yt/python/yt/wrapper/py_wrapper.py
```diff
@@ -756,3 +759,184 @@ def with_formats(func, input_format=None, output_format=None):
...
+        process = subprocess.Popen(
+            command,
+            stdout=sys.stdout,
```
Kontakter, line 876: `stdin` is forgotten?

yt/python/yt/wrapper/py_wrapper.py
```diff
@@ -756,3 +759,184 @@ def with_formats(func, input_format=None, output_format=None):
...
+            stderr=sys.stderr,
+            bufsize=1,
+            universal_newlines=True,
```
Kontakter, line 879: What the idea is behind specifying these options?

```
bufsize=1,
universal_newlines=True,
```

yt/python/yt/wrapper/py_wrapper.py
```diff
@@ -756,3 +759,184 @@ def with_formats(func, input_format=None, output_format=None):
...
+        self._image = image
+        self._platform = target_platform
+        self._docker = docker
+        self._env = env
+        if self._env is None:
+            # by default, we forward only YT-specific variables
+            self._env = {
+                key: value
+                for key, value in os.environ.items()
+                if key.startswith('YT_')
```
Kontakter, line 788: We usually use double quotes

yt/python/yt/wrapper/py_wrapper.py
```diff
@@ -756,3 +759,184 @@ def with_formats(func, input_format=None, output_format=None):
...
+            }
+        self._main_script_path = main_scipt_path if main_scipt_path is not None else os.path.abspath(sys.argv[0])
+        self._cwd = cwd if cwd is not None else os.path.abspath(os.getcwd())
+        self._homedir = homedir if homedir is not None else os.path.expanduser("~")
```
Kontakter, line 792: Let use `get_home_dir` here: https://github.com/ytsaurus/ytsaurus/blob/main/yt/python/yt/wrapper/common.py#L279

yt/python/yt/wrapper/py_wrapper.py
```diff
@@ -756,3 +759,184 @@ def with_formats(func, input_format=None, output_format=None):
...
+        # we need to provide all python libraries to a container to ensure picking works
```
Kontakter, line 793: Comments should be written as a normal sentences, I mean it should start with capital letter and have a dot at the end.

Kontakter, line 793: pickLing

yt/python/yt/wrapper/py_wrapper.py
```diff
@@ -756,3 +759,184 @@ def with_formats(func, input_format=None, output_format=None):
...
+            python_lib_paths
+            if python_lib_paths is not None else
+            [os.path.abspath(path) for path in sys.path]
+        )
+        self._mount = mount
+        self._need_sudo = need_sudo
+
+    def _make_docker_env_args(self, env):
+        return list(itertools.chain(*[
+            ("-e", "{0}={1}".format(key, value))
```
Kontakter, line 804: Let use `chain.from_iterable` here and in other similar situations.

Kontakter, line 804: Is it safe to pass env variables as `"{0}={1}"`? What if it contains white space or some other special characters?

yt/python/yt/wrapper/py_wrapper.py
```diff
@@ -756,3 +759,184 @@ def with_formats(func, input_format=None, output_format=None):
...
+            for path in paths
+        ]))
+
+    def _make_mount_paths(self):
+        # use user provided values
+        if self._mount is not None:
+            return self._mount
+        # if we mount the entire local file system in docker under an arbitrary name
+        # we will break the symlinks,
+        # so we have to find only the paths necessary for work and mount them along the same paths
```
Kontakter, line 820: Let fix style of comment

yt/python/yt/wrapper/py_wrapper.py
```diff
@@ -756,3 +759,184 @@ def with_formats(func, input_format=None, output_format=None):
...
+        ]))
+
+    def _make_docker_mount_args(self, paths):
+        return list(itertools.chain(*[
+            ("-v", "{0}:{0}".format(path))
+            for path in paths
+        ]))
+
+    def _make_mount_paths(self):
+        # use user provided values
```
Kontakter, line 815: Let drop this comment, in my opinion it gives no additional information that is not expressed in the code

yt/python/yt/wrapper/py_wrapper.py
thenno: It is a very good point.
There is a similar problem not only with environment variables, but also with other values (e.g. mount points).

So I'll add escaping for the generated command.

yt/python/yt/wrapper/py_wrapper.py
thenno: Discussed in person.

get_home_dir returns None if it can't get homedir.
But here we should work only if we CAN get it. The ability to mount the home directory is our assumption for the normal work of a user's script. Otherwise a user can set mount points in params.

yt/python/yt/wrapper/py_wrapper.py
thenno: This is an old artifact from the time I tried to process stdout/stderr.
I'll just delete it.

yt/python/yt/wrapper/py_wrapper.py
thenno: I couldn't think of a single use case for using stdin in a user's script.

It's not difficult to add, but I had to make a workaround for pytest to work https://github.com/ytsaurus/ytsaurus/pull/572/files#diff-6dbdc11b5fea85a267a31195b9d207dd30e701ad8994db47661e32604a590f72R175

Kontakter (approved):

yt/python/yt/wrapper/tests/test_docker_respawn.py
```diff
@@ -0,0 +1,190 @@
+import tempfile
+
+from .conftest import authors
+
+from yt.wrapper.py_wrapper import DockerRespawner, respawn_in_docker
+
+
+@authors("thenno")
+def test_docker_respawner(monkeypatch):
```
Kontakter, line 9: Let add `pytest.skip()` for case of ya make

yt/python/yt/wrapper/tests/test_docker_respawn.py
```diff
@@ -0,0 +1,190 @@
+import tempfile
+
+from .conftest import authors
+
+from yt.wrapper.py_wrapper import DockerRespawner, respawn_in_docker
```
Kontakter, line 5: Let use import from `yt.wrapper`

yt/python/yt/wrapper/tests/test_docker_respawn.py
```diff
@@ -0,0 +1,190 @@
...
+        "-it",
+        "--rm",
+        "-e", "CWD=/root",
+        "-e", "PYTHONPATH=/usr/lib/python4.2/site-packages:/home/user/.venv/",
+        "-e", "YT_RESPAWNED_IN_CONTAINER=1",
+        "-e", "YT_SOME_KEY=YT_SOME_VALUE",
+        "-e", "YT_BASE_LAYER=some_image",
+        "-v", "/home/user:/home/user",  # user's homedir
+        "-v", "/home/user2/yt:/home/user2/yt",  # main script's dir
+        "-v", "/root:/root",  # current cwd
```
Kontakter, line 40: Let here (and in other places in this file) fix the comments:
* Write it on the previous line.
* Write it as `# Specifying user's homedir.`

yt/python/yt/wrapper/tests/test_docker_respawn.py
```diff
@@ -0,0 +1,190 @@
...
+        # a part of pythonpython outside homedir
```
Kontakter, line 41: pythonpython?

Kontakter (changes requested):
See comments above

Kontakter (approved):

Issue #222, opened 2023-12-06
ytsaurus-client warns about incompatible chardet & charset_normalizer on modern Ubuntu


zlobober:
On Ubuntu 23.10 ytsaurus-client reports the followin warning whenever yt.wrapper is imported:
```
$ yt
/home/max/.local/lib/python3.11/site-packages/yt/packages/requests/__init__.py:115: RequestsDependencyWarning: urllib3 (1.26.9) or chardet (5.1.0)/charset_normalizer (3.3.2) doesn't match a supported version!
  warnings.warn("urllib3 ({}) or chardet ({})/charset_normalizer ({}) doesn't match a supported "
usage: yt [-h] [--proxy PROXY] [--prefix PREFIX] [--config CONFIG] [--tx TX] [--master-cell-id MASTER_CELL_ID]
          [--ping-ancestor-txs] [--trace] [--version]
          command ...
yt: error: the following arguments are required: command
```

I suspect the following logic to be broken (or at least not be ready for fresh library versions):
https://github.com/ytsaurus/ytsaurus/blob/main/yt/python/contrib/python-requests/requests/__init__.py#L51-L98

ytsaurus-client does not vendor charset_normalizer nor chardet, so they are taken from the host environment. At the same time, ytsaurus-client fixes the charset-normalizer version >=3.3.0, which contradicts the assertion from the check above.

https://github.com/ytsaurus/ytsaurus/blob/e11de159f4964a42fd33380a0a718ac9a2bafb72/yt/python/packages/ytsaurus-client/setup.py#L49

Finally, ytsaurus-client does not have a dependency on chardet, but it somehow appeared in my system (possibly, as a dependency of some other package), and it is also of much fresher version (5.1.0), which also does not fit into the requirements of the code above.

I am not sure, which one of these two libraries is actually used, but my YT client works perfectly fine for a couple of months of active use, so I'd suggest relaxing these checks one way or another. At the very least, two statements for charset_normalizer should be contradicting.

Reproduce this issue by running the following command:
```
$ docker run ubuntu:mantic bash -c "apt update; apt install -y python3 python3-pip; pip3 install ytsaurus-client --break-system-packages; yt"
...
/usr/local/lib/python3.11/dist-packages/yt/packages/requests/__init__.py:115: RequestsDependencyWarning: urllib3 (1.26.9) or chardet (None)/charset_normalizer (3.3.2) doesn't match a supported version!
[...]

PR #579, opened 2024-05-13
Increase chardet version limit for python-requests in contrib

thenno:
Fix warning from https://github.com/ytsaurus/ytsaurus/issues/222

1. [zlobober](https://github.com/zlobober) is using charset_normalizer == `5.1.0` for a couple of months and confirms that everything works fine.
2. As far as I can see, difference between `5.1.0` and `5.2.0` is only in one commit - https://github.com/chardet/chardet/commit/0e9b7bc20366163efcc221281201baff4100fe19

Thus I propose to increase charset_normalizer version limit to 5.2.0.

### DIFF
--- a/yt/python/contrib/python-requests/requests/__init__.py
+++ b/yt/python/contrib/python-requests/requests/__init__.py
@@ -87,8 +87,8 @@ def check_compatibility(urllib3_version, chardet_version, charset_normalizer_ver
     if chardet_version:
         major, minor, patch = chardet_version.split('.')[:3]
         major, minor, patch = int(major), int(minor), int(patch)
-        # chardet_version >= 3.0.2, <= 5.2.0
-        assert (3, 0, 2) <= (major, minor, patch) <= (5, 2, 0)
+        # chardet_version >= 3.0.2, < 6.0.0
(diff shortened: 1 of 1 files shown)


yt/python/contrib/python-requests/requests/__init__.py
```diff
@@ -87,8 +87,8 @@ def check_compatibility(urllib3_version, chardet_version, charset_normalizer_ver
     if chardet_version:
         major, minor, patch = chardet_version.split('.')[:3]
         major, minor, patch = int(major), int(minor), int(patch)
-        # chardet_version >= 3.0.2, < 5.0.0
-        assert (3, 0, 2) <= (major, minor, patch) < (5, 0, 0)
+        # chardet_version >= 3.0.2, <= 5.2.0
+        assert (3, 0, 2) <= (major, minor, patch) <= (5, 2, 0)
```
Kontakter, line 91: Let use the condition from the latest requests version: https://github.com/psf/requests/blob/main/src/requests/__init__.py#L79

thenno: A bit risky, but ok:)

I've updated the pr.

Kontakter (approved):

PR #587, opened 2024-05-16
docker image to run python wrapper tests in

dmi-feo:


yt/python/docker_test_image/README.md
```diff
@@ -0,0 +1,9 @@
+## Run tests
+```
+./run_wrapper_tests.sh test_cypress_commands.py
+```
+
+## Create new image
+```
+make docker tag=ytsaurus/pywrapper_tests_base_image:1.2.3
```
dmi-feo, line 8: TODO: change to the real tag

Kontakter:
2 remarks

1. This docker image share a lot with odin docker image, could we make some deduplication here? https://github.com/ytsaurus/ytsaurus/blob/main/yt/odin/Dockerfile
2. I would prefer to place it at `yt/python/docker/test_image`, since it is also can be used for `yt/yt/tests/integration`

yt/python/docker/test_image/Makefile
```diff
@@ -0,0 +1,11 @@
+.PHONY: docker
+
+docker:
+	# TODO: check there is no uncommited changes in yt/python/yt/packages to prevent prepare_python_modules fail.
+	docker build \
+	  ../../../../ \
+	  -f ../../../odin/Dockerfile \
+	  --build-arg="BUILD_TARGETS=" \
+	  --target lib-builder \
+	  -t odin/lib-builder && \
```
Kontakter, line 10: I think `&& \` is unncesessary here

dmi-feo:
It took a little longer than I expected, but now the PR and me are both ready for discussion.

- since odin and the test image now share a common base, I find it reasonable to keep all the dockerfiles in the same place. I guess `yt/docker` can be such a place. The whole statement is discussable though, nothing prevents to move the files to anywhere else
- all of a sudden clang-16 was introduced: https://github.com/ytsaurus/ytsaurus/commit/6021165602a8506451b22e8b9c63d0222b77e386. Unfortunately I didn't manage to adopt these changes, build fails with `Configuring incomplete, errors occurred!` when run on clang-16. Original Odin dockerfile is broken too, so I added `clang14.toolchain` as a temporary solution - until some experienced c++-ninja doesn't fix it :)
- while managing multistage docker builds without bake is just painful, doing so with "branching" builds (when a stage can be built with different args and be used in different contexts) should be outlawed. Thus please welcome `targets.hcl`

.dockerignore
```diff
@@ -0,0 +1,3 @@
+yt/docker/*
```
Kontakter, line 1: What is the point of this `.dockerignore` ?

yt/docker/targets.hcl
```diff
@@ -0,0 +1,71 @@
+variable "REPO_PATH" {}
+
+variable "BUILD_ROOT" {
+  default = "${REPO_PATH}/yt/docker"
+}
+
+
+# Common targets
+
+target "built_yson" {
```
Kontakter, line 10: It is better to name `python_binary_libraries`

yt/docker/targets.hcl
```diff
@@ -0,0 +1,71 @@
...
+  }
+  contexts = {
+    base_img = "docker-image://ubuntu:jammy"
+  }
+
+  dockerfile = "${BUILD_ROOT}/bake/built_yt.Dockerfile"
+  context = "${REPO_PATH}"
+}
+
+target "built_yt_all" {
```
Kontakter, line 22: Let name it `yt_all`?

I don't understand why we want to precede the name with prefix `built_`

yt/docker/run_bake
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
Kontakter, line 1: Let add `-ex` here

yt/docker/run_bake
```diff
@@ -0,0 +1,19 @@
...
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+BAKE_FILES=(
+    targets.hcl
+)
+
+for BFILE in "${BAKE_FILES[@]}"
+do
+  FILE_ARGS+=(--file "${SCRIPT_DIR}/${BFILE}")
```
Kontakter, line 13: Let use 4 spaces identation

yt/docker/run_bake
```diff
@@ -0,0 +1,19 @@
...
+done
+
+YT_SOURCE_ROOT="${SCRIPT_DIR}/../../"
+export REPO_PATH=${YT_SOURCE_ROOT}
+
+exec docker buildx bake "${FILE_ARGS[@]}" "$@"
```
Kontakter, line 19: Why exec is needed here?

yt/python/yt/run_wrapper_tests.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
Kontakter, line 1: Let add `-ex`

Kontakter, line 1: Let name this script `run_tests_with_docker.sh`

.dockerignore
dmi-feo: It ignores changes in `yt/docker`, so that yt would not be rebuilt after changes in dockerfiles

yt/docker/bake/built_yt.Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+    && unzip protoc.zip -d /usr/local \
+    && rm protoc.zip
+
+COPY --link ./ ${SOURCE_ROOT}/
+
+WORKDIR ${ROOT}
+
+RUN mkdir -p ${BUILD_ROOT} ; cd ${BUILD_ROOT} \
+    && cmake -G Ninja \
+        -DCMAKE_BUILD_TYPE=Release \
```
achulkov2, line 51: Is `Release` the best build type for tests?

dmi-feo: But is it a real case when you need to debug ytserver while running python sdk tests?

zlobober:
Went discussing to SIG SDK and SIG CI/CD/Devtools

PR #591, opened 2024-05-17
Move state managing logic from proxies to qt

Krisha11:
Changing the architecture of communication between QT and HTTP/RPC proxies.

**Problem:**

Now we want to support multiple ACOs in a single query. This causes incompatibility between older and newer components. SDK/Proxies/QT. Therefore, we need to update all these components simultaneously, which is not an easy task. A similar problem existed when we added the first version of ACO, and possibly we will have similar problems in future developments.

**Solution:**

Step 1.
Move all complex logic from proxies to QT. After that proxies will simply resend all requests to QT service. And QT will write to state tables.
(This PR)

Step 2
Make it possible to work with RPC requests from different proxy versions. Add support for different versions of the RPC API in the QT proxy server.
(Next PR, together with multiple ACO support)

Step 3
Make it possible to work with requests from different SDK versions in proxies. Add support for different versions of the RPC API in the RPC proxy, and feature flags to the HTTP protocol.
(Next PR, together with multiple ACO support)

**Benefits:**

All communication with QT state tables will take place through the QT service. Therefore, this logic will always be delivered in the same image. This removes all starting incompatibilities and makes it possible to update mentioned components in the next order.
QT->Proxies->SDK

### DIFF
--- a/yt/yt/server/query_tracker/proxy_service.cpp
+++ b/yt/yt/server/query_tracker/proxy_service.cpp
@@ -1,0 +1,268 @@
+#include "private.h"
+#include "proxy_service.h"
+#include "query_tracker_proxy.h"
+
+#include <yt/yt/client/api/rpc_proxy/helpers.h>
+
+#include <yt/yt/ytlib/query_tracker_client/proto/query_tracker_service.pb.h>
+
+#include <yt/yt/ytlib/query_tracker_client/query_tracker_service_proxy.h>
+#include <yt/yt/ytlib/query_tracker_client/helpers.h>
+
+#include <yt/yt/core/rpc/service_detail.h>
+
+namespace NYT::NQueryTracker {
+
+using namespace NApi;
+using namespace NConcurrency;
+using namespace NQueryTrackerClient;
+using namespace NRpc;
+using namespace NYTree;
+
+static const TYsonString EmptyMap = TYsonString(TString("{}"));
+
+////////////////////////////////////////////////////////////////////////////////
+
+class TProxyService
+    : public TServiceBase
+{
+public:
+    TProxyService(IInvokerPtr proxyInvoker, TQueryTrackerProxyPtr queryTracker)
+        : TServiceBase(
+            std::move(proxyInvoker),
+            TQueryTrackerServiceProxy::GetDescriptor(),
+            QueryTrackerLogger)
+        , QueryTracker_(std::move(queryTracker))
+    {
+        RegisterMethod(RPC_SERVICE_METHOD_DESC(StartQuery));
+        RegisterMethod(RPC_SERVICE_METHOD_DESC(AbortQuery));
+        RegisterMethod(RPC_SERVICE_METHOD_DESC(GetQueryResult));
+        RegisterMethod(RPC_SERVICE_METHOD_DESC(ReadQueryResult));
+        RegisterMethod(RPC_SERVICE_METHOD_DESC(GetQuery));
+        RegisterMethod(RPC_SERVICE_METHOD_DESC(ListQueries));
+        RegisterMethod(RPC_SERVICE_METHOD_DESC(AlterQuery));
+    }
+
+private:
+    const TQueryTrackerProxyPtr QueryTracker_;
+
+    DECLARE_RPC_SERVICE_METHOD(NQueryTrackerClient::NProto, StartQuery)
+    {
+        auto queryId = TQueryId::Create();
+        ToProto(response->mutable_query_id(), queryId);
+
+        TStartQueryOptions options;
+        if (request->has_settings()) {
+            options.Settings = ConvertToNode(TYsonStringBuf(request->settings()));
+        }
+        if (request->has_annotations()) {
+            options.Annotations = ConvertToNode(TYsonStringBuf(request->annotations()))->AsMap();
+        }
+        if (request->has_access_control_object()) {
+            options.AccessControlObject = request->access_control_object();
+        }
+        options.Draft = request->draft();
+        options.Files = ConvertTo<std::vector<TQueryFilePtr>>(TYsonStringBuf(request->files()));
+
+        auto engine = static_cast<EQueryEngine>(request->engine());
+        auto query = request->query();
+        auto user = context->GetAuthenticationIdentity().User;
+
+        context->SetRequestInfo("QueryId: %v, User: %v", queryId, user);
+        context->SetResponseInfo("QueryId: %v", queryId);
+
+        QueryTracker_->StartQuery(queryId, engine, query, options, user);
+
+        context->Reply();
+    }
+
+    DECLARE_RPC_SERVICE_METHOD(NQueryTrackerClient::NProto, AbortQuery)
+    {
+        TQueryId queryId;
+        FromProto(&queryId, request->query_id());
+
+        TAbortQueryOptions options;
+        options.AbortMessage = request->has_abort_message()
+            ? std::make_optional(request->abort_message())
+            : std::nullopt;
+
+        auto user = context->GetAuthenticationIdentity().User;
+
+        context->SetRequestInfo("QueryId: %v, User: %v", queryId, user);
+        context->SetResponseInfo("QueryId: %v", queryId);
+
+        QueryTracker_->AbortQuery(queryId, options, user);
+
+        context->Reply();
+    }
+
+    DECLARE_RPC_SERVICE_METHOD(NQueryTrackerClient::NProto, GetQueryResult)
+    {
+        TQueryId queryId;
+        FromProto(&queryId, request->query_id());
+
+        auto resultIndex = request->result_index();
+        auto user = context->GetAuthenticationIdentity().User;
+
+        context->SetRequestInfo("QueryId: %v, ResultIndex: %v, User: %v", queryId, resultIndex, user);
+        context->SetResponseInfo("QueryId: %v", queryId);
+
+        auto queryResult = QueryTracker_->GetQueryResult(queryId, resultIndex, user);
+
+        ToProto(response->mutable_query_id(), queryResult.Id);
+        response->set_result_index(queryResult.ResultIndex);
+        if (queryResult.Schema) {
+            ToProto(response->mutable_schema(), queryResult.Schema);
+        }
+        ToProto(response->mutable_error(), queryResult.Error);
+        response->set_is_truncated(queryResult.IsTruncated);
+        ToProto(response->mutable_data_statistics(), queryResult.DataStatistics);
+
+        context->Reply();
+    }
+
+    DECLARE_RPC_SERVICE_METHOD(NQueryTrackerClient::NProto, ReadQueryResult)
+    {
+        TQueryId queryId;
+        FromProto(&queryId, request->query_id());
+
+        TReadQueryResultOptions options;
+        options.LowerRowIndex = request->has_lower_row_index()
+            ? std::make_optional(request->lower_row_index())
+            : std::nullopt;
+        options.UpperRowIndex = request->has_upper_row_index()
+            ? std::make_optional(request->upper_row_index())
+            : std::nullopt;
+        options.Columns = request->has_columns()
+            ? std::make_optional(FromProto<std::vector<TString>>(request->columns().items()))
+            : std::nullopt;
+
+        auto resultIndex = request->result_index();
+        auto user = context->GetAuthenticationIdentity().User;
+
+        context->SetRequestInfo("QueryId: %v, ResultIndex: %v, User: %v", queryId, resultIndex, user);
+        context->SetResponseInfo("QueryId: %v", queryId);
+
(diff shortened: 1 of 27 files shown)


yt/yt/ytlib/query_tracker_client/query_tracker_service_proxy.h
Krisha11, line 1: It is very similar to the existing queries description structures
https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt_proto/yt/client/api/rpc_proxy/proto/api_service.proto#L3077
https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/client/api/query_tracker_client.h#L40

But it seems reasonable to duplicate here, for possible changes in communication between proxies and qt.

yt/yt/server/query_tracker/proxy_service.cpp
```diff
@@ -0,0 +1,268 @@
+#include "private.h"
```
Krisha11, line 1: This logic is completely moved from client_queries_impl.cpp with small structural changes

yt/yt/server/query_tracker/proxy_service.cpp
Krisha11, line 1: There is only the conversion from proto to native_client_structs

yt/yt/server/query_tracker/query_tracker_proxy.cpp
Krisha11, line 1: This logic is completely moved from client_queries_impl.cpp with small structural changes

yt/yt/ytlib/api/native/connection.cpp
Krisha11, line 1: There is only the conversion from native_client_structs to proto

yt/yt/ytlib/api/native/connection.cpp
```diff
@@ -1104,10 +1126,8 @@ class TConnection
             }));
     }
 
-    //! Returns a pair consisting of the client for the Query Tracker cluster and the path to the Query Tracker root in Cypress.
-    std::pair<IClientPtr, TString> GetQueryTrackerStage(const TString& stage) override
-    {
-        auto findStage = [&stage, this] (const TString& cluster) -> std::pair<IClientPtr, TString> {
+    std::pair<IClientPtr, TQueryTrackerStageConfigPtr> FindQueryTrackerStage(const TString& stage) {
```
Krock21, line 1129: [Opening braces of classes and functions are placed on the next line.](https://github.com/ytsaurus/ytsaurus/blob/main/yt/styleguide/cpp.md#braces)

Can we move it back?

yt/yt/ytlib/query_tracker_client/helpers.cpp
```diff
@@ -51,4 +58,132 @@ bool IsFinishedState(EQueryState state)
 
 ////////////////////////////////////////////////////////////////////////////////
 
+void ToProto(NProto::TQuery* protoQuery, const NApi::TQuery& query)
+{
+    protoQuery->Clear();
+
```
Krock21, line 64: Can we do `static_assert(NProto::TQuery::FieldsCount == 42)` and `static_assert::(NApi::TQuery::FieldCount == 654)` so people don't forget to update this part.

yt/yt/ytlib/query_tracker_client/proto/query_tracker_service.proto
```diff
@@ -0,0 +1,147 @@
+package NYT.NQueryTrackerClient.NProto;
```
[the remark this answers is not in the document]
Krock21: Is it better to make a dependency instead of duplication, like this?
```
message TReqStartQuery
{
    required NYT.NApi.NRpcProxy.NProto.TReqStartQuery rpcProxyRequest = 1;
    required string additionalField1 = 2;
    ...
}
```
It will save us from writing mapping between structures

yt/yt/ytlib/query_tracker_client/config.cpp
```diff
@@ -10,6 +18,8 @@ void TQueryTrackerStageConfig::Register(TRegistrar registrar)
         .Default("//sys/query_tracker");
     registrar.Parameter("user", &TThis::User)
         .Default("query_tracker");
+    registrar.Parameter("channel", &TThis::Channel)
```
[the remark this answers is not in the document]
psushin: Marina says, that we will use endpointset for discovery here in internal installations.

yt/yt/ytlib/query_tracker_client/helpers.cpp
```diff
@@ -51,4 +58,132 @@ bool IsFinishedState(EQueryState state)
...
+    ToProto(protoQuery->mutable_id(), query.Id);
+
+    if (query.Engine) {
```
apollo1321, line 67: Code duplication
https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/client/api/rpc_proxy/helpers.cpp#L1337

yt/yt/ytlib/query_tracker_client/query_tracker_service_proxy.h
```diff
@@ -0,0 +1,29 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+class TQueryTrackerServiceProxy
+    : public NRpc::TProxyBase
+{
+public:
+    DEFINE_RPC_PROXY(TQueryTrackerServiceProxy, TQueryTrackerService,
+        .SetProtocolVersion(NRpc::TProtocolVersion{0, 0}));
+
+    DEFINE_RPC_PROXY_METHOD(NProto, StartQuery);
```
apollo1321, line 18: Is it possible to remove code duplication here too? I think that having multiple places with the same RPC_PROXY_METHOD definitions may be confusing. Do we have a plan for the future refactoring of the Query Tracker RPC Proxy methods?
https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/client/api/rpc_proxy/api_service_proxy.h#L187

yt/yt/ytlib/api/native/client_queries_impl.cpp
```diff
@@ -383,706 +54,214 @@ TQueryId TClient::DoStartQuery(EQueryEngine engine, const TString& query, const
...
     }
 
-    auto [client, root] = GetNativeConnection()->GetQueryTrackerStage(options.QueryTrackerStage);
-
-    static const TYsonString EmptyMap = TYsonString(TString("{}"));
-
-    auto queryId = TQueryId::Create();
-
-    YT_LOG_DEBUG("Starting query (QueryId: %v, Draft: %v)", queryId, options.Draft);
+    TQueryTrackerServiceProxy proxy(
```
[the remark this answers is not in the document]
apollo1321: Ah, I see that QueryTrackerStage is passed in the options.

yt/yt/server/query_tracker/query_tracker_proxy.h
```diff
@@ -0,0 +1,80 @@
...
+
+    TListQueriesResult ListQueries(
+        const TListQueriesOptions& options,
+        const TString& user);
+
+    void AlterQuery(
+        const TQueryId queryId,
+        const TAlterQueryOptions& options,
+        const TString& user);
+
```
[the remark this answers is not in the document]
Krisha11: At the moment, there is no logic in it that should be assigned to qt (related to the qt state) As soon as it appears, we will add this method

yt/yt/server/query_tracker/query_tracker_proxy.cpp
```diff
@@ -0,0 +1,1088 @@
...
+    , StateRoot_(std::move(stateRoot))
+{ }
+
+void TQueryTrackerProxy::StartQuery(
+        const TQueryId queryId,
+        const EQueryEngine engine,
+        const TString& query,
+        const TStartQueryOptions& options,
+        const TString& user)
+{
```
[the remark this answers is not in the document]
Krisha11: It was a micro-optimization. There is no need to move large files to another service - if you can check them here

But maybe it's not that important. And it's better to standardize everything by removing all meaningful logic in qt altogether.

yt/yt/ytlib/query_tracker_client/query_tracker_service_proxy.h
Krisha11: I don't understand how we can remove this. I would appreciate any advice on how to do it.

There are no plans for refactoring, but it would be great if you could describe how it could be done better.

yt/yt/ytlib/query_tracker_client/proto/query_tracker_service.proto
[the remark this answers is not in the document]
Krisha11: This dependency will not save us from mapping between structures. To do this, you will have to remove common code for different proxies in the native client.

Also, this dependency will remove the ability for us to change the protocol only in the RPC API or only between qt <-> proxies. But maybe it's not so necessary..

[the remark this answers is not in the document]
Krisha11: Сhanged it, as you suggested

yt/yt/ytlib/query_tracker_client/helpers.cpp
Krisha11: Removed this part

yt/yt/server/query_tracker/query_tracker_proxy.cpp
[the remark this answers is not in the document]
Krisha11: Moved it to other code

yt/yt/server/query_tracker/proxy_service.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+        context->Reply();
+    }
+
+    DECLARE_RPC_SERVICE_METHOD(NQueryTrackerClient::NProto, GetQueryResult)
+    {
+        TQueryId queryId;
+        FromProto(&queryId, request->query_id());
+
+        auto resultIndex = request->result_index();
```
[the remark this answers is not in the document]
Krisha11: Removed code in rpc_proxy, after generalizing protos

yt/yt/ytlib/query_tracker_client/helpers.cpp
Krisha11: Removed

yt/yt/server/query_tracker/proxy_service.cpp
```diff
@@ -0,0 +1,268 @@
...
+private:
+    const TQueryTrackerProxyPtr QueryTracker_;
+
+    DECLARE_RPC_SERVICE_METHOD(NQueryTrackerClient::NProto, StartQuery)
+    {
+        auto queryId = TQueryId::Create();
+        ToProto(response->mutable_query_id(), queryId);
+
+        TStartQueryOptions options;
+        if (request->has_settings()) {
```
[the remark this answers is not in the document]
Krisha11: Added runtime check because I couldn't find a way to check statically

yt/yt/ytlib/query_tracker_client/query_tracker_service_proxy.h
apollo1321: I realized that the protobuf types are different. I think it's fine in this case.

yt/yt/server/query_tracker/query_tracker_proxy.h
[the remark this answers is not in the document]
Krock21: I think we should move GetQueryTrackerInfo to QT. The idea is to get info about query tracker version, and this API tries to analyze QT state for it

The logic may also simplify a lot, as we will know QT version from QT code and can hardcode supported features (such as Access Control), instead of invasive QT state analysis (`//sys/ACO/queries` existence check)

For now just copying the current logic is fine. For new features it will be easier to add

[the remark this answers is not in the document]
Krisha11: Added GetQueryTrackerInfo

yt/yt/server/query_tracker/query_tracker_proxy.cpp
```diff
@@ -0,0 +1,1164 @@
...
+    TNodeExistsOptions nodeExistsOptions;
+    nodeExistsOptions.ReadFrom = EMasterChannelKind::Cache;
+
+    static const TYsonString EmptyMap = TYsonString(TString("{}"));
+    TYsonString supportedFeatures = EmptyMap;
+    if (attributes.AdmitsKeySlow("supported_features")) {
+        supportedFeatures = BuildYsonStringFluently()
+            .BeginMap()
+                .Item("access_control").Value(true)
+            .EndMap();
```
Krock21, line 1129: Would be nice to do it like here: https://github.com/ytsaurus/ytsaurus/blob/45a5e988eaed39c6b9790b094f6ad5dbf7d6ff6d/yt/yt/client/driver/etc_commands.cpp#L78-L81
With a comment "These features are guaranteed to be deployed before or with this code."

Krock21 (approved):

Krisha11:
Shipped!

PR #597, opened 2024-05-21
CRI job environenment tests

koct9i:
- Add user job test for each job environment
  Mix a little bit magic to duplicate test classes for each job environment.
  
- Migrate CRI job environment to python3
  There is no reason to keep any python2 compatibility.

### DIFF
--- a/yt/yt/tests/integration/node/test_user_job.py
+++ b/yt/yt/tests/integration/node/test_user_job.py
@@ -451,7 +451,7 @@ class TestSandboxTmpfs(TestPortoJobEnvironmentBase):
         write_table("//tmp/t_input", {"foo": "bar"})
 
         op = map(
-            command="python -c 'import time; x = \"0\" * (200 * 1000 * 1000); time.sleep(2)'",
+            command="python3 -c 'import time; x = \"0\" * (200 * 1000 * 1000); time.sleep(2)'",
             in_="//tmp/t_input",
             out="//tmp/t_output",
             spec={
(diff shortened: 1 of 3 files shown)


yt/yt/tests/integration/node/test_user_job.py
```diff
@@ -55,19 +55,46 @@ def find_operation_by_mutation_id(mutation_id):
...
     }
 
+
+class TestPortoJobEnvironmentBase(TestJobEnvironmentBase):
+    USE_PORTO = True
+    # TODO(khlebnikov): For now tmpfs available only for porto.
+
+
+class TestCriJobEnvironmentBase(TestJobEnvironmentBase):
+    JOB_ENVIRONMENT_TYPE = 'cri'
```
Kontakter, line 80: Let use double quotes here and below

yt/yt/tests/integration/node/test_user_job.py
```diff
@@ -55,19 +55,46 @@ def find_operation_by_mutation_id(mutation_id):
...
+
+
+class TestEachJobEnvironmentBase(TestJobEnvironmentBase):
```
Kontakter, line 83: I am not happy with this idea because class names are hidden within this code and it complicated discovery of the tests.

I think that more verbose scheme is actually better. Like this:
```python
class BaseTestJobInputCache(...):
    ...

class TestJobInputCachePorto(BaseTestJobInputCache):
     USE_PORTO = True

class TestJobInputCacheCri(BaseTestJobInputCache):
     JOB_ENVIRONMENT_TYPE = 'cri'

```

koct9i: Ok

PR #611, opened 2024-05-27
Data nodes disk throttlers for uncategorized load

Krock21:
Replace unlimited disk throttlers with uncategorized disk throttlers that:
- combine in and out load
- are configurable

Please cherry-pick it to stable/23.2 and stable/24.1

### DIFF
--- a/yt/yt/server/node/data_node/config.cpp
+++ b/yt/yt/server/node/data_node/config.cpp
@@ -100,6 +100,9 @@ void TChunkLocationConfig::ApplyDynamicInplace(const TChunkLocationDynamicConfig
     }
     UpdateYsonStructField(ThrottleDuration, dynamicConfig.ThrottleDuration);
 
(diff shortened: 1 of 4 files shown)


gritukan (approved):

psushin:
Shipped!

psushin:
Ok, this seems dangerous, but reasonable for cases, when one wants to keep the total disk load well below it's physical limits.

Krock21:
> Ok, this seems dangerous, but reasonable for cases, when one wants to keep the total disk load well below it's physical limits.

This was our case. Our exec and data nodes share the same disk, and we saw that jobs struggle in the "spawning job proxy" stage because data nodes utilized the whole disk

We used network throttlers to limit the load, but it limits all disks on data nodes which is bad for performance

Throttlers in this PR allow fine grained throttling per disk

psushin:
Сherry-picked to stable/23.2 and stable/24.1

PR #626, opened 2024-05-30
Remove chyt-controller and query-tracker from ytsaurus core-like images

achulkov2:
These components are deployed from their own images for quite a while now.


Krock21:
And it will help us to avoid typo errors when users specify `ytsaurus/ytsaurus` image for queryTracker with a very old version of QT (happened to us)

yt/docker/ytsaurus/credits/chyt-controller.CREDITS
```diff
@@ -1,2007 +0,0 @@
-Go (the standard library)
```
savnadya, line 1: It's used in yt/docker/strawberry, let's move it

achulkov2: Fixed.

savnadya:
Shipped!

PR #631, opened 2024-06-02
Add dynamic table log writer

achulkov2:
Adds a new `dynamic_table` log writer, which writes log events to the configured sorted or ordered dynamic table.
Log events are serialized into a YSON representation, which is interpreted similarly to the `insert-rows` cli command.
It is your responsibility to create the table with a compatible schema. 
See integration test for example configuration.

The writer is available for use in all YT server components.


yt/yt/client/logging/dynamic_table_log_writer.cpp
```diff
@@ -0,0 +1,372 @@
...
+
+        if (Suspended_) {
+            return 0;
+        }
+
+        auto guard = Guard(SpinLock_);
+
+        // This is necessary to avoid hitting the limit of allowed row modifications within a single transaction.
+        // Also, there is no upside from huge batches.
+        if (CurrentRowCount_ > Config_->MaxBatchRowCount || std::ssize(CurrentBuffer_) > Config_->MaxBatchWeight) {
```
gritukan, line 124: Technically >=?

yt/yt/client/logging/dynamic_table_log_writer.cpp
```diff
@@ -0,0 +1,372 @@
...
+        : TRateLimitingLogWriterBase(
+            BIND(&TDynamicTableLogWriter::WriteImpl, Unretained(this)),
+            name,
+            config)
+        , TSegmentSizeReporter(name)
+        , ClientHolder_(std::move(clientHolder))
+        , Formatter_(std::move(formatter))
+        , Config_(config)
+        , Invoker_(std::move(invoker))
+    , FlushExecutor_(New<TPeriodicExecutor>(
```
gritukan, line 81: indentation

yt/yt/client/logging/dynamic_table_log_writer.cpp
```diff
@@ -0,0 +1,372 @@
...
+        VERIFY_INVOKER_AFFINITY(Invoker_);
+
+        auto client = ClientHolder_->Client.Load();
+
+        if (!client) {
+            THROW_ERROR_EXCEPTION("Cannot flush rows without configured YT client");
+        }
+
+        auto tableMountCache = client->GetTableMountCache();
+        auto tableInfo = WaitFor(tableMountCache->GetTableInfo(Config_->TablePath))
```
gritukan, line 255: WaitForFast can be done here to avoid a context switch in a typical case. It does not make much sense in this case, but is a good pattern anyway

yt/yt/client/logging/dynamic_table_log_writer.cpp
```diff
@@ -0,0 +1,372 @@
...
+
+        auto guard = Guard(SpinLock_);
+
+        // This is necessary to avoid hitting the limit of allowed row modifications within a single transaction.
+        // Also, there is no upside from huge batches.
+        if (CurrentRowCount_ > Config_->MaxBatchRowCount || std::ssize(CurrentBuffer_) > Config_->MaxBatchWeight) {
+            RotateBuffer();
+        }
+
+        TStringOutput outputStream(CurrentBuffer_);
```
gritukan, line 128: Why do we convert batch of events to a single string first and then convert them into a list of rows? Is it possible to create rows here?

yt/yt/core/logging/log_writer_detail.cpp
```diff
@@ -73,50 +73,131 @@ i64 TRateLimitCounter::GetAndResetLastSkippedEventsCount()
...
+{
+    CategoryToRateLimit_.clear();
+    for (const auto& it : categoryRateLimits) {
+        GetCategoryRateLimitCounter(it.first)->SetRateLimit(it.second);
+    }
+}
+
+void TRateLimitingLogWriterBase::WriteLogSkippedEvent(i64 eventsSkipped, TStringBuf skippedBy)
+{
+    if (eventsSkipped > 0) {
```
gritukan, line 148: These ifs can be joined with &&

yt/yt/client/logging/dynamic_table_log_writer.cpp
achulkov2: The main problem is that we need to know the table schema in order to convert the yson strings into unversioned rows, and getting the schema is a YT client request.

The tricky part is that the schema might change, for example, we forgot to add a column, saw the errors in logs and then altered the table to contain the forgotten column.

So either we need some periodic that will fetch and store the schema, or we fetch it upon write.

For now I decided to go with the second approach. Later on, if we switch to some writer that needs to know the schema upon creation, we can switch to the first approach and move parsing into `Write`. However, the writer will have to support schema-alerting :)

achulkov2: This is considered resolved (without any changes), but I will keep this thread open for now, so that future reviewers see it.

yt/yt/tests/integration/misc/test_logging.py
```diff
@@ -0,0 +1,166 @@
...
+                    "common_fields": {
+                        "cluster": "primary",
+                    },
+                    "flush_period": 400,
+                },
+            },
+            # This is specified to test rate limits.
+            "category_rate_limits": {
+                # No sequoia logs for you!
+                "SequoiaServer": 0,
```
achulkov2, line 66: It may be wise to use another category, since `Sequoia` logs are relatively new and are not present in 23.2, for example.

yt/yt/core/logging/formatter.cpp
```diff
@@ -86,10 +90,35 @@ TLogEvent GetSkippedLogStructuredEvent(i64 count, TStringBuf skippedBy)
...
+            return GetSkippedLogStructuredEvent(count, skippedBy);
+        default:
+            YT_ABORT();
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+TPlainTextLogFormatter::TPlainTextLogFormatter(bool enableSourceLocation)
+    : TLogFormatterBase( enableSourceLocation)
```
achulkov2, line 121: Extra space.

yt/yt/client/logging/dynamic_table_log_writer.cpp
```diff
@@ -0,0 +1,373 @@
...
+using namespace NTabletClient;
+using namespace NTransactionClient;
+using namespace NYson;
+
+////////////////////////////////////////////////////////////////////////////////
+
+// This log category is special in the default log manager: events with this category are logged solely to stderr.
+// Note that this behaviour does not depend on the log level of the message.
+// We use this logger to (try our best) to avoid perpetual log message production.
+// Currently it is impossible to disable logging along a chain of invocations completely, so some induced log messages will loop back into our writer.
```
maxim-babenko, line 47: Some lines here are way too long, let's split them

yt/yt/client/logging/dynamic_table_log_writer.cpp
```diff
@@ -0,0 +1,373 @@
...
+        CurrentRowCount_ = 0;
+    }
+
+    // Periodic executor guarantees that at most one callback is executed at once.
+    // This is important to avoid log reordering.
+    void DoFlush()
+    {
+        VERIFY_INVOKER_AFFINITY(Invoker_);
+
+        TString ysonRows;
```
maxim-babenko, line 202: Why TString, not TYsonString?
Or maybe even shared refs to ease memory tracking?

yt/yt/client/logging/dynamic_table_log_writer.cpp
```diff
@@ -0,0 +1,373 @@
...
+    {
+        ClientHolder_->Client.Store(std::move(client));
+    }
+
+    void ValidateConfig(const NYTree::IMapNodePtr& configNode) override
+    {
+        auto config = ParseConfig(configNode);
+        if (config->Format != ELogFormat::Yson) {
+            THROW_ERROR_EXCEPTION(
+                "Writers of type %Qv must be configured to use %Qv format",
```
maxim-babenko, line 327: Qlv for format

yt/yt/client/logging/dynamic_table_log_writer.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+
+class TDynamicTableLogWriterFactory
+    : public IDynamicTableLogWriterFactory
+{
+public:
+    TDynamicTableLogWriterFactory()
+        : ClientHolder_(New<TClientHolder>())
+        , WriteThreadPool_(CreateThreadPool(1, "LogYTWrite"))
```
maxim-babenko, line 314: Do we need a thread pool here?

maxim-babenko (changes requested):

yt/yt/tests/integration/misc/test_logging.py
achulkov2: Changed to Concurrency.

yt/yt/client/logging/dynamic_table_log_writer.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+        YT_LOG_INFO("Created dynamic table log writer (Config: %v)", ConvertToYsonString(Config_, EYsonFormat::Text));
+    }
+
+    // No synchronous YT writes/calls are actually performed by Write/Flush, since we cannot
+    // switch contexts in log manager's main logging thread.
+    // Moreover, the current log manager is not designed for handling asynchronous event
+    // writers. All write/flush calls are made consecutively and are considered successful
+    // at once. Changing this would require modifying the logging event pipeline significantly.
+    i64 WriteImpl(const TLogEvent& event)
```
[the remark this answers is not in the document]
achulkov2: Moved to private here and in log_writer_detail.h.

yt/yt/client/logging/dynamic_table_log_writer.cpp
achulkov2: There is some serialization/deserialization load, but I guess we can live with a single action queue thread for now. If necessary, we can change it to a thread pool with a configurable number of threads.

yt/yt/client/ya.make
```diff
@@ -180,6 +180,9 @@ SRCS(
     complex_types/uuid_text.cpp
     complex_types/time_text.cpp
 
+    logging/config.cpp
```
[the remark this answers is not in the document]
achulkov2: Sure, why not. It still depends on `yt/yt/client` though.

yt/yt/client/logging/dynamic_table_log_writer.cpp
achulkov2: Changed to `TSharedRef` for better memory tracking. But more generally speaking I would prefer to move towards the ability to pass yson input to C++ client write methods.

achulkov2: Switched to shared refs.

achulkov2:
Addressed all comments by @maxim-babenko and added some profiling which we found necessary after deploying this feature locally. Placed these changes in a separate commit for easier reading.

yt/yt/core/logging/formatter.cpp
```diff
@@ -205,6 +213,9 @@ i64 TStructuredLogFormatter::WriteFormatted(IOutputStream* stream, const TLogEve
                     .Item("level").Value(FormatEnum(event.Level))
                     .Item("category").Value(event.Category->Name);
             })
+            .DoIf(EnableHostField_, [&] (auto fluent) {
+                fluent.Item("host").Value(NNet::GetLocalHostName());
+            })
```
eshcherbin, line 218: Could you explain the need for this "host" field?
I don't like having it separated from the other system fields. If we want to be able to enable only a subset of system fields, I'd suggest changing the bool option with a flag enum. However, it should be done in a separate PR, this one's already complex enough.

yt/yt/core/logging/log_manager.cpp
```diff
@@ -503,6 +504,11 @@ class TLogManager::TImpl
             Synchronize(TInstant::Now() + Config_->ShutdownGraceTimeout);
         }
 
+        // For now this is the only way to wait for log writers that perform asynchronous flushes.
+        if (Config_->ShutdownBusyTimeout) {
+            Sleep(Config_->ShutdownBusyTimeout);
+        }
```
eshcherbin, line 510: I don't really like this sleep here. However, I understand that it would require a considerable refactoring to be able to wait for the dynamic table log writer closing. For now, let's at least add a TODO for later.

eshcherbin (changes requested):

yt/yt/core/logging/formatter.cpp
achulkov2: It is useful when you have multiple hosts producing logs to the same dynamic table. While this is something that we could append in the specific implementation for dynamic tables, I think it is closer to the other system fields and might be useful in some other log writer later on. 

I don't like it separated either, but I don't want to break existing structured logs. Changing to a flag enum is a good idea, I'll write a TODO.

yt/yt/core/logging/log_writer_detail.cpp
```diff
@@ -73,50 +73,128 @@ i64 TRateLimitCounter::GetAndResetLastSkippedEventsCount()
 
 ////////////////////////////////////////////////////////////////////////////////
 
-TStreamLogWriterBase::TStreamLogWriterBase(
-    std::unique_ptr<ILogFormatter> formatter,
-    TString name)
-    : Formatter_(std::move(formatter))
+TSegmentSizeReporter::TSegmentSizeReporter(TString logWriterName)
+    : CurrentSegmentSizeGauge_(
+        TProfiler{"/logging"}.WithSparse().WithTag("writer", std::move(logWriterName)).Gauge("/current_segment_size"))
```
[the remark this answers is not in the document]
achulkov2: I am dangerously scared of using global variables in loggers, because a bunch of the logic happens during static initialization. I had my fun with siof while writing the code. Plus this is code by prime@, whom I tend to trust in these matters :) 

But I will give it some thought in hopes of convincing myself that using a global profiler would be fine.

yt/yt/core/logging/log_writer_detail.cpp
```diff
@@ -73,50 +73,128 @@ i64 TRateLimitCounter::GetAndResetLastSkippedEventsCount()
...
-        }
-        if (!RateLimit_.IsLimitReached() && !categoryRateLimit->IsLimitReached()) {
-            auto bytesWritten = Formatter_->WriteFormatted(stream, event);
-            CurrentSegmentSize_ += bytesWritten;
-            CurrentSegmentSizeGauge_.Update(CurrentSegmentSize_);
+    // We do not do this in the constructor since it happens to be called during TLogManager construction itself.
+    if (!SystemCategory_) {
+        SystemCategory_ = GetDefaultLogManager()->GetCategory(SystemLoggingCategoryName);
+    }
+    bool isSystemEvent = event.Category == SystemCategory_;
```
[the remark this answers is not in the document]
achulkov2: It is new, it seemed reasonable from a general perspective when I was writing the code.

Now that I think about it, I can't say it makes too much of a difference: log writers that would receive system category events won't actually have any limits configured, since they are handled by a designated error log writer anyway. Meanwhile, log skipped system events will get written in any case, since they are not looped back into the rate limiting logic. 

I am fine with reverting this part, if you feel like it would improve readability.

yt/yt/core/logging/log_writer_detail.h
```diff
@@ -38,37 +40,90 @@ class TRateLimitCounter
...
+public:
+    explicit TSegmentSizeReporter(TString logWriterName);
+
+    void IncrementSegmentSize(i64 size);
+    void ResetSegmentSize(i64 size);
+
+private:
+    std::atomic<i64> CurrentSegmentSize_ = 0;
+    NProfiling::TGauge CurrentSegmentSizeGauge_;
+};
```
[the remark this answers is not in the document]
achulkov2: This class got separated during the many attempts of making this whole construction work, it is entirely possible that it doesn't need to be a separate class anymore (as it was originally). I will give it another look.

yt/yt/core/logging/log_writer_detail.cpp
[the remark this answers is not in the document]
eshcherbin: OK, no strong opinion here.

yt/yt/core/logging/log_writer_detail.cpp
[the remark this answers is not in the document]
eshcherbin: I'm not really worried about readability here, I just couldn't verify that this is 100% safe with a quick glance (and I couldn't allocate more time to think about it, given that it's not essential for the PR's primary goal).

Not really a strong opinion, but reverting it would make the code a bit tidier, yes.

maxim-babenko:
Shipped!

yt/yt/core/logging/log_writer_detail.cpp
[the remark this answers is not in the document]
achulkov2: Reverted this part.

yt/yt/client/logging/dynamic_table_log_writer.cpp
```diff
@@ -0,0 +1,422 @@
...
+        Formatter_->WriteFormatted(&countingOutputStream, event);
+
+        auto writtenBytes = static_cast<i64>(countingOutputStream.Counter());
+
+        BacklogWeight_.fetch_add(writtenBytes, std::memory_order::relaxed);
+        BacklogWeightGauge_.Update(BacklogWeight_.load(std::memory_order::relaxed));
+
+        ++CurrentRowCount_;
+        BacklogEventCount_.fetch_add(1, std::memory_order::relaxed);
+        BacklogEventsGauge_.Update(BacklogEventCount_.load(std::memory_order::relaxed));
```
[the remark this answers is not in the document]
achulkov2: Cool, didn't know about this functionality before!

yt/yt/core/logging/log_writer_detail.h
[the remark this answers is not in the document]
achulkov2: Moved this logic back into `TRateLimitingLogWriterBase` as protected methods.

yt/yt/core/logging/log_writer_detail.cpp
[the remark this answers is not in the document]
achulkov2: I gave it some thought and decided that I do not want to interfere with the ability to configure `TLogManager` before main. Whether this is actually allowed now is debatable and depends on the current release (for example, loggers themselves became siof-safe in trunk). To be on the safer side while still introducing some cleanliness, I moved the profiler into a class field.

maxim-babenko:
Shipped!

PR #634, opened 2024-06-03
Do not retry execve in process spawn action

koct9i:
Parent is blocked on vfork() - sleeping here for 5 seconds is a bad idea.

It's better to fail with error EProcessErrorCode::CannotStartProcess and
let initiator decide what to do: retry or abort operation.


E1pp (approved):

Kontakter:
I think that psushin@ should take a look, since he added these retries.

psushin:
Shipped!

PR #636, opened 2024-06-03
QT; everyone-share

Krisha11:


yt/yt/server/query_tracker/query_tracker_proxy.cpp
```diff
@@ -46,6 +46,8 @@ static TLogger Logger("QueryTrackerProxy");
 
 namespace NDetail {
 
+constexpr TStringBuf EveryoneShareAccessControlObject = "everyone-share";
```
Krock21, line 49: Can we add a comment above?

`// Hidden from ListQueries, but allows access by query id to [everyone | people who have access to this ACO].`

or

`// This ACO is the same as all others, but does not affect ListQueries. This is needed to enable link sharing.`

Krock21 (approved):

Krisha11:
Shipped!

PR #640, opened 2024-06-04
Add connect timeout

kaikash:
https://github.com/ytsaurus/ytsaurus/issues/641

### DIFF
--- a/yt/yt/core/net/config.cpp
+++ b/yt/yt/core/net/config.cpp
@@ -17,6 +17,7 @@ void TDialerConfig::Register(TRegistrar registrar)
     registrar.Parameter("rto_scale", &TThis::RtoScale)
         .GreaterThan(0.0)
         .Default(2.0);
+    registrar.Parameter("connect_timeout", &TThis::ConnectTimeout);
 
     registrar.Postprocessor([] (TThis* config) {
         if (config->MaxRto < config->MinRto) {
--- a/yt/yt/core/net/config.cpp
+++ b/yt/yt/core/net/config.cpp
@@ -18,7 +18,7 @@ void TDialerConfig::Register(TRegistrar registrar)
         .GreaterThan(0.0)
         .Default(2.0);
     registrar.Parameter("connect_timeout", &TThis::ConnectTimeout)
-        .Default();
+        .Default(TDuration::Seconds(15));
 
     registrar.Postprocessor([] (TThis* config) {
         if (config->MaxRto < config->MinRto) {
--- a/yt/yt/core/net/config.cpp
+++ b/yt/yt/core/net/config.cpp
@@ -17,7 +17,8 @@ void TDialerConfig::Register(TRegistrar registrar)
     registrar.Parameter("rto_scale", &TThis::RtoScale)
         .GreaterThan(0.0)
         .Default(2.0);
-    registrar.Parameter("connect_timeout", &TThis::ConnectTimeout);
+    registrar.Parameter("connect_timeout", &TThis::ConnectTimeout)
+        .Default();
 
     registrar.Postprocessor([] (TThis* config) {
         if (config->MaxRto < config->MinRto) {
--- a/yt/yt/core/net/config.h
+++ b/yt/yt/core/net/config.h
@@ -22,6 +22,7 @@ public:
     TDuration MinRto;
     TDuration MaxRto;
     double RtoScale;
+    std::optional<TDuration> ConnectTimeout;
 
     REGISTER_YSON_STRUCT(TDialerConfig);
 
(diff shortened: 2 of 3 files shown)


yt/yt/core/net/dialer.cpp
```diff
@@ -287,10 +292,14 @@ class TAsyncDialerSession
             return;
         }
 
-        if (Config_->EnableAggressiveReconnect) {
+        if (Config_->EnableAggressiveReconnect || Deadline_) {
+            const auto deadline = Min(
```
maxim-babenko, line 296: Typically we don't use const for locals

yt/yt/core/net/dialer.cpp
```diff
@@ -287,10 +292,14 @@ class TAsyncDialerSession
...
+                Deadline_ ? *Deadline_ : TInstant::Max(),
```
maxim-babenko, line 297: Let's use value_or

yt/yt/core/net/dialer.cpp
```diff
@@ -359,6 +368,16 @@ class TAsyncDialerSession
             Timeout_ *= Config_->RtoScale * GetRandomVariation();
         }
 
+        if (Deadline_ && TInstant::Now() >= *Deadline_) {
+            auto error = TError(NRpc::EErrorCode::TransportError, "Connect timed out")
+                << TErrorAttribute("Timeout", Config_->ConnectTimeout);
```
maxim-babenko, line 373: Timeout -> timeout (attributes are in lower case)

maxim-babenko: "Connect timed out" above vs "Connect timeout" below -- let's choose a single spelling

maxim-babenko (changes requested):

maxim-babenko:
Shipped!

yt/yt/core/net/config.cpp
```diff
@@ -17,6 +17,7 @@ void TDialerConfig::Register(TRegistrar registrar)
     registrar.Parameter("rto_scale", &TThis::RtoScale)
         .GreaterThan(0.0)
         .Default(2.0);
+    registrar.Parameter("connect_timeout", &TThis::ConnectTimeout);
```
maxim-babenko, line 20: Please add ``Default()`` here

maxim-babenko (changes requested):

Kontakter (approved):
To run internal tests

yt/yt/core/net/config.cpp
```diff
@@ -17,6 +17,8 @@ void TDialerConfig::Register(TRegistrar registrar)
     registrar.Parameter("rto_scale", &TThis::RtoScale)
         .GreaterThan(0.0)
         .Default(2.0);
+    registrar.Parameter("connect_timeout", &TThis::ConnectTimeout)
+        .Default();
```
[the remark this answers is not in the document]
kaikash: The reason I haven't set any timeout here is that doing so would change the current behavior of internal clusters and could potentially cause some problems.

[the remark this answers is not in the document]
gritukan: I see. Honestly I do not see any way how reasonable timeout can break something here, but let's wait for code maintainers.

maxim-babenko:
Shipped!

yt/yt/core/net/config.cpp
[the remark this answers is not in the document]
maxim-babenko: 15 secs seems a reasonable limit, let's try this revision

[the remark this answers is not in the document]
gritukan: Discussed with @maxim-babenko, 15 seconds should be reasonable in all cases.

yt/yt/core/net/config.h
```diff
@@ -22,6 +22,7 @@ class TDialerConfig
     TDuration MinRto;
     TDuration MaxRto;
     double RtoScale;
+    std::optional<TDuration> ConnectTimeout;
```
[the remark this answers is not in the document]
kaikash: Makes sense. Also, renamed `Timeout_` to `ReconnectTimeout_` for a better readability.

yt/yt/core/net/dialer.cpp
```diff
@@ -355,12 +359,22 @@ class TAsyncDialerSession
...
+            YT_LOG_ERROR(error);
+            Finished_ = true;
+            guard.Release();
+            OnFinished_(error);
+            return;
         }
 
-        YT_LOG_DEBUG("Connect timeout; trying to reconnect (Timeout: %v)",
-            Timeout_);
+        YT_LOG_DEBUG("Connect timeout; trying to reconnect (Reconnect timeout: %v)",
```
gritukan, line 376: ReconnectTimeout

maxim-babenko:
Shipped!

PR #642, opened 2024-06-05
Increment timeout counter on dns resolve timeout

kaikash:
Seems to be a bug, `TimeoutCounter` is created in `TDnsResolver` but never incremented.

### DIFF
--- a/yt/yt/core/dns/ares_dns_resolver.cpp
+++ b/yt/yt/core/dns/ares_dns_resolver.cpp
@@ -457,11 +457,12 @@ private:
     #endif
     }
 
-    static void OnRequestTimeout(TPromise<TNetworkAddress> promise, TGuid requestId)
+    void OnRequestTimeout(TPromise<TNetworkAddress> promise, TGuid requestId)
     {
         auto timeoutError
             = TError(NNet::EErrorCode::ResolveTimedOut, "Ares DNS resolve timed out");
         if (promise.TrySet(std::move(timeoutError))) {
+            TimeoutCounter_.Increment();
             YT_LOG_WARNING(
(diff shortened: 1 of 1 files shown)


Kontakter (approved):

achulkov2:
Hi! Thank you for fixing this!
Let's cherry-pick this into release branches (both 24.1 and 23.2)?
There have been multiple occasions on which this sensor would have helped us in our installations, and I was thinking of adding this functionality myself. 

Would either @kaikash or @Kontakter be willing to do this?

kaikash:
I think I don’t have enough permissions to cherry-pick to stable branches.

achulkov2:
@kaikash you can create pull requests into stable branches the same way you create PRs into main. robot-piglet will be able to import them into the release branch.

kaikash:
@achulkov2 didn't work xd 

https://github.com/ytsaurus/ytsaurus/pull/664

achulkov2:
@kaikash yeah, it won't work if you use the same branch as in the PR.
You have to checkout a new branch from `stable/xx.y`, cherry-pick your commit from `main` into it (it is best to use the final commit produced by robot-piglet) and then create a PR from that branch into `stable/xx.y`.

kaikash:
Cherry-picking this commit into `stable/23.2` and `stable/24.1` is quite challenging since `TAresDnsResolver` was refactored in the `main` branch. This would require cherry-picking a large chain of other commits, and I'm not sure which commits should be cherry-picked and which should not. Perhaps @Kontakter can help with this?

Another approach would be to address this issue with a new commit for the current version of `TAresDnsResolver` in the `stable/xx.y` branches. However, this means that the `main` branch and the `stable/xx.y` branches will diverge.

PR #645, opened 2024-06-06
[RFC] Remove hard-coded replication tuning for files below 16MiB

koct9i:
This kludge randomly breaks picking for dev cluster with single node.
This is definitely wrong place and wrong way for such tuning.

### DIFF
--- a/yt/python/yt/wrapper/file_commands.py
+++ b/yt/python/yt/wrapper/file_commands.py
@@ -236,14 +236,6 @@ def write_file(destination, stream,
         filename_hint = stream.filename_hint
     if size_hint is None:
         size_hint = stream.size
-    if size_hint is not None and size_hint <= 16 * MB:
-        file_writer = update(
-            {
-                "enable_early_finish": True,
-                "upload_replication_factor": 3,
-                "min_upload_replication_factor": 2,
-            },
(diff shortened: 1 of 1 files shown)


Kontakter:
I understand that it may look like a hack, but unfortunately it is not ok to just delete this code and cannot accept this PR.

I think you should add some configuration flag, enabled by default. That we try to explicitly enable it for inner users and after that disable it by default.

koct9i:
Heh. Probably we could empoy max_replication_factor?

https://github.com/ytsaurus/ytsaurus/blob/10ed460f884b5aeaf797d5dd0a3ebfa16c9a7c11/yt/python/yt/testlib/test_environment.py#L203

Kontakter:
If I understand correclty – this hack is not about tests. It was made for all small files in various APIs.

Kontakter:
The proper solution can be moving this logic (under some option) to the HTTP and RPC proxies

zlobober:
> For more robust solution client must fetch defaults from cluster.
We already have a client config in Cypress introduced by @denvr, btw. Maybe that's the best place?

Kontakter:
I am going to discard this PR, since such a solution cannot be accepted.

PR #667, opened 2024-06-19
Add default value of EnableGeoData to chyt config

thenno:


yt/chyt/controller/internal/chyt/controller.go
```diff
@@ -32,11 +32,13 @@ type Config struct {
 	AddressResolver           map[string]any       `yson:"address_resolver"`
 	EnableYandexSpecificLinks *bool                `yson:"enable_yandex_specific_links"`
 	ExportSystemLogTables     *bool                `yson:"export_system_log_tables"`
+	DefaultEnableGeoData      *bool                `yson:"default_enable_geo_data"`
 }
 
 const (
 	DefaultEnableYandexSpecificLinks = false
 	DefaultExportSystemLogTables     = false
+	DefaultEnableGeoData             = true
```
DimasKovas, line 41: Let's set it to `false` by default as we discussed in the chat

yt/chyt/controller/internal/chyt/controller.go
```diff
@@ -32,11 +32,13 @@ type Config struct {
 	AddressResolver           map[string]any       `yson:"address_resolver"`
 	EnableYandexSpecificLinks *bool                `yson:"enable_yandex_specific_links"`
 	ExportSystemLogTables     *bool                `yson:"export_system_log_tables"`
+	DefaultEnableGeoData      *bool                `yson:"default_enable_geo_data"`
```
DimasKovas, line 35: The naming is inconsistent here.

It should be either just `EnableGeoData` (like `ExportSystemLogTables` option above), or we need to change the naming of the `ExportSystemLogTables` as well

yt/chyt/controller/internal/chyt/controller.go
```diff
@@ -60,6 +62,13 @@ func (c *Config) ExportSystemLogTablesOrDefault() bool {
 	return DefaultExportSystemLogTables
 }
 
+func (c *Config) EnableGeoDataOrDefault() bool {
+	if c.DefaultEnableGeoData != nil {
```
DimasKovas, line 66: The naming of those `OrDefault` functions is `<FieldName>OrDefault()`

DimasKovas (changes requested):

yt/chyt/controller/internal/chyt/controller.go
thenno: Oh, I got the idea. Fixed.

DimasKovas (approved):

PR #423, opened 2024-03-07
Introduce batch_row_count for ordered map/merge

achulkov2:
This PR imlements #375, adding the ability to request jobs with sizes divisible by a certain batch_row_count.
Additionally, it enhances ordered chunk pools to perform chunk splits based on estimated data weight to form jobs with weights almost exactly to equal to data_size_per_job, which was previously only achieved with a certain margin.

Also refactored some of the comments in the touched files into YT style. Left comments in some of the chunk pool unittests unchanged, as there are too many lines.

PR #668, opened 2024-06-19
Fix safe core in ordered chunk pool

achulkov2:
This commit fixes a bug introduced in this [commit](https://github.com/ytsaurus/ytsaurus/commit/b688be4aa792a5367ccd11c5ad429abbb24ed99e) by the new `batch_row_count` related logic. 

The root cause was that the adding an unversioned slice now relies on the fact that the current job is not larger than `data_size_per_job`, i.e. that the job was ended after adding the previous slice.
Meanwhile, the old logic that remained for versioned/unsplittable slices did things in the opposite order: we first checked the size of the current job and added the active slice next. So we would actually end the current job when adding the next slice after the one that hit the limit.
Obviously, these two approaches don't mix well :)
Primarily a problem arises when mixing sorted dynamic and static tables in ordered operations.

I decided to fix this twice:
1. I changed the order in `AddUnsplittablePrimaryDataSlice` to add the slice first and check the current job size later. Now all slice addition methods operate the same way, so everything is fine.
2. I added a safety fallback in `AddSplittablePrimaryDataSlice` that will end the current job if it is too large and reprocess the current slice.

I also addressed new post-commit comments from #423.

### DIFF
--- a/yt/yt/server/lib/chunk_pools/ordered_chunk_pool.cpp
+++ b/yt/yt/server/lib/chunk_pools/ordered_chunk_pool.cpp
@@ -418,19 +418,18 @@ private:
         RowCountUntilJobSplit_ = 0;
         JobCarryOverDataWeight_ = 0;
 
-        IChunkPoolOutput::TCookie result = IChunkPoolOutput::NullCookie;
+        auto dataSliceCopy = CreateInputDataSlice(dataSlice);
+        dataSliceCopy->Tag = cookie;
+        CurrentJob()->AddDataSlice(dataSliceCopy, cookie, /*isPrimary*/ true);
 
         bool jobIsLargeEnough =
             CurrentJob()->GetPreliminarySliceCount() >= JobSizeConstraints_->GetMaxDataSlicesPerJob() ||
             GetCurrentJobDataWeight() >= dataSizePerJob;
         if (jobIsLargeEnough && !SingleJob_) {
-            result = EndJob();
+            return EndJob();
         }
 
-        auto dataSliceCopy = CreateInputDataSlice(dataSlice);
-        dataSliceCopy->Tag = cookie;
-        CurrentJob()->AddDataSlice(dataSliceCopy, cookie, /*isPrimary*/ true);
-        return result;
+        return IChunkPoolOutput::NullCookie;
     }
 
     bool IsDataSliceSplittable(const TLegacyDataSlicePtr& dataSlice) const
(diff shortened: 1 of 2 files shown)


yt/yt/server/lib/chunk_pools/ordered_chunk_pool.cpp
```diff
@@ -441,7 +442,8 @@ class TOrderedChunkPool
 
         // Unbounded dynamic store cannot be split.
         if (dataSlice->GetSingleUnversionedChunkSlice()->GetInputChunk()->IsOrderedDynamicStore() &&
-            !dataSlice->GetSingleUnversionedChunkSlice()->UpperLimit().RowIndex) {
+            !dataSlice->GetSingleUnversionedChunkSlice()->UpperLimit().RowIndex)
```
sergey-v-galtsev, line 445: IMHO
It is better to separate stylistic and substantive changes into different pull requests.

yt/yt/server/lib/chunk_pools/ordered_chunk_pool.cpp
```diff
@@ -484,13 +486,16 @@ class TOrderedChunkPool
         int chunkSliceIndex = 0;
         while (chunkSliceIndex < std::ssize(chunkSlices) || pocket) {
             auto chunkSlice = pocket ? pocket : chunkSlices[chunkSliceIndex];
-            pocket = nullptr;
 
             if (!chunkSlice->GetRowCount()) {
+                // No empty pockets!
+                YT_VERIFY(!pocket);
```
sergey-v-galtsev, line 492: The comment says otherwise.

yt/yt/server/lib/chunk_pools/ordered_chunk_pool.cpp
```diff
@@ -512,16 +517,27 @@ class TOrderedChunkPool
...
+                        "Creating ordered job prematurely (CurrentJobDataWeight: %v, JobCarryOverDataWeight: %v, ActiveSliceDataWeight: %v, ActiveSliceRowCount: %v)",
+                        GetCurrentJobDataWeight(),
+                        JobCarryOverDataWeight_,
+                        chunkSlice->GetDataWeight(),
+                        chunkSlice->GetRowCount());
+                    JobCarryOverDataWeight_ = 0;
+                    EndJob();
+                    // Current job data weight should be zero now.
+                    pocket = chunkSlice;
+                    continue;
```
sergey-v-galtsev, line 540: Looks like `++chunkSliceIndex` is missing.
If this is indeed the case, it might be better to increase it only where it is used
```
auto chunkSlice = pocket ? pocket : chunkSlices[chunkSliceIndex++];
```
Instead of incrementing it on different paths of the loop iteration.

yt/yt/tests/integration/controller/test_merge_operation.py
```diff
@@ -387,6 +387,39 @@ def test_ordered_batch_row_count_disables_teleports_and_sampling(self):
...
+            {"name": "value", "type": "string"},
+        ]
+
+        create_dynamic_table("//tmp/t_dynamic", schema=schema, enable_dynamic_store_read=False)
+        sync_mount_table("//tmp/t_dynamic")
+        dynamic_table_rows = [{"key": i, "value": "some large value"} for i in range(1)]
+        insert_rows("//tmp/t_dynamic", dynamic_table_rows)
+        # Flush.
+        sync_freeze_table("//tmp/t_dynamic")
+        sync_unfreeze_table("//tmp/t_dynamic")
```
sergey-v-galtsev, line 405: `sync_flush_table()`?

yt/yt/server/lib/chunk_pools/ordered_chunk_pool.cpp
achulkov2: Intended meaning: the slice that is saved in `pocket` cannot have zero rows.
It is just a consequence of the logic below and for this reason we do not need to write `if (!pocket) ++chunkSliceIndex`.

I will reword this comment.

sergey-v-galtsev:
Shipped!

PR #676, opened 2024-06-20
Do not pause scheduling on waiting allocations

Krock21:
Do not pause scheduling on waiting allocations

Please cherry-pick to stable/24.1

### DIFF
--- a/yt/yt/server/scheduler/node_shard.cpp
+++ b/yt/yt/server/scheduler/node_shard.cpp
@@ -585,29 +585,20 @@ void TNodeShard::DoProcessHeartbeat(const TScheduler::TCtxNodeHeartbeatPtr& cont
     }
 
     std::vector<TAllocationPtr> runningAllocations;
-    bool hasWaitingAllocations = false;
     YT_PROFILE_TIMING("/scheduler/analysis_time") {
         ProcessHeartbeatAllocations(
             request,
             response,
             node,
             strategyProxy,
-            &runningAllocations,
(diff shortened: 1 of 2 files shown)


yt/yt/server/scheduler/node_shard.cpp
```diff
@@ -597,17 +597,10 @@ void TNodeShard::DoProcessHeartbeat(const TScheduler::TCtxNodeHeartbeatPtr& cont
     }
 
     bool skipScheduleAllocations = false;
-    if (hasWaitingAllocations || isThrottlingActive) {
```
k-pogorelov, line 600: Let's remove hasWaitingAllocations completely

k-pogorelov (changes requested):

yt/yt/server/scheduler/node_shard.cpp
Krock21: We need it to pass it by reference to `ProcessHeartbeatAllocations` on line 596

k-pogorelov: It is true, but look, what we pass it in ProcessHeartbeatAllocations for :)

k-pogorelov (approved):

yt/yt/server/scheduler/node_shard.cpp
Krock21: Removed it from ProcessHeartbeatAllocations too

PR #679, opened 2024-06-23
Add pool and pool tree params to ListOperations

zlobober:
These parameters are long supported by the server-side and by CLI:

```
$ yt list-operations --help
...
  --pool-tree POOL_TREE
                        filter operations by pool tree
  --pool POOL           filter operations by pool. If --pool-tree is set,
                        filters operations with this pool in specified pool
                        tree
...
```

I did not any tests for them, considering that there is no convenient way of running Go SDK tests in OSS. Though I do not expect any surprises here, this PR cannot break any existing test, and the functionality should immediately work.


tinarsky (approved):
LGTM

PR #681, opened 2024-06-24
Detect job OOM

koct9i:
- Detect when any process was killed by OOM killer in user job environment
  Always report here abort reason resource overdraft regardless of
  exit status of job main process - correct handing is not expected.
  
- Detect when job proxy is killed by OOM killer
  Report that abort reason is resource overdraft.
  
- Add option memory_oom_group for killing all tasks at OOM


Kontakter:
@k-pogorelov is reading this PR, comments will be published soon

k-pogorelov (changes requested):

Kontakter:
For transparency: k-pogorelov is on vacation till 4th August. If we need to accept this PR sooner – tell us.

k-pogorelov (approved):

k-pogorelov (approved):

PR #685, opened 2024-06-27
Add yt cache to ya make builds

kozubaeff:
Created YT cluster to store cache for ya make builds and specified configurations in ya.conf


kozubaeff:
To be able to use it, we should add secret YA_YT_TOKEN or YA_YT_TOKEN_PATH to workflows

Kontakter:
These changes should be implemented in such a way, let discard this PR

PR #686, opened 2024-06-28
Add auth env variables to jupyt

thenno:
We've added a custom Authorizer and IdentityProvider to our Jupyter Server. These providers are able to identify users by yt cookies and check permissions by ACO objects.

In this PR I add a new config for JupyterServer.

### DIFF
--- a/yt/chyt/controller/internal/jupyt/configs.go
+++ b/yt/chyt/controller/internal/jupyt/configs.go
@@ -7,19 +7,19 @@ import (
 	"go.ytsaurus.tech/yt/go/yt"
 )
 
-type YTServerConfig struct {
+type jupytServerConfig struct {
 	YTProxy          string `json:"yt_proxy"`
 	YTAuthCookieName string `json:"yt_auth_cookie_name"`
-	YTAcoName        string `json:"yt_aco_name"`
-	YTAcoNamespace   string `json:"yt_aco_namespace"`
-	YTAcoRootPath    string `json:"yt_aco_root_path"`
+	YTACOName        string `json:"yt_aco_name"`
+	YTACONamespace   string `json:"yt_aco_namespace"`
+	YTACORootPath    string `json:"yt_aco_root_path"`
 }
 
 func (c Controller) artifactDir(alias string) ypath.Path {
 	return c.root.Child(alias).Child("artifacts")
 }
 
-func (c *Controller) uploadConfig(ctx context.Context, alias string, filename string, config YTServerConfig) (richPath ypath.Rich, err error) {
+func (c *Controller) uploadConfig(ctx context.Context, alias string, filename string, config jupytServerConfig) (richPath ypath.Rich, err error) {
 	configJson, err := json.MarshalIndent(config, "", "    ")
 	if err != nil {
 		return
--- a/yt/chyt/controller/internal/jupyt/configs.go
+++ b/yt/chyt/controller/internal/jupyt/configs.go
@@ -1,0 +1,46 @@
+package jupyt
+
+import (
+	"context"
+	"encoding/json"
+	"go.ytsaurus.tech/yt/go/ypath"
+	"go.ytsaurus.tech/yt/go/yt"
+)
+
+type YTServerConfig struct {
+	YTProxy          string `json:"yt_proxy"`
+	YTAuthCookieName string `json:"yt_auth_cookie_name"`
+	YTAcoName        string `json:"yt_aco_name"`
+	YTAcoNamespace   string `json:"yt_aco_namespace"`
+	YTAcoRootPath    string `json:"yt_aco_root_path"`
+}
+
+func (c Controller) artifactDir(alias string) ypath.Path {
+	return c.root.Child(alias).Child("artifacts")
+}
+
+func (c *Controller) uploadConfig(ctx context.Context, alias string, filename string, config YTServerConfig) (richPath ypath.Rich, err error) {
+	configJson, err := json.MarshalIndent(config, "", "    ")
+	if err != nil {
+		return
+	}
+	path := c.artifactDir(alias).Child(filename)
+	_, err = c.ytc.CreateNode(ctx, path, yt.NodeFile, &yt.CreateNodeOptions{IgnoreExisting: true})
+	if err != nil {
+		return
+	}
+	w, err := c.ytc.WriteFile(ctx, path, nil)
+	if err != nil {
+		return
+	}
+	_, err = w.Write(configJson)
+	if err != nil {
+		return
+	}
+	err = w.Close()
+	if err != nil {
+		return
+	}
+	richPath = ypath.Rich{Path: path, FileName: filename}
+	return
+}
--- a/yt/chyt/controller/internal/jupyt/controller.go
+++ b/yt/chyt/controller/internal/jupyt/controller.go
@@ -72,6 +72,11 @@ func (c *Controller) Prepare(ctx context.Context, oplet *strawberry.Oplet) (
 
 	var filePaths []ypath.Rich
 
+	err = c.prepareCypressDirectories(ctx, oplet.Alias())
+	if err != nil {
+		return
+	}
+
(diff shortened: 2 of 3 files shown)


yt/chyt/controller/internal/jupyt/controller.go
```diff
@@ -17,6 +16,26 @@ const (
 )
 
 type Config struct {
+	YtAuthCookieName *string           `yson:"yt_auth_cookie_name"`
+	ExtraEnvVars     map[string]string `json:"extra_env_vars"`
```
DimasKovas, line 20: `yson`, not `json`

yt/chyt/controller/internal/jupyt/controller.go
```diff
@@ -17,6 +16,26 @@ const (
 )
 
 type Config struct {
+	YtAuthCookieName *string           `yson:"yt_auth_cookie_name"`
```
DimasKovas, line 19: `YTAuthCookieName`

yt/chyt/controller/internal/jupyt/controller.go
```diff
@@ -31,17 +50,27 @@ func (c *Controller) UpdateState() (changed bool, err error) {
...
-		"NB_USER": "root",
+	jupyterEnv := map[string]string{
+		"NB_GID":                   "0",
+		"NB_UID":                   "0",
+		"NB_USER":                  "root",
+		"JUPYTER_YT_HOST":          c.cluster,
+		"JUPYTER_AUTH_COOKIE_NAME": c.config.YtAuthCookieNameOrDefault(),
+		"JUPYTER_ACO_NAME":         oplet.Alias(),
+		"JUPYTER_ACO_NAMESPACE":    c.Family(),
+		"JUPYTER_ACO_ROOT_PATH":    strawberry.AccessControlNamespacesPath.String(),
```
DimasKovas, line 68: I don't understand the purpose of the `JUPYTER_ACO_ROOT_PATH` variable. It's just a constant, there is no need to pass it through env variables. Access control objects are master objects, not Cypress objects, so they cannot exist in cypress outside the `//sys/access_control_object_namespaces` directory. (this directory is similar to `/proc` on linux - it doesn't really exist in Cypress, but when you try to access it, the master virtually maps existing ACOs and ACO namespaces there)

yt/chyt/controller/internal/jupyt/controller.go
```diff
@@ -17,6 +16,26 @@ const (
...
+		return *c.YtAuthCookieName
+	}
+	return DefaultYtAuthCookieName
+}
+
+func (c *Config) ExtraEnvVarsOrDefault() map[string]string {
+	if c.ExtraEnvVars != nil {
+		return c.ExtraEnvVars
+	}
+	return make(map[string]string)
```
DimasKovas, line 38: No need to allocate an empty map, just `return nil`.
In golang it's perfectly fine to iterate over a nil map (acts like an empty map).
And it's questionable whether we need `OrDefault` method for that if the default is `nil`. May be just use `c.ExtraEnvVars`?

yt/chyt/controller/internal/jupyt/controller.go
```diff
@@ -31,17 +50,27 @@ func (c *Controller) UpdateState() (changed bool, err error) {
...
 		"bash -x start.sh /opt/conda/bin/jupyter lab --ip '*' --port %v --LabApp.token='' --allow-root >&2", JupytPort)
-	return cmd, map[string]string{
-		"NB_GID":  "0",
-		"NB_UID":  "0",
-		"NB_USER": "root",
+	jupyterEnv := map[string]string{
+		"NB_GID":                   "0",
+		"NB_UID":                   "0",
+		"NB_USER":                  "root",
+		"JUPYTER_YT_HOST":          c.cluster,
```
DimasKovas, line 64: Why `JUPYTER_YT_HOST`?
`JUPYTER_YT_PROXY` or `JUPYTER_YT_CLUSTER_PROXY` look more reasonable for me

DimasKovas (changes requested):

yt/chyt/controller/internal/jupyt/controller.go
thenno: Oh, my bad, copilot is need to be double checked:(

yt/chyt/controller/internal/jupyt/controller.go
thenno: On the one hand, I understand that aco_root_path is a constant and unlikely to change. However, I would prefer not to hardcode this knowledge into Jupyter.
While it won't solve any major problems (such as potentially allowing the path overridden on local YT), passing JUPYTER_ACO_ROOT_PATH doesn’t complicate the code much.

I would suggest to keep this environment variable.

zlobober: I'd suggest off-loading all these configuration variables into a config-file for JupYT. Having multitude of environment variables looks less convenient than accepting that there is some (schematized) configuration file, filling it in the controller and attaching it as a regular Cypress file, similar to how it's done for CHYT.

thenno:
FYI:
I still think it's a good idea to keep `ExtraEnvVars` in the controller's config for ad-hoc actions such as enabling debug mode or ad-hoc change log level.

yt/chyt/controller/internal/jupyt/configs.go
```diff
@@ -0,0 +1,46 @@
+package jupyt
+
+import (
+	"context"
+	"encoding/json"
+	"go.ytsaurus.tech/yt/go/ypath"
+	"go.ytsaurus.tech/yt/go/yt"
+)
+
+type YTServerConfig struct {
```
DimasKovas, line 10: The config name doesn't make sense to me.
In CHYT the binary was initially called `ytserver-clickhouse`, that's why in the chyt controller the config is called `ytServerClickHouseConfig`.

Also, we don't need to export this type outside of this package, so it's name should start with lower case letter.

`jupytConfig`? `jupytServerConfig`? `jupYTServerConfig`?

yt/chyt/controller/internal/jupyt/configs.go
```diff
@@ -0,0 +1,46 @@
...
+	YTProxy          string `json:"yt_proxy"`
+	YTAuthCookieName string `json:"yt_auth_cookie_name"`
+	YTAcoName        string `json:"yt_aco_name"`
```
DimasKovas, line 13: Golang [codestyle](https://google.github.io/styleguide/go/decisions#initialisms) dictates abbreviations to be the same case
```
YTACOName
YTACONamespace
YTACORootPath
```

Also names `ACOName`, `ACONamespace`, etc are clear to me without `YT` prefix, so probably we can get rid of it. But both cases are OK to me, so choose whichever you like the most

DimasKovas (changes requested):

yt/chyt/controller/internal/jupyt/configs.go
thenno: I think it would be be better to specify that this value is related to YT.

yt/chyt/controller/internal/jupyt/controller.go
```diff
@@ -107,6 +136,22 @@ func (c *Controller) GetOpBriefAttributes(parsedSpeclet any) map[string]any {
 	return nil
 }
 
+func (c *Controller) appendConfigs(ctx context.Context, oplet *strawberry.Oplet, speclet *Speclet, filePaths *[]ypath.Rich) error {
+	ytServerConfig := jupytServerConfig{
```
DimasKovas, line 140: `ytServerConfig`/`ytServerConfigPath` still look like chyt-specific naming.
But it's a minor style issue, so we can merge the PR even if it isn't fixed

DimasKovas (approved):

yt/chyt/controller/internal/jupyt/controller.go
thenno: I mean "jupyterServerConfig on YT" here (not jupyter kernel). Renamed.

PR #688, opened 2024-07-01
Enable access log from leader in single-master configuration

proller:


yt/yt/server/master/security_server/access_log.cpp
```diff
@@ -278,6 +280,23 @@ bool IsAccessLoggedType(const EObjectType type)
     return typesForAccessLog.contains(type);
 }
 
+bool IsAccessLogEnabled(NCellMaster::TBootstrap* bootstrap) {
```
zlobober, line 283: You can check with the code-style located here: 
https://github.com/ytsaurus/ytsaurus/blob/main/yt/styleguide/cpp.md

In particular, the opening bracket should be on the following line.

yt/yt/server/master/security_server/access_log.cpp
```diff
@@ -278,6 +280,23 @@ bool IsAccessLoggedType(const EObjectType type)
...
+    if (!bootstrap->GetConfigManager()->GetConfig()->SecurityManager->EnableAccessLog) {
+        return false;
+    }
+
+    if (!bootstrap->GetHydraFacade()->GetHydraManager()->IsLeader() &&
+        !bootstrap->GetHydraFacade()->GetHydraManager()->IsRecovery()) {
```
zlobober, line 289: And this opening bracket should also be on the following line, as it is a multi-line if condition. The rest of opening brackets in this function are fine.

yt/yt/server/master/security_server/access_log.h
```diff
@@ -27,9 +25,7 @@
 //! Evaluates an expression the result of which is to be access-logged later.
 //! Crucially, skips the evaluation if access logging is no-op.
 #define YT_EVALUATE_FOR_ACCESS_LOG(...) \
-    ((Bootstrap_->GetConfigManager()->GetConfig()->SecurityManager->EnableAccessLog && \
-        !Bootstrap_->GetHydraFacade()->GetHydraManager()->IsLeader() && \
-        !Bootstrap_->GetHydraFacade()->GetHydraManager()->IsRecovery()) \
+    ((IsAccessLogEnabled(Bootstrap_)) \
```
zlobober, line 28: Looks like an extra pair of parentheses now.

yt/yt/server/master/security_server/access_log.h
```diff
@@ -92,4 +88,6 @@ bool IsAccessLoggedType(const NCypressClient::EObjectType type);
 
 ////////////////////////////////////////////////////////////////////////////////
 
+bool IsAccessLogEnabled(NCellMaster::TBootstrap* bootstrap);
```
zlobober, line 91: Namespace should always terminate with "separator line" aka "otbivka", so either you should but a new line consisting of ///...//, or move this declaration together with some of the declarations above.

zlobober (changes requested):

yt/yt/server/master/security_server/access_log.cpp
```diff
@@ -278,6 +280,25 @@ bool IsAccessLoggedType(const EObjectType type)
...
 }
 
+bool IsAccessLogEnabled(NCellMaster::TBootstrap* bootstrap)
+{
+    if (!bootstrap->GetConfigManager()->GetConfig()->SecurityManager->EnableAccessLog) {
+        return false;
+    }
+
+    if (!bootstrap->GetHydraFacade()->GetHydraManager()->IsLeader() &&
+        !bootstrap->GetHydraFacade()->GetHydraManager()->IsRecovery())
```
gritukan, line 290: Let's not log anything in recovery both in a single-peer and multi-peer cases

gritukan (changes requested):

gritukan (approved):
LGTM!

yt/yt/server/master/security_server/access_log.cpp
```diff
@@ -278,6 +280,27 @@ bool IsAccessLoggedType(const EObjectType type)
...
+
+    if (bootstrap->GetHydraFacade()->GetHydraManager()->IsRecovery()) {
+        return false;
+    }
+
+    if (!bootstrap->GetHydraFacade()->GetHydraManager()->IsLeader()) {
+        return true;
+    }
+
+    if (bootstrap->GetCellManager()->GetTotalPeerCount() <= 1) {
```
maxim-babenko, line 297: I don't quite get it: if there's just one peer then it must be the leader. So why the if above is not taken then?

Anyway, it would be great to see a new test for this change.

gritukan: There is a negation above: we enable logging in case if the peer is not a leader. And also we enable logging here in case if there are no followers because only one peer exists.

maxim-babenko: Indeed!

maxim-babenko:
Shipped!

yt/yt/server/master/security_server/access_log.cpp
```diff
@@ -278,6 +280,27 @@ bool IsAccessLoggedType(const EObjectType type)
     return typesForAccessLog.contains(type);
 }
 
+bool IsAccessLogEnabled(NCellMaster::TBootstrap* bootstrap)
```
raundn, line 283: I'd suggest marking this function inline and moving it to access_log-inl.h.

yt/yt/server/master/security_server/access_log-inl.h
```diff
@@ -0,0 +1,39 @@
+#ifndef ACCESS_LOG_INL_H_
+#error "Direct inclusion of this file is not allowed, include access_log.h"
+// For the sake of sane code completion.
+#include "access_log.h"
+#endif
+
+#include <yt/yt/server/master/cell_master/config.h>
+#include <yt/yt/server/master/cell_master/config_manager.h>
+#include <yt/yt/server/master/cell_master/hydra_facade.h>
+#include <yt/yt/ytlib/election/cell_manager.h>
```
Kontakter, line 10: Add newline between server and ytlib includes, please

Kontakter (approved):
Approve for run tests in internal CI

proller:
fixed ^ ?

Kontakter (approved):

PR #694, opened 2024-07-03
Let jupyt use port from operation env

thenno:


yt/chyt/controller/internal/jupyt/controller.go
```diff
@@ -45,12 +39,10 @@ func (c *Controller) UpdateState() (changed bool, err error) {
 }
 
 func (c *Controller) buildCommand(speclet *Speclet) (command string, env map[string]string) {
-	// TODO(max): take port from YT_PORT_0.
 	// TODO(max): come up with a solution how to pass secrets (token or password) without exposing them in the
 	// strawberry attributes.
 
-	cmd := fmt.Sprintf(
-		"bash -x start.sh /opt/conda/bin/jupyter lab --ip '*' --port %v --LabApp.token='' --allow-root >&2", JupytPort)
+	cmd := "bash -x start.sh /opt/conda/bin/jupyter lab --ip '*' --port $YT_PORT_0 --LabApp.token='' --allow-root >&2"
```
thenno, line 45: I'm not sure this will work. I've checked only `yt vanilla --tasks '{task={job_count=1; command="echo \$YT_PORT_0 >&2"; port_count=1; cpu_limit=1};}' --spec '{resource_limits={user_slots=1}}'`

yt/chyt/controller/internal/jupyt/api.go
```diff
@@ -16,6 +17,33 @@ type GetEndpointResult struct {
 	JobID       yt.JobID       `yson:"job_id"`
 }
 
+func getEndpointPortByJob(ctx context.Context, job yt.JobStatus, ytc yt.Client) (port int, err error) {
+	path := ypath.Path("//sys/exec_nodes").
+		Child(job.Address).
+		Child("/orchid/exec_node/job_controller/active_jobs/").
```
DimasKovas, line 23: First, it doesn't look right. There should be no `/` in the beginning and at the end of an argument.
Second, you have many `.Child()` here. The correct method to use is `.JoinChild` to join multiple children.

yt/chyt/controller/internal/jupyt/api.go
```diff
@@ -36,13 +64,20 @@ func GetEndpoint(a *api.API, ctx context.Context, alias string) (result GetEndpo
 		return
 	}
 
-	address, _, err := net.SplitHostPort(jobs.Jobs[0].Address)
+	job := jobs.Jobs[0]
+
+	address, _, err := net.SplitHostPort(job.Address)
```
DimasKovas, line 69: FYI: This works only if the job is running without any network isolation. Otherwise the job address differs from the node address.

Do not need to fix it now, because it hadn't worked before you changed the code either. But it makes sense to leave a comment for those who will read this code later.

DimasKovas (changes requested):

yt/chyt/controller/internal/jupyt/controller.go
thenno: it works

DimasKovas (approved):

PR #703, opened 2024-07-10
describe_options for jupyt

dmi-feo:


yt/chyt/controller/internal/jupyt/controller.go
```diff
@@ -123,7 +132,48 @@ func (c *Controller) ParseSpeclet(specletYson yson.RawValue) (any, error) {
...
+		{
+			Title: "Jupyt params",
+			Options: []strawberry.OptionDescriptor{
+				{
+					Title:        "Docker image",
+					Name:         "jupyter_docker_image",
+					Type:         strawberry.TypeString,
+					CurrentValue: speclet.JupyterDockerImage,
+					DefaultValue: c.config.PreferredDockerImageOrDefault(),
+					Description:  "A docker image containing jupyt and required stuff",
```
DimasKovas, line 147: .

yt/chyt/controller/internal/jupyt/controller.go
```diff
@@ -123,7 +132,48 @@ func (c *Controller) ParseSpeclet(specletYson yson.RawValue) (any, error) {
...
+	return []strawberry.OptionGroupDescriptor{
+		{
+			Title: "Jupyt params",
+			Options: []strawberry.OptionDescriptor{
+				{
+					Title:        "Docker image",
+					Name:         "jupyter_docker_image",
+					Type:         strawberry.TypeString,
+					CurrentValue: speclet.JupyterDockerImage,
+					DefaultValue: c.config.PreferredDockerImageOrDefault(),
```
DimasKovas, line 146: If you state that it's a default value, you need to use it in the code as a default value.

And since `JupyterDockerImage` is a `string` now (not a `*string`), the `CurrentValue` will be an empty string, not a `nil`. So UI will not show you that default value either.

DimasKovas (changes requested):

yt/chyt/controller/internal/jupyt/controller.go
dmi-feo: My bad, it's not the place for default docker image at all. I deleted the default value

DimasKovas (approved):

PR #705, opened 2024-07-11
Properly handle unknown GPU device

gritukan:
In case if GPU info provider returns information about the device that was not observed in `ListGpuDevices` node reacts in a weird way: newly observed GPU will be reported to a scheduler but job requiring this GPU will be aborted since no corresponding GPU slot exists.

### DIFF
--- a/yt/yt/server/node/exec_node/gpu_manager.cpp
+++ b/yt/yt/server/node/exec_node/gpu_manager.cpp
@@ -188,6 +188,8 @@ TGpuManager::TGpuManager(IBootstrap* bootstrap)
     auto now = TInstant::Now();
     for (const auto& descriptor : descriptors) {
         GpuDevices_.push_back(descriptor.DeviceName);
+        InsertOrCrash(GpuDeviceIndices_, descriptor.DeviceIndex);
+
         FreeSlots_.emplace_back(descriptor.DeviceIndex);
 
         EmplaceOrCrash(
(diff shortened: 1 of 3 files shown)


Kontakter:
It should be reviewed by @eshcherbin as a responsible for GPU provider logic.

yt/yt/server/node/exec_node/gpu_manager.cpp
```diff
@@ -332,6 +335,13 @@ void TGpuManager::OnHealthCheck()
             }
 
             for (auto& gpuInfo : gpuInfos) {
+                if (!GpuDeviceIndices_.contains(gpuInfo.Index)) {
+                    YT_LOG_WARNING("Found unknown GPU device (DeviceName: %v)",
+                        GetGpuDeviceName(gpuInfo.Index));
```
eshcherbin, line 340: I'd prefer logging device index instead of using `GetGpuDeviceName`, but I understand we already do the latter for lost devices, so no call to action here.

Ideally, we should revise device identification and use device index (aka minor number) or UUID everywhere.

eshcherbin (approved):

Kontakter (approved):

yt/yt/server/node/exec_node/gpu_manager.cpp
gritukan: Will do it in a separate PR!

PR #731, opened 2024-07-24
Promote YDB commit for fix YQL Skiff writer column order; add tests

gritukan:


Krock21 (approved):

.github/workflows/query_tracker_release.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: ydb-platform/ydb
-          ref: 79a0619ccd71da017e02fc8ff64c1019141030ea
+          ref: 906c9369de7871e83ed54d43d890334e859d5939
```
Krock21, line 54: Were you able to test this commit anywhere?
I think it will be broken without a patch [here](https://github.com/ydb-platform/ydb/blob/906c9369de7871e83ed54d43d890334e859d5939/yt/yt/core/bus/public.h#L36) (Adding Journal 4 and RealTime 5)

Krock21: I see the [breaking change was reverted](https://github.com/ytsaurus/ytsaurus/commit/9ee7a14dca2abd905c23dd1181f976031f5ac575) so may be fine now

gritukan: > Were you able to test this commit anywhere?

I tested it locally and everything was fine because of the revert you mentioned above.

psushin:
Shipped!

PR #732, opened 2024-07-24
Introduce local_exported_object_count attribute and use it in the chunk teleporter

gritukan:
Using exported_object_count attribute for validation of the number of teleported chunks is incorrect since the value of this attribute for primary master is a total number of exported object among all the master cells not just primary.


maxim-babenko:
Shipped

PR #749, opened 2024-07-31
Add hosts path config option in go client

kaikash:
In the C++ and Python SDKs, this option can be configured. However, in the Go SDK, it is currently hard-coded to `hosts`.

### DIFF
--- a/yt/go/yt/config.go
+++ b/yt/go/yt/config.go
@@ -43,6 +43,11 @@ type Config struct {
 	// If not set, default role is used.
 	ProxyRole string
 
+	// HostsPath specifies the path used to discover HTTP proxies.
+	//
+	// If not set, defaults to "hosts".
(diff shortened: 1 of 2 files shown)


tinarsky (approved):
LGTM

tinarsky:
@kaikash, what exactly do you need this parameter for? Are there no other fields in the config (e.g. ProxyRole) to solve your problem?

kaikash:
> @kaikash, what exactly do you need this parameter for? Are there no other fields in the config (e.g. ProxyRole) to solve your problem?

No, `proxy role` is not what we need. This is analog of `network_name` in rpc proxies. We have multiple networks and in each network proxy address is different. Just like in your installations you have `/hosts/fb` endpoint, we have `/hosts/<network_name>` endpoints in our proxies.

PR #753, opened 2024-08-01
Add System Python UDFs to yql agent

Krock21:
A branch with enabled CI (this PR + CI enable commit): https://github.com/ytsaurus/ytsaurus/tree/members/krock21/systempython-udf-1-test-workflow

Tested on internal clusters([CI run](https://github.com/ytsaurus/ytsaurus/actions/runs/10198298190)). It works

### DIFF
--- a/yt/docker/ytsaurus/Dockerfile
+++ b/yt/docker/ytsaurus/Dockerfile
@@ -116,6 +116,9 @@ RUN chmod 755 /setup_cluster_for_chyt.sh
 
 FROM base-ytsaurus-python-packages AS query-tracker-bloated
 
+# Libpython shared objects. TODO(aleksandr.gaev) use stubs instead.
+RUN apt-get install -y libpython3.8 libpython3.9 libpython3.10 libpython3.11 libpython3.12
+
 # Query tracker binaries.
 COPY ./ytserver-all /usr/bin/ytserver-all
 RUN ln -s /usr/bin/ytserver-all /usr/bin/ytserver-query-tracker
--- a/.github/workflows/build-jobs.yaml
+++ b/.github/workflows/build-jobs.yaml
@@ -299,3 +299,7 @@ jobs:
           do
             ydb-ya-make  ${udfs_prefix}/${udf_name}
           done
+
+          export YDB_SOURCE_ROOT=$(realpath $(pwd)/../ytsaurus/ydb)
+          export YA_BUILD_PATH=$(pwd)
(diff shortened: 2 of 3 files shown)


Krock21:
CI run: https://github.com/ytsaurus/ytsaurus/actions/runs/10198191440

Krock21:
Rebased to pick up the latest YDB commit
CI run: https://github.com/ytsaurus/ytsaurus/actions/runs/10198298190

Krock21:
CI run: https://github.com/ytsaurus/ytsaurus/actions/runs/10198298190

Tested on internal cluster. Works with `execute_udf_locally_if_possible = %false` in yql agent gateway spec

Krisha11:
Lets make it false in trunk to make it workable in result builds?

https://github.com/ytsaurus/ytsaurus/blob/main/yt/yql/agent/config.cpp#L270

Krock21:
> Lets make `execute_udf_locally_if_possible` `false` in trunk to make it workable in result builds?
> 
> https://github.com/ytsaurus/ytsaurus/blob/main/yt/yql/agent/config.cpp#L270

https://github.com/ytsaurus/ytsaurus/pull/754

yt/docker/ytsaurus/Dockerfile
```diff
@@ -116,6 +116,9 @@ RUN chmod 755 /setup_cluster_for_chyt.sh
 
 FROM base-ytsaurus-python-packages AS query-tracker-bloated
 
+# Libpython shared objects. TODO(aleksandr.gaev) use stubs instead.
```
Krisha11, line 119: What does this TODO mean?

Krock21: It means that we don't need working python and `libpython3.N.so` files in this image, we only need stubs that declare required symbols

I was creating stubs using this script(passing libpython3.N.so as a first argument):
```
#/usr/bin/env bash
SYMBOLS=$(nm -D --defined-only "$1" | awk '{print $3}')
NAME=$(basename "$1")
echo "#include <stdlib.h>" > ${NAME}_stub.c
for SYMBOL in $SYMBOLS
do
        echo "void $SYMBOL(){exit(42);};" >> ${NAME}_stub.c
done

cc -shared ${NAME}_stub.c -o ${NAME}
```
and it worked on an internal cluster

Krock21: We may implement something similar to `YQL-7971`, when stubs are created automatically and added to yql_agent
In my tests I added them to `/usr/local/lib` and it made it worse for some other tools that required libpython (gdb)

Krisha11 (approved):

PR #754, opened 2024-08-02
Change default value of execute_udf_locally_if_possible to false

Krock21:
This flag allows untrusted udfs to be executed locally (in yqlAgent). We don't want it by default since it can break yqlAgent or interfere with other queries


Krisha11:
Change here too, please
https://github.com/ytsaurus/ytsaurus/blob/996703963c7266c61fec82edec7f69320e92c13c/yt/yql/tests/test_simple.py#L344

Krock21:
> Change here too, please

Yes, did in the latest push

Krisha11 (approved):

Issue #709, opened 2024-07-12
Update doc for use chyt with datagrip


ykuc:
Description in the current doc not work

kozubaeff:
Modern versions of Datagrip use Clickhouse versions currently incompatible with our http proxy. We will update our documentation as soon as we will fix this

ykuc:
@kozubaeff created [PR](https://github.com/ytsaurus/ytsaurus/pull/755) for update doc

PR #755, opened 2024-08-02
add doc for use chyt in datagrip

ykuc:
Update doc for use chyt in datagrip
https://github.com/ytsaurus/ytsaurus/issues/709


kozubaeff:
Hi, and we haven't posted the http protocol fix yet and as a consequence the doc will be irrelevant

Issue #451, opened 2024-03-13
Support optimize_read_in_order in CHYT


achulkov2:
When using CH queries over logs/events stored in YT it is often necessary to run queries along the lines of the following:
```
SELECT *
FROM //table_sorted_by_key_columns
WHERE <filter on some columns that is satisfied by many rows>
ORDER BY <key_columns> <ASC/DSC>
LIMIT 50
```
NB: Note that the input table is sorted by exactly the same columns as the ones specified in the order by clause.

The table is typically stored in the columnar format, with only a few of the columns containing most of the data weight. 
The filter is often quite wide: e.g. log_level = info, event_type = create_vm, etc.
A typical example of a key column is just a single timestamp column.

It would make sense for the query above to stop its execution when 50 rows, satisfying the filter are collected. This is allowed, since the table is sorted by the exact set of keys specified in the `ORDER BY` clause.

Native ClickHouse provides the [optimize_read_in_order](https://clickhouse.com/docs/en/sql-reference/statements/select/order-by#optimization-of-data-reading) option, which should be effective in this case.
However, it only [seems](https://github.com/search?q=repo%3AClickHouse/ClickHouse%20requestReadingInOrder&type=code) to be supported for specific table engines.

**The general goal is to support the native `optimize_read_in_order` option in the CHYT table engine.**

Together with @DimasKovas we propose the following plan:

1. Support ORDER BY ASC:
    - Add CHYT `read_in_order` option. The behaviour below will be enabled by this option, but only if the set of columns in the `ORDER BY` clause is equal to the table's key column vector.
    - The coordinator will create a single separate subquery for each input table **dropping** the `ORDER BY` clause and without shuffling any input chunks around.
    - The coordinator will schedule these queries across instances. In case of a single instance, multiple subqueries will be handled by the same instance.
    - Each instance will use a single reading thread and configure its reader in such a way, that data is fed into the resulting stream in the table order.
2. Support ORDER BY DSC:
    - When dropping the `ORDER BY` clauses, the coordinator will pass along a flag in case a descending order was requested.
    - The instances will use this flag to trigger a reading order in which data is fed in reverse order from the end of the table. This will have to be implemented as some sort of ChunkReader/BlockFetcher wrapper.
3. Support the native ClickHouse option:
    - Generalise the native ClickHouse code to support making the `optimize_read_in_order` induced changes to the query execution plan for custom engines. **This would be a non-trivial change to the CH upstream.**
[...]

PR #757, opened 2024-08-05
Introduce enable_optimize_read_in_order in CHYT

achulkov2:
This PR introduces a new mode in CHYT that aims to mimic the behaviour of [optimize_read_in_order](https://clickhouse.com/docs/en/sql-reference/statements/select/order-by#optimization-of-data-reading) in native ClickHouse. This PR implements #451, albeit using a different approach.
It works by using a sorted pool to produce jobs of exponentially increasing/decreasing sizes, which turn into subqueries that are processed combined sequentially without additional sorting steps. This should allow queries described in #451 to execute without reading the whole table. 
Notable modifications with generic impact:
* Support for exponentially increasing job sizes in new sorted pool and related classes.
* Chunk slice fetcher is now used when creating a sorted pool in CHYT.
* New column in query log table exporter.

### DIFF
--- a/yt/chyt/server/config.cpp
+++ b/yt/chyt/server/config.cpp
@@ -141,9 +141,8 @@ void TExecutionSettings::Register(TRegistrar registrar)
     registrar.Parameter("enable_min_max_filtering", &TThis::EnableMinMaxFiltering)
         .Default(true);
 
-    // TODO(achulkov2): Set to false by default. This is temporary to run all tests and verify that nothing fails.
     registrar.Parameter("enable_optimize_read_in_order", &TThis::EnableOptimizeReadInOrder)
-        .Default(true);
+        .Default(false);
     registrar.Parameter("assume_no_null_keys", &TThis::AssumeNoNullKeys)
         .Default(false);
     registrar.Parameter("assume_no_nan_keys", &TThis::AssumeNoNanKeys)
--- a/yt/chyt/server/config.cpp
+++ b/yt/chyt/server/config.cpp
@@ -140,6 +140,10 @@ void TExecutionSettings::Register(TRegistrar registrar)
 
     registrar.Parameter("enable_min_max_filtering", &TThis::EnableMinMaxFiltering)
         .Default(true);
+
+    // TODO(achulkov2): Set to false by default. This is temporary to run all tests and verify that nothing fails.
+    registrar.Parameter("enable_optimize_read_in_order", &TThis::EnableOptimizeReadInOrder)
+        .Default(true);
 }
 ////////////////////////////////////////////////////////////////////////////////
 
@@ -280,6 +284,8 @@ void TShowTablesConfig::Register(TRegistrar registrar)
 
 void TSubqueryConfig::Register(TRegistrar registrar)
 {
+    registrar.Parameter("columnar_statistics_fetcher", &TThis::ColumnarStatisticsFetcher)
+        .DefaultNew();
     registrar.Parameter("chunk_slice_fetcher", &TThis::ChunkSliceFetcher)
         .DefaultNew();
     registrar.Parameter("max_job_count_for_pool", &TThis::MaxJobCountForPool)
--- a/yt/chyt/server/config.cpp
+++ b/yt/chyt/server/config.cpp
@@ -144,6 +144,10 @@ void TExecutionSettings::Register(TRegistrar registrar)
     // TODO(achulkov2): Set to false by default. This is temporary to run all tests and verify that nothing fails.
     registrar.Parameter("enable_optimize_read_in_order", &TThis::EnableOptimizeReadInOrder)
         .Default(true);
+    registrar.Parameter("assume_no_null_keys", &TThis::AssumeNoNullKeys)
+        .Default(false);
+    registrar.Parameter("assume_no_nan_keys", &TThis::AssumeNoNanKeys)
+        .Default(false);
 }
 ////////////////////////////////////////////////////////////////////////////////
 
--- a/yt/chyt/server/job_size_constraints.cpp
+++ b/yt/chyt/server/job_size_constraints.cpp
@@ -105,7 +105,7 @@ TClickHouseJobSizeSpec CreateClickHouseJobSizeSpec(
         jobCount += jobSizeTrackerOptions.LimitProgressionLength - 1;
 
         // This helps us actually produce small jobs when we need them.
-        inputSliceDataWeight = std::max<int>(1, startDataWeightPerJob * 0.1);
+        inputSliceDataWeight = std::max<i64>(1, startDataWeightPerJob * 0.1);
     }
 
     // It is important to watch the ratio of total data weight to input slice data weight, to avoid dealing with too many slices.
--- a/yt/chyt/server/job_size_constraints.cpp
+++ b/yt/chyt/server/job_size_constraints.cpp
@@ -1,33 +1,43 @@
 #include "job_size_constraints.h"
 
+#include <yt/yt/server/lib/chunk_pools/job_size_tracker.h>
+
 #include <yt/yt/server/lib/controller_agent/job_size_constraints.h>
 
 #include "config.h"
 
 namespace NYT::NClickHouseServer {
 
+using namespace NChunkPools;
 using namespace NControllerAgent;
 using namespace NLogging;
 
 ////////////////////////////////////////////////////////////////////////////////
 
+constexpr int MaxTotalSliceCount = 1'000'000;
+
+////////////////////////////////////////////////////////////////////////////////
+
 // TODO(max42): consider introducing new job size constraints to encapsulate all these heuristics.
 
-IJobSizeConstraintsPtr CreateClickHouseJobSizeConstraints(
-    TSubqueryConfigPtr config,
+TClickHouseJobSizeSpec CreateClickHouseJobSizeSpec(
+    const TExecutionSettingsPtr& executionSettings,
+    const TSubqueryConfigPtr& subqueryConfig,
     i64 totalDataWeight,
     i64 totalRowCount,
     int jobCount,
     std::optional<double> samplingRate,
+    EReadInOrderMode readInOrderMode,
     const TLogger& logger)
 {
     const auto& Logger = logger;
 
+    YT_VERIFY(jobCount);
     auto dataWeightPerJob = totalDataWeight / jobCount;
 
     if (samplingRate) {
         double rate = samplingRate.value();
-        auto adjustedRate = std::max(rate, static_cast<double>(jobCount) / config->MaxJobCountForPool);
+        auto adjustedRate = std::max(rate, static_cast<double>(jobCount) / subqueryConfig->MaxJobCountForPool);
         auto adjustedJobCount = std::floor(jobCount / adjustedRate);
         YT_LOG_INFO("Adjusting job count and sampling rate (OldSamplingRate: %v, AdjustedSamplingRate: %v, OldJobCount: %v, AdjustedJobCount: %v)",
             rate,
@@ -38,7 +48,7 @@ IJobSizeConstraintsPtr CreateClickHouseJobSizeConstraints(
         jobCount = adjustedJobCount;
     } else {
         // Try not to form too small ranges when total data weight is small.
-        auto maxJobCount = totalDataWeight / std::max<i64>(1, config->MinDataWeightPerThread) + 1;
+        auto maxJobCount = totalDataWeight / std::max<i64>(1, subqueryConfig->MinDataWeightPerThread) + 1;
         if (maxJobCount < jobCount) {
             jobCount = maxJobCount;
             dataWeightPerJob = std::max<i64>(1, totalDataWeight / maxJobCount);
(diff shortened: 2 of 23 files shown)


yt/chyt/server/config.cpp
```diff
@@ -140,6 +140,10 @@ void TExecutionSettings::Register(TRegistrar registrar)
 
     registrar.Parameter("enable_min_max_filtering", &TThis::EnableMinMaxFiltering)
         .Default(true);
+
+    // TODO(achulkov2): Set to false by default. This is temporary to run all tests and verify that nothing fails.
+    registrar.Parameter("enable_optimize_read_in_order", &TThis::EnableOptimizeReadInOrder)
+        .Default(true);
```
achulkov2, line 146: This is set to `true` temporarily, so that we run all CHYT tests in the import and see that nothing breaks.

yt/chyt/server/query_analyzer.cpp
```diff
@@ -728,6 +744,125 @@ void TQueryAnalyzer::OptimizeQueryProcessingStage()
...
+                THROW_ERROR_EXCEPTION_IF(
+                    orderByElement->nulls_direction_was_explicitly_specified,
+                    "Specified nulls direction is incompatible with enabled optimize_read_in_order mode; "
+                    "please turn it off via query settings, specify another nulls direction or use the default nulls direction");
+                YT_LOG_DEBUG(
+                    "Reversing ORDER BY nulls direction (ColumnName: %v, Direction: %v, NullsDirection: %v -> -1)",
+                    orderByElement->children.front()->getColumnName(),
+                    orderByElement->direction,
+                    orderByElement->nulls_direction);
+                orderByElement->nulls_direction = -1;
```
DimasKovas, line 853: Looks suspicious.
First, I'm not sure if we are allowed to change the query AST. I have no idea how CH code would handle this "unexpected change".
Second, it could change the query results compared to the execution without the optimization.

The second point would prevent this optimization from being turned on by default. If this behavior is desired, we could introduce a separate option to configure it.

yt/chyt/server/query_analyzer.cpp
```diff
@@ -728,6 +744,125 @@ void TQueryAnalyzer::OptimizeQueryProcessingStage()
...
+    // In YT, NULL values compare less than any other values. In ClickHouse, NULL value placement
+    // depends solely on the placement requested in the query, which defaults to NULLS LAST.
+    // Since we combine both YT and CH sorting in this mode, there is only one acceptable direction
+    // depending on the requested sort order. In cases when the unsupported option turns out to be
+    // the default (for ascending sorts), we tweak the query to avoid additional user turmoil.
+    if (commonDirection != EReadInOrderMode::None) {
+        for (const auto& orderByElementAst : selectQuery->orderBy()->children) {
+            auto orderByElement = orderByElementAst->as<DB::ASTOrderByElement>();
+
+            // The value of nulls_direction is equal to direction for NULLS LAST and to the opposite
```
DimasKovas, line 839: We can skip `required` columns. Don't need to required 'nulls first/last` for them

yt/chyt/server/query_analyzer.cpp
```diff
@@ -728,6 +744,125 @@ void TQueryAnalyzer::OptimizeQueryProcessingStage()
...
+
+        // Judging by CH code and experiments, ASTOrderByElement always has a single child.
+        if (std::ssize(orderByElement->children) != 1) {
+            THROW_ERROR_EXCEPTION("Query ORDER BY ast is malformed; "
+                "this is a bug; please contact cluster administrators");
+        }
+
+        // The name can contain functions. While it is possible to support tuples
+        // and even monotonic functions, I do not think there is much use.
+        auto columnName = orderByElement->children.front()->getColumnName();
```
DimasKovas, line 797: Need to check that the orderByElement is a `column` somehow before accessing it.

We don't want to break queries with an expression in the `order by` clause.

yt/chyt/server/query_analyzer.cpp
```diff
@@ -728,6 +744,125 @@ void TQueryAnalyzer::OptimizeQueryProcessingStage()
...
+                    orderByElement->nulls_direction);
+                orderByElement->nulls_direction = -1;
+                orderByElement->nulls_direction_was_explicitly_specified = true;
+            }
+        }
+    }
+    // Sadly, with NANs we are just screwed. In YT these values are larger than any others, but in
+    // CHYT they are always placed together with NULLs. These two approaches contradict each other.
+    // To warn about such cases without disabling this optimization for tables with floating-point
+    // key columns completely, we require to specify nulls direction explicitly as a green light.
```
DimasKovas, line 861: I offer to introduce a separate option for `float` behavior too, so we can safely turn `optimize_read_in_order` on by default

yt/chyt/server/subquery.cpp
```diff
@@ -254,7 +255,7 @@ class TInputFetcher
                     EKeyConditionScale::TopLevelDataSlice,
                     dataSlice->LowerLimit().KeyBound,
                     dataSlice->UpperLimit().KeyBound,
-                    dataSlice->GetInputStreamIndex()).can_be_true;
+                    operandIndex).can_be_true;
```
DimasKovas, line 258: I don't understand why this change is needed

yt/chyt/server/subquery.cpp
```diff
@@ -426,7 +427,7 @@ class TInputFetcher
 
             for (auto& dataSlice : InputDataSlices_[tableIndex]) {
                 YT_VERIFY(!dataSlice->IsLegacy);
-                dataSlice->SetInputStreamIndex(operandIndex);
+                dataSlice->SetInputStreamIndex(tableIndex);
```
DimasKovas, line 430: And I don't understand that one either

yt/yt/server/lib/chunk_pools/job_size_tracker.cpp
```diff
@@ -143,6 +147,26 @@ class TJobSizeTracker
             return;
         }
 
+        if (Options_.LimitProgressionRatio && LimitProgressionLength_ < Options_.LimitProgressionLength) {
```
DimasKovas, line 150: I don't understand what is going on here, but if it works as expected, that's fine

DimasKovas (changes requested):

yt/chyt/server/query_context.h
```diff
@@ -278,3 +281,7 @@ void InvalidateCache(
 ////////////////////////////////////////////////////////////////////////////////
 
 } // namespace NYT::NClickHouseServer
+
+#define QUERY_CONTEXT_INL_H_
+#include "query_context-inl.h"
```
DimasKovas, line 286: There is no `query_context-inl.h` in this PR

achulkov2: That's a pity! Added id, so you can import again for tests.

yt/chyt/server/subquery.cpp
achulkov2: This is the consequence of the change in input stream index (it used to reflect the operand index and now it doesn't).

yt/chyt/server/query_context-inl.h
```diff
@@ -0,0 +1,28 @@
...
+            key,
+            NYson::ConvertToYsonString(RuntimeVariables_->GetYson(key), NYson::EYsonFormat::Text),
+            NYson::ConvertToYsonString(value, NYson::EYsonFormat::Text));
+    }
+    RuntimeVariables_->Set(key, value);
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+} // namespace NYT::NClickHouseServer
```
DimasKovas, line 28: '\n'

DimasKovas (approved):
To run tests

yt/chyt/server/subquery.cpp
achulkov2: In order for sorted pool to work properly, we need to attribute slices to correct input streams: the pool uses these streams to decide whether to use chunk slice fetcher (used for unversioned streams only) and in the end it validates that output slices from a single stream are actually sorted. I am pretty sure these indexes are also used while building jobs, otherwise it would be impossible to have multiple potentially intersecting sorted tables as input.

Previously, we used the operand index as input stream index, so all concatenated tables seemed like one big table to the pool. This has no chances of working, since this "one big table" is not actually sorted.

yt/chyt/server/query_analyzer.cpp
achulkov2: For the first point, I don't know how to check whether it is okay to change it like that or not. I can say that first I only changed `nulls_direction` and it didn't have any effect, but once you also set `nulls_direction_was_explicitly_set` it starts having effect. There are tests for this and they pass. 
Also there are already some AST changes nearby in `Analyze`, so I think it should be fine...

achulkov2: As for the second point, my intent was exactly so that we could enable it by default and the users wouldn't have to manually write `NULLS FIRST` or `NULLS LAST` in their queries. 
I agree that this is all quite annoying. I am also concerned that optional key columns are probably not uncommon, since that is pretty much what you get by default.

I can introduce a separate option that would enable this behaviour and turn it off by default. If it is disabled, then we will throw an exception here instead of changing the AST. However, I still want to change the AST, because enabling this extra option is much easier for the user than thinking about the correct NULLS direction. It is also much easier to copy and paste an enabled option from query to query. 

What do you think?

yt/chyt/server/query_analyzer.cpp
achulkov2: We shouldn't? A few lines below we compare this name to the name of the corresponding column (by index) in the schema. This should never be true for expressions, so we won't enable the optimization. There is one test for that.

yt/yt/server/lib/chunk_pools/job_size_tracker.cpp
achulkov2: The class itself is quite complicated, yeah, it took me a while. You should double check the general intent and whether I computed parameters correctly. 
I spent quite some time debugging errors around this and idk if there are more mistakes.

yt/chyt/server/config.cpp
achulkov2: > Why not to use native ClickHouse's enable_optimize_read_in_order option?
Do we have the ability to enable/disable native CH-options for the whole clique?

achulkov2: >Did you measure the worst possible performance degradation from this optimization? If the degradation is not significant, we can leave it on by default.

What is the best way to go about measuring this? Do we have perf tests? :)

yt/chyt/server/job_size_constraints.cpp
```diff
@@ -49,11 +59,78 @@ IJobSizeConstraintsPtr CreateClickHouseJobSizeConstraints(
...
+        switch (readInOrderMode) {
+            case EReadInOrderMode::Forward:
+                // The first range is small!
+                dataWeightPerJob = startDataWeightPerJob;
+                break;
+            case EReadInOrderMode::Backward:
+                // We will be reading these ranges backwards!
+                jobSizeTrackerOptions.LimitProgressionRatio = 0.5;
+                // We need to perform an actual split for the optimization to work.
+                if (jobCount == 1) {
```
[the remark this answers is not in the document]
achulkov2: No, because in forward mode we start with `startDataWeightPerJob`, which is small, instead of `dataWeightPerJob` computed above. If this small size is larger than the entire table, then that's fine and we only need one small job. 

The problem in backward mode is that we need to produce a few slices first before starting our exponential descent, and here we have no chance of doing that if we put everything in one single slice in the very beginning.

yt/chyt/server/config.cpp
DimasKovas: > Do we have the ability to enable/disable native CH-options for the whole clique?

Yes, we do. The default CH settings may be specified in the cliques's config like this: `{clickhouse={settings={enable_optimize_read_in_order=1}}}`

But afaik, it's turned on by default in CH

> What is the best way to go about measuring this? Do we have perf tests? :)

Unfortunately, we don't have perf tests yet. I think that a simple query on a large table, with all the rows being filtered out, would give us an idea of how bad the performance degradation might be.

DimasKovas (changes requested):
There are quite a few tests failing after this change.

Here is the list:
```
test_atomicity.py::TestClickHouseAtomicity test_join_with_locks (gudqeit)
test_input_fetching.py::TestInputFetching[0] test_reading_dynamic_and_static_tables (gudqeit)
test_input_fetching.py::TestInputFetching[0] test_reading_multiple_dynamic_tables (gudqeit)
test_join_and_in.py::TestJoinAndInStress test_sorted_join_stress (max42)
test_join_and_in.py::TestJoinAndIn[0] test_complex_join_key (dakovalkov)
test_join_and_in.py::TestJoinAndIn[0] test_filter_joined_subquery (dakovalkov)
test_join_and_in.py::TestJoinAndIn[0] test_join_table_and_array_join (dakovalkov)
test_join_and_in.py::TestJoinAndIn[0] test_tricky_join (max42)
test_join_and_in.py::TestJoinAndIn[1] test_join_dynamic_tables_with_dynamic_stores (max42)
test_join_and_in.py::TestJoinAndIn[1] test_join_for_unsorted_tables (gudqeit)
test_join_and_in.py::TestJoinAndIn[1] test_sorted_join_simple (max42)
test_join_and_in.py::TestJoinAndIn[1] test_sorted_join_with_wider_key_condition (max42)
test_join_and_in.py::TestJoinAndIn[1] test_tricky_join2 (max42)
```

yt/chyt/server/query_analyzer.cpp
DimasKovas: I expected it to throw an error if the `orderByElement` is not a column literal. But yeah, since the tests work as expected, it's not the case.

Anyway, imho, an explicit check that the element is a literal would show your intent clearer in the code.

achulkov2:
> There are quite a few tests failing after this change.

I fixed a stupid bug, let's import again?

DimasKovas (approved):
To run tests

achulkov2:
I fixed another bug in which I moved `queryAnalyzer.Prepare()` a bit too early in `getQueryProcessingStage` in `storage_distributor.cpp`.
I am also concerned that I introduced a call to `Analyze` there, which might actually change the query AST (in my new way, but also in already existing ways). 
Once we are finished with some of the discussions in the comments, I plan to separate the inference of `ReadInOrderMode`, which will happen in `ParseQuery` or `Prepare`, and the modification of ORDER BY in the AST, which will happen in `Analyze`. 
I also propose renaming `Analyze` into something more verbose of the fact that it can modify the AST.

DimasKovas (approved):
To run tests

yt/chyt/server/query_analyzer.cpp
DimasKovas: I thought about it one more time and now I'm pretty sure that this type of change is illegal.

For example, let's look at this query:

```sql
select * from (
    select * from "//sorted_table"
    order by key
    limit 5
    union all
    select * from "//unsorted_table"
    order by key
    limit 5
) order by key
```

In this case ClickHouse may substitute the last `order by key` phase with `sorted merge`, because both subqueries have already been sorted by the key.

But if you change the order of nulls, two subqueries would be sorted differently, and sorted merge may either throw an error or return an incorrect result.

The same applies to NaNs - I'm not sure that it's a good idea to have this logic at all, even under an option.

achulkov2: The problem is that without this logic we might get incorrect results when read_in_order is enabled. What should we do?

yt/chyt/server/query_analyzer.cpp
```diff
@@ -728,6 +744,125 @@ void TQueryAnalyzer::OptimizeQueryProcessingStage()
...
+        for (const auto& orderByElementAst : selectQuery->orderBy()->children) {
+            auto orderByElement = orderByElementAst->as<DB::ASTOrderByElement>();
+
+            // The value of nulls_direction is equal to direction for NULLS LAST and to the opposite
+            // of direction for NULLS FIRST.
+            // For ASC (d = 1), we need NULLS FIRST (nd = -d = -1).
+            // For DSC (d = -1), we need NULLS LAST (nd =  d = -1).
+            if (orderByElement->nulls_direction != -1) {
+                THROW_ERROR_EXCEPTION_IF(
+                    orderByElement->nulls_direction_was_explicitly_specified,
```
DimasKovas, line 845: We have discussed that there should be no errors here - just disable the optimization if the order by clause is incompatible.

yt/chyt/server/query_analyzer.cpp
achulkov2: New approach: we do not change AST in any way, but add two additional options `assume_no_null_keys` and `assume_no_nan_keys`.
We check nulls direction for a key column if it could have either nulls or nans, but not both at the same time (only for floats), and only optimize if it is correct. If a float key column could have both nulls and nans we never optimize.

yt/chyt/server/query_analyzer.cpp
achulkov2: Removed this, see other comments.

yt/chyt/server/config.cpp
achulkov2: Since we want to disable the option by default for now, I will keep our separate option. I don't think it is easy (or that we should) disable a native CH option from our code. 

If we decide to enable it by default, we should probably deprecate it in favor of the native option (which is turned on by default).

DimasKovas: Ok

yt/chyt/server/query_analyzer.cpp
achulkov2: Changed it to a cast to `ASTIdentifier`. It is even better, because it works properly with aliases introduced by `WITH`!

achulkov2:
I have implemented what we have talked about, if there are no more issues, let's import the PR, wait for tests to pass internally, then change the default to `false` and import and merge.

DimasKovas (approved):
Tu run tests

yt/chyt/server/query_context-inl.h
```diff
@@ -0,0 +1,28 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+
+template <class T>
+void TQueryContext::SetRuntimeVariable(const TString& key, const T& value)
+{
+    auto guard = Guard(QueryLogLock_);
+    YT_VERIFY(RuntimeVariables_);
+    if (RuntimeVariables_->Contains(key)) {
+        YT_LOG_WARNING(
```
DimasKovas, line 17: Just in case:
TQueryContext is shared between all subqueries in the query.
E.g. if you have query like this:
```sql
SELECT * FROM t1
ORDER BY key
LIMIT 10
UNION ALL
SELECT * FROM t2
ORDER BY key
LIMIT 10
```
You will override the runtime variables when executing the second subquery and trigger this `YT_LOG_WARNING`

DimasKovas (approved):
LGTM

Please, run `ya test --style` in `yt/chyt/server/tests` - there are a log of `flake8` issues in your tests.

Unfortunately, CHYT tests weren't run internally because of an infrastructure issue. I've just restarted the check. If it passes, let's change default to `false` and merge it

yt/chyt/server/query_context-inl.h
achulkov2: After our discussion, I don't think it is worth investing into fixing this. Changed to YT_LOG_DEBUG, so that we could see it in the log.

yt/chyt/server/query_analyzer.cpp
```diff
@@ -728,6 +754,115 @@ void TQueryAnalyzer::OptimizeQueryProcessingStage()
...
+    }
+
+    auto commonDirection = EReadInOrderMode::None;
+
+    // Columns from the ORDER BY clause must form a prefix of the table's primary key.
+    if (std::ssize(selectQuery->orderBy()->children) > schema->GetKeyColumnCount()) {
+        return;
+    }
+
+    std::vector<TString> floatKeyColumnsWithoutExplicitNullsDirection;
```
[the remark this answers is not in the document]
achulkov2: Forgot to fix this in the last commit, done.

DimasKovas:
I don't see any relevant test failures. Let's change the default to `false` and merge

achulkov2:
> I don't see any relevant test failures. Let's change the default to `false` and merge

Done.

DimasKovas (approved):

PR #764, opened 2024-08-06
Make a few improvements to component version discovery

achulkov2:
Implements enhancements requested in #652:
* No more distinction between primary and secondary masters in favor of using the flat list of masters.
* Accounts for the possibility of incomplete results in master Get/List (and sets a large MaxSize).
* Uses native batch requests for fetching the contents of many orchids.
* Cleans up some compats.
* **Starts using MasterCache for version discovery**.

I decided to go with batch requests instead of direct RPC calls to orchids because this allows us to use master caches (now) and add some rate-limiting safeguards (later) by using a separate non-root user for discovery. Direct RPC calls much harder to rate-limit and do not have native caching, and I am more concerned about proxies (which there could be a lot of) making these kind of requests to *all* services. At least with masters we know our enemy in the face. Direct RPC calls are also harder to implement, since the current proxy bootstrap doesn't have any configured bus config, channel factory, etc.

### DIFF
--- a/yt/yt/server/http_proxy/component_discovery.cpp
+++ b/yt/yt/server/http_proxy/component_discovery.cpp
@@ -1,7 +1,13 @@
 #include "component_discovery.h"
 #include "coordinator.h"
 
+#include <yt/yt/ytlib/api/native/client.h>
+
+#include <yt/yt/ytlib/object_client/object_service_proxy.h>
+
 #include <yt/yt/client/api/client.h>
+
+#include <yt/yt/core/ytree/ypath_proxy.h>
 
 #include <util/string/split.h>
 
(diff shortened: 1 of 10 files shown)


yt/yt/server/http_proxy/component_discovery.cpp
```diff
@@ -149,15 +171,22 @@ std::vector<TClusterComponentInstance> TComponentDiscoverer::ListProxies(ECluste
...
     }
 
+    options.MaxSize = DefaultCypressMaxSize;
+
     auto nodeYson = WaitFor(Client_->GetNode(GetCypressDirectory(component), options))
         .ValueOrThrow();
-    auto addressToNode = ConvertTo<THashMap<TString, IMapNodePtr>>(nodeYson);
+    auto nodeMap = ConvertTo<IMapNodePtr>(nodeYson);
+
+    if (nodeMap->Attributes().Get("incomplete", false)) {
```
Kontakter, line 180: Let extract this check to separate method `ValueResultCompleteness` and avoid copypaste

PR #769, opened 2024-08-08
Support Decimal128/Decimal256 in Arrow parser

zlobober:


yt/yt/library/formats/arrow_parser.cpp
```diff
@@ -158,6 +163,31 @@ class TArraySimpleVisitor
         return ParseNull();
     }
 
+    // Decimal types. For now, YT natively supports only Decimal128 with scale up to 35.
```
savnadya, line 166: Making string serialization the default for decimals could lead to breaking changes in the future when a standard decimal256 type is introduced. To avoid potential compatibility issues, I recommend implementing string serialization only when a specific flag is enabled. We can achieve this by adding an attribute to the format (like `skip_null_values` in the yson format), and pass it only in `yt/yt/tools/import_table`.

How about naming it `allow_wide_decimal_as_string`?

savnadya (changes requested):

nadya002:
Shipped!

PR #771, opened 2024-08-08
Use multi-stage dockerfile for ya-build/ytsaurus ya package build

achulkov2:
Conditional base stages aren't very pretty, but it is not that bad and seems to work alright.


savnadya:
Shipped!

PR #774, opened 2024-08-12
[24.1] Use multi-stage dockerfile for ya-build/ytsaurus ya package build

achulkov2:
Conditional base stages aren't very pretty, but it is not that bad and seems to work alright.

---
c83a7ce69f3916663ac66ced840ad39e69f3cae0

Pull Request resolved: https://github.com/ytsaurus/ytsaurus/pull/771


savnadya:
Shipped!

PR #777, opened 2024-08-13
Add max_simultaneous_queries in YQL Agent

Krisha11:
Now we are limited to the total number of threads executing them (256) per yqla. If requests are received above this limit, the agents begin to suffer.
https://github.com/ytsaurus/ytsaurus/blob/main/yt/yql/agent/config.cpp#L333

To prevent this, we make a limit on the number of requests executed at the same time on each agent
Add a call to the yql agent to see if it is still ready to accept new queries. In query_tracker, the request does not start immediately, but may hang in the queue for some time until there are no free yqla.

In an ideal world, waiting situations should not happen. But in the case when yql agents are overloaded, it's better to stay in the queue a little bit than to go to a non-working machine


yt/yql/tests/test_simple.py
```diff
@@ -280,6 +281,58 @@ def test_pragma_refselect(self, query_tracker, yql_agent):
         """, [rows, [{"a": 42, "c": "test"}]])
 
 
+class TestAllYqlAgentsOverload(TestQueriesYqlBase):
+    YQL_AGENT_DYNAMIC_CONFIG = {"max_simultaneous_queries": 1}
```
Krock21, line 285: Just for clarity

```suggestion
    YQL_AGENT_DYNAMIC_CONFIG = {"max_simultaneous_queries": 1}
    NUM_YQL_AGENTS = 1
```

yt/yt/server/query_tracker/query_tracker.cpp
```diff
@@ -318,14 +318,9 @@ class TQueryTracker
...
             optionalRecord->State);
 
-        // If current query state is "pending", switch it to "running". Otherwise, keep the existing state of a query;
-        // in particular, it may be "failing" or "completing" if the previous incarnation succeeded in reaching pre-terminating state.
-        auto newState = optionalRecord->State == EQueryState::Pending ? EQueryState::Running : optionalRecord->State;
-
         auto rowBuffer = New<TRowBuffer>();
         TActiveQueryPartial newRecord{
             .Key = queryRecord.Key,
-            .State = newState,
```
Krock21, line 328: Why do we remove it? Shouldn't we keep the same state?

`.State = optionalRecord->State`

yt/yt/server/query_tracker/yql_engine.cpp
```diff
@@ -141,11 +138,84 @@ class TYqlQueryHandler
...
+            }
+
+            YT_LOG_DEBUG("Start YQL query (Stage: %v, Channel: %v)",
+                Stage_,
+                yqlServiceChannel->GetEndpointDescription());
+
+            QueryState_ = EYqlQueryState::Running;
+
+            AsyncQueryResult_ = startQueryReq->Invoke();
+            AsyncQueryResult_.Subscribe(BIND(&TYqlQueryHandler::OnYqlResponse, MakeWeak(this)).Via(GetCurrentInvoker()));
```
Krock21, line 208: So we don't actually limit amount of running YQL queries to `MaxSimultaneousQueries`

There could be a delay between `rsp->ready_to_get_queries()` and increasing `ActiveQueries_` in yqlAgent in `startQueryReq->Invoke()` (For example, if QT -> YQLAgents network stops for a moment). Then many queries will think that yqlAgent is ready and start themselves, overloading yqlAgents

To prevent it I suggest introducing `Throttled` error in `YqlResponse` and handling it in QueryTracker. Probably a new `EYqlQueryState::SentToYqlAgent` state will be needed for it(or making sync call to yqlAgents)

For a first version, however, current approach is fine

yt/yt/server/query_tracker/query_tracker.cpp
Krisha11: Here, we only update the specified fields, so "State" will remain the same if it is not specified.

yt/yt/server/query_tracker/yql_engine.cpp
Krisha11: Thank you for the suggestion to return Throttling errors from the YQL agent.

However, I'm not sure that it is necessary to introduce a new "SentToYqlAgent" state or make a sync call to implement this. New states requires changes in proxies and sdk - I would like to avoid this.

Instead of introducing a new state, we could simply use a combination of the "Running" and "Pending" states. When we attempt to run a query, we would move it to the "Running" state, and if there is a throttling failure, we would revert it to "Pending".

Krock21: >Instead of introducing a new state, we could simply use a combination of the "Running" and "Pending" states. When we attempt to run a query, we would move it to the "Running" state, and if there is a throttling failure, we would revert it to "Pending".

👍 . I see you introduced "Throttled" state too. Why?

Krisha11: I didn't introduce it. What do you mean?

yt/yql/agent/yql_agent.cpp
```diff
@@ -183,11 +212,18 @@ class TYqlAgent
...
     TFuture<std::pair<TRspStartQuery, std::vector<TSharedRef>>> StartQuery(TQueryId queryId, const TString& user, const TReqStartQuery& request) override
     {
+        if (ActiveQueriesGuard::Get() >= DynamicConfig_->MaxSimultaneousQueries) {
+            YT_LOG_INFO("Query was throttled (QueryId: %v, User: %v)", queryId, user);
+            THROW_ERROR_EXCEPTION(NYqlClient::EErrorCode::Throttled, "Query was throttled");
+        }
+
         YT_LOG_INFO("Starting query (QueryId: %v, User: %v)", queryId, user);
 
         return BIND(&TYqlAgent::DoStartQuery, MakeStrong(this), queryId, user, request)
```
Krock21, line 229: Next line is `.AsyncVia(ThreadPool_->GetInvoker())`

As long as `MaxSimultaneousQueries`(128) is lower than `YqlThreadCount`(256), the delay between checking `ActiveQueriesGuard::Get() >= DynamicConfig_->MaxSimultaneousQueries` and increasing `ActiveQueries` should be minimal (`ThreadPool_` queue should be empty)

In the opposite case this throttling will stop working as we will never reach values above `YqlThreadCount`. So we may want to add this check (`YqlThreadCount >= MaxSimultaneousQueries`) somewhere

But even with this small delay between checking ActiveQueries (line 222) and increasing it (line 287) we may see a situation with a long queue. This can be illustrated with this unlikely example:
1. 1000 queries is simultaneously dispatched to yql agent and start executing `StartQuery`
2. All of them execute line 222 and continue forward to line 229
3. 256 of them start executing in the `ThreadPool_`, the rest are in ThreadPool's queue
4. 256 queries increase ActiveQueries counter, finally preventing new queries to be added to `ThreadPool_`

This is usually resolved by using conditional increase, as shown in [this stackoverflow question](https://stackoverflow.com/questions/13778110/is-there-atomic-increment-with-the-check-preconditions-that-the-atomic-value-wa)

So I suggest using it to atomically check that `ActiveQueries < MaxSimultaneousQueries` and increase `ActiveQueries` if this is true:
1. We add a check that `YqlThreadCount` is strictly more than `MaxSimultaneousQueries`, to make sure that we have at least 1 free thread in `ThreadPool_`
2. In DoStartQuery in guard constructor we do:
```
IsTaken_ = true;
auto activeQueries = ActiveQueries_.load();
do
{
  if (activeQueries >= MaxSimultaneousQueries) {
    IsTaken_ = false;
    return;
  }
} while (!ActiveQueries_.compare_exchange_weak(activeQueries, activeQueries + 1));
```
3. In DoStartQuery after guard creation we do `if (!guard.IsTaken_) THROW_THROTTLED_ERROR`
4. In guard destructor we do nothing if `!IsTaken_`

This can still lead to queue in `ThreadPool_`. But it should be quickly emptied by at least 1 thread that will be taking requrests from there and throwing throttled exceptions.

One solution to avoid it is to conditionally increase ActiveQueries in `StartQuery` and decrease it after `DoStartQuery` finish. But it will add more complexity and will require passing a guard from `StartQuery` to `DoStartQuery`, so I suggest sticking to a solution described above as a balanced one

yt/yql/agent/bin/ytserver-yql-agent
```diff
@@ -0,0 +1 @@
+/home/mpereskokova/.ya/build/symres/79be5630d0861cb493abd77919bcb424/ytserver-yql-agent
```
Krock21, line 1: Remove the file (seems like a symlink)

yt/yql/tests/yt-yql-tests
```diff
@@ -0,0 +1 @@
+/home/mpereskokova/.ya/build/symres/1e37945c6e43193002c56a6203671fa5/yt-yql-tests
```
Krock21, line 1: Remove the file (seems like a symlink)

Krock21 (commented):
LGTM after removing symlinks

Krisha11:
Shipped!

PR #778, opened 2024-08-13
Some grafana-related improvements for dashboard generator

achulkov2:
This PR contains the following changes:
* New user-provided/user-defined postprocessors are introduced, along with few general-purposes custom postprocessor useful for grafana installations + usage examples.
* Grafana backend tries to use exact comparison operators where possible.
* Postprocessors are refactored to use dicts.
* When a backend is explicitly specified via CLI, all other configured backends are ignored. 
* Some grafana project service names are renamed to follow a common pattern.

TODO in future PRs:
* Generalize dashboard parameters with suggested values across both backends. It should be implemented via `label_values(query, label_name)` in Grafana. 
* Gradually work through dashboards and make sure they actually work in Grafana.
* Support dashboard creation (vs existing dashboard) in specific folders for Grafana.

### DIFF
--- a/yt/admin/dashboard_generator/yt_dashboard_generator/backends/grafana/__init__.py
+++ b/yt/admin/dashboard_generator/yt_dashboard_generator/backends/grafana/__init__.py
@@ -9,6 +9,11 @@ from ...postprocessors import DollarTemplateTagPostprocessor
 import requests
 import enum
 import lark
+import logging
+
+##################################################################
+
+logger = logging.getLogger()
 
 ##################################################################
 
@@ -67,20 +72,28 @@ class GrafanaProxy():
 
     def submit_dashboard(self, serialized_dashboard, dashboard_id):
         try:
-            current = self.fetch_dashboard(dashboard_id)["dashboard"]
-            print("Current version:", current["version"])
-        except BaseException:
+            current = self.fetch_dashboard(dashboard_id)
+            current_dashboard = current["dashboard"]
+            logger.info("Current version:", current_dashboard["version"])
+        except BaseException as ex:
+            logger.exception("Failed to retrieve current dashboard configuration, using plain configuration")
             current = {}
+            current_dashboard = {}
 
         request = serialized_dashboard
         if dashboard_id is not None:
             request["uid"] = dashboard_id
-        request.setdefault("title", current.get("title", "some default title"))
-        request.setdefault("title", "arusntarstrst")
-        request.setdefault("templating", current.get("templating", {}))
-        request["version"] = current.get("version", 1)
+        request.setdefault("title", current_dashboard.get("title", "some default title"))
+        request.setdefault("templating", current_dashboard.get("templating", {}))
+        request["version"] = current_dashboard.get("version", 1)
 
         request = {"dashboard": request}
+
+        folder_id = current.get("meta", {}).get("folderUid")
+        if folder_id is not None:
+            request["folderUid"] = folder_id
+        elif not current:
+            logger.warning("Submitting Grafana dashboard without specified folder and knowledge about current dashboard, this might cause the dashboard to move to your root folder")
 
         rsp = requests.post(
             f"{self.base_url}/api/dashboards/db",
@@ -146,6 +159,16 @@ class GrafanaSerializerBase(SerializerBase):
             regex.append(c)
         return "".join(regex)
 
+    # This is a bit hacky, but should work for most reasonable cases of non-regex values and variables.
+    def _is_exact(self, value):
+        return all(c.isalnum() or c in "-_<>" for c in value.removeprefix("$"))
+
+    def _get_comparison_operator(self, negate, exact):
+        if exact:
+            return "!=" if negate else "="
+        else:
+            return "!~" if negate else "=~"
+
     def _prepare_bin_op_expr_query(self, expression):
         assert expression.args[0] in "+-*/", f"Binary operation {expression.args[0]} is not supported. Expression: {expression}, expression type: {type(expression)}"
         left_query, left_tags = self._prepare_expr_query(expression.args[1])
@@ -194,7 +217,7 @@ class GrafanaSerializerBase(SerializerBase):
         if not issubclass(type(sensor), Sensor):
             raise Exception(f"Cannot serialize cell content of type {type(sensor)}")
 
-        tags = list(sensor.get_tags().items())
+        tags = sensor.get_tags()
         sensor_name = sensor.sensor
 
         if self.tag_postprocessor is not None:
@@ -204,7 +227,7 @@ class GrafanaSerializerBase(SerializerBase):
 
         string_tags = []
         other_tags = {}
-        for k, v in tags:
+        for k, v in tags.items():
             if k == sensor.sensor_tag_name:
                 continue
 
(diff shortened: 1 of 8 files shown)


ifsmirnov:
I'll take a look this week.

yt/admin/dashboard_generator/yt_dashboard_generator/backends/grafana/__init__.py
```diff
@@ -146,6 +159,16 @@ def _glob_to_regex(self, glob):
             regex.append(c)
         return "".join(regex)
 
+    # This is a bit hacky, but should work for most reasonable cases of non-regex values and variables.
+    def _is_exact(self, value):
+        return all(c.isalnum() or c in "-_<>" for c in value.removeprefix("$"))
```
Kontakter, line 164: Is it possible to make an explicit annotation on values? And use this heuristic in case of missing annotation

yt/admin/dashboards/yt_dashboards/common/sensors.py
```diff
@@ -13,16 +13,20 @@ class YtSystemTags(Enum):
 
 yt_host = YtSystemTags.HostContainer
 
+MonitoringServiceTag = MonitoringTag("service")
+# TODO(achulkov): This is actually promised to be "job" in README. Should we change it?
```
Kontakter, line 17: What do you mean in this comment?

yt/admin/dashboard_generator/yt_dashboard_generator/cli.py
```diff
@@ -139,17 +139,21 @@ def run(self, args):
                 headers="firstrow"))
             return
 
+        def is_selected_backend(backend):
```
Kontakter, line 142: More natural name `is_backend_selected`

Kontakter:
To be sure – do you check that dashboard tests successfully pass?

yt/admin/dashboards/yt_dashboards/common/sensors.py
achulkov2: [Here](https://github.com/ytsaurus/ytsaurus/tree/main/yt/admin/dashboards#yt-helpers) under `ProjectSensor` we say that it is `service=` for Monitoring and `job=` for Grafana.

achulkov2:
> To be sure – do you check that dashboard tests successfully pass?

image

As it turns out there are tests, but they don't seem to be opensourced: https://github.com/ytsaurus/ytsaurus/blob/0565ce0c78062c77d98141e08985009e53d7dd56/yt/admin/dashboards/ya.make#L47-L50
I did check that it works for our Grafana, and I hope that Ivan will double check everything on the monitoring side. If we can opensource the tests that would be great, but I do not know whether they rely on any private APIs/mocks.

Kontakter (approved):
Approve for testing in internal CI only

Kontakter:
```
yt/admin/dashboards/yt_dashboards/common/postprocessors.py:11: [F401] 'typing.Iterable' imported but unused
```

And tests should be recanonized of course

Kontakter:
@ifsmirnov – kindly remind

yt/admin/dashboards/yt_dashboards/bin/main.py
```diff
@@ -54,10 +65,21 @@
...
         },
     },
+    # Supported backends: monitoring, grafana.
     "scheduler-pool": {
         "func": build_scheduler_pool,
         "monitoring": {},
-        "grafana": {},
+        "grafana": {
+            "id": "<your-dashboard-id>",
+            "postprocessor": combine_postprocessors(
```
ifsmirnov, line 74: It may be reasonable to move these postprocessors out of the list and make a generic function since there is almost no specifics to scheduler. This generic function may take value_mapping or so.

yt/admin/dashboard_generator/yt_dashboard_generator/backends/grafana/__init__.py
ifsmirnov: I agree here. Maybe we can introduce `ExactMatch("__workspace__")` like it is done with Regex now.

yt/admin/dashboards/yt_dashboards/bin/main.py
```diff
@@ -54,10 +65,21 @@
             "args": [False],
         },
     },
+    # Supported backends: monitoring, grafana.
```
ifsmirnov, line 68: Current convention is that Grafana backend is supported iff there is `"grafana"` section in the map, so the comment looks unnecessary.

yt/admin/dashboards/yt_dashboards/common/sensors.py
ifsmirnov: Our testing installation had `job=`, that's why I included it into the readme. I encourage you changing readme as well if you change the name here.

ifsmirnov (approved):

yt/admin/dashboards/yt_dashboards/bin/main.py
achulkov2: I assumed that this list in `main.py` is an example to follow, rather than the final configuration one should use for their own installation. That is why I only provided one example with dummy parameters and extra comments, since the configuration would obviously differ for each particular use (and the actual dashboards-as-code should be in a separate file).

In an actual configuration this whole part will be about the same length as a function call with `extra_tags` and `value_mapping` as arguments, IMO, so it is fine to repeat it (probably storing these tags/mappings in a global variable in the beginning of the file). 
For this reason I don't see much use in a function.

But maybe you have an example/rationale that will change my mind.

yt/admin/dashboards/yt_dashboards/bin/main.py
achulkov2: Reasonable.

yt/admin/dashboard_generator/yt_dashboard_generator/backends/grafana/__init__.py
achulkov2: I will try to introduce something like `ExactMatch`, but I am quite convinced that `~=` should not be the default, since it cannot be faster than exact match (and it might not work for some labels). So I will use this heuristic in case exact/regex is not specified.

yt/admin/dashboards/yt_dashboards/common/sensors.py
achulkov2: Frankly, this differs too much depending on your specific Grafana installation + collection specifics. I guess I will change to `service` in both places and add some more details in the readme on what this label should represent.

yt/admin/dashboard_generator/yt_dashboard_generator/backends/grafana/__init__.py
ifsmirnov: `~=` was the default since internal dashboards heavily exploited globs and we used them in labels frequently. Perhaps Grafana has some specific labels that should be matched exactly. I can't tell due to lack of experience.

ifsmirnov:
This PR is pending on the author's side.

Kontakter (changes requested):
Set "Request changes" to mark this PR properly

savnadya:
test comment

PR #781, opened 2024-08-14
[queues] Take shared lock for static exports

savnadya:


yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -169,9 +169,10 @@ class TQueueExportTask
 
     static constexpr TStringBuf ExportProgressAttributeName_ = "queue_static_export_progress";
     static constexpr TStringBuf ExportDestinationAttributeName_ = "queue_static_export_destination";
+    static constexpr TStringBuf ExporterAttributeName_ = "queue_static_exporter";
 
     //! Performs the following steps:
-    //!   1) Starts transaction, obtains locks on input queue (snapshot) and export directory (exclusive).
+    //!   1) Starts transaction, obtains locks on input queue (snapshot), and <export_directory>/@queue_static_exporter attribute (shared).
```
achulkov2, line 175: Since you are fixing commas in comments, I don't think one is actually needed :) 
Starts transaction, obtains locks on X and Y.

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -330,7 +339,7 @@ class TQueueExportTask
         if (QueueObject_.Type != EObjectType::Table) {
             THROW_ERROR_EXCEPTION(
                 "Invalid type of %v: expected %Qlv, %Qlv found",
-                QueueObject_.GetPath(),
+                Queue_,
```
achulkov2, line 342: (idk how to add comments to unchanged lines)
Maybe `YT_VERIFY` that object id returned from attributes is equal to our saved object id in path?
Probably easier the other way around: `FromObjectId(objId) = GetPath()`.

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -194,9 +195,17 @@ class TQueueExportTask
...
         auto transactionId = transaction->GetId();
-        WaitFor(transaction->LockNode(Queue_, ELockMode::Snapshot))
-            .ThrowOnError();
-        WaitFor(transaction->LockNode(ExportConfig_.ExportDirectory, ELockMode::Exclusive))
+        auto queueObjectId = WaitFor(transaction->LockNode(Queue_, ELockMode::Snapshot))
+            .ValueOrThrow()
+            .NodeId;
+
+        // Take it for guarantee that only one queue agent instance do this export.
+        WaitFor(transaction->LockNode(
```
achulkov2, line 203: Let's add a test in which we take a shared lock with attribute_key on the export directory (to mimic what Logfeller does) and check that exports do not fail.

achulkov2 (approved):
Otherwise LGTM.

PR #792, opened 2024-08-20
Remove unused TMultiReaderManagerBase.CreatingReader_

koct9i:
No effect - never set to true.

### DIFF
--- a/yt/yt/ytlib/chunk_client/multi_reader_manager_base.cpp
+++ b/yt/yt/ytlib/chunk_client/multi_reader_manager_base.cpp
@@ -133,10 +133,6 @@ void TMultiReaderManagerBase::OpenNextReaders()
         return;
     }
 
-    if (CreatingReader_) {
-        return;
-    }
-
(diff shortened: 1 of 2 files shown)


yt/yt/ytlib/chunk_client/multi_reader_manager_base.cpp
```diff
@@ -191,11 +187,6 @@ void TMultiReaderManagerBase::OnReaderCreated(
         return;
     }
 
-    {
-        auto guard = Guard(PrefetchLock_);
-        CreatingReader_ = false;
```
Kontakter, line 196: It seems that correct solution is to set here `true`

maxim-babenko: I agree.

Instead of dropping this (clearly invalid) code I suggest we try to fix it.

Kontakter:
We discussed it with @maxim-babenko: he will fix the behaviour of this flag, and I suppose just to drop this PR @koct9i

PR #793, opened 2024-08-21
Create query tracker image build script

Krock21:
Query tracker image can only be built in CI right now.

This PR creates a script that can be used to build query tracker image from source and extracts common logic of building yql related parts into a separate script.

It also extracts common part between query-tracker and ytsaurus images into a separate `base` ya package that is included in both

Tested building locally. Tested created image on a local cluster


Krisha11:
Maybe we can test it once in CI? Just to ensure that we have enough time\memory guarantees for success run

Krock21:
> Maybe we can test it once in CI? Just to ensure that we have enough time\memory guarantees for success run

Yes I plan to do it today

yt/docker/ya-build/query-tracker/build_yql_agent.sh
```diff
@@ -0,0 +1,76 @@
...
+                stat \
+                streaming \
+                string \
+                top \
+                topfreq \
+                unicode_base \
+                url_base \
+                yson2
+do
+    ${YDB_SOURCE_PATH}/ya make -T ${BUILD_FLAGS} --ignore-recurses -DSTRIP=yes --output=${YQL_BUILD_PATH} ${YDB_SOURCE_PATH}/ydb/library/yql/udfs/common/${udf_name}
```
Krisha11, line 59: Why are we starting to ignore STRIP flag here?

yt/docker/ya-build/query-tracker/build_query_tracker.sh
```diff
@@ -0,0 +1,53 @@
...
+strip --remove-section=.gnu_debuglink ${YTSAURUS_BUILD_PATH}/yt/yt/python/yson_shared/*.so ${YTSAURUS_BUILD_PATH}/yt/yt/python/driver/rpc_shared/*.so ${YTSAURUS_BUILD_PATH}/yt/yt/python/driver/native_shared/*.so
+
+mkdir -p ${OUTPUT_PATH}/artifacts
+
+cp ${YTSAURUS_BUILD_PATH}/yt/yt/python/yson_shared/yson_lib.so ${OUTPUT_PATH}/artifacts/libyson_lib.so
+cp ${YTSAURUS_BUILD_PATH}/yt/yt/python/driver/native_shared/driver_lib.so ${OUTPUT_PATH}/artifacts/libdriver_lib.so
+cp ${YTSAURUS_BUILD_PATH}/yt/yt/python/driver/rpc_shared/driver_rpc_lib.so ${OUTPUT_PATH}/artifacts/libdriver_rpc_lib.so
+
+# Build ytserver-all
+${YTSAURUS_SOURCE_PATH}/ya make -T ${BUILD_FLAGS} --output=${YTSAURUS_BUILD_PATH} ${YTSAURUS_SOURCE_PATH}/yt/yt/server/all
```
Krisha11, line 50: Does this mean that from this moment we will start building yt/yt/server/all twice?

Krock21:
CI run: https://github.com/ytsaurus/ytsaurus/actions/runs/10557484230

yt/docker/ya-build/query-tracker/build_yql_agent.sh
Krock21: I looked at python libraries and saw that they are always stripped:
https://github.com/ytsaurus/ytsaurus/blob/b47834934522a0d9d07f98c087fa8aaf3b56e63b/.github/workflows/build-jobs.yaml#L232

I decided that it makes sense to strip these .so's as well. Isn't this the case?

yt/docker/ya-build/query-tracker/build_query_tracker.sh
Krock21: No. This script is not used in CI. CI uses `build_yql_agent.sh` script and does `build_query_tracker.sh` parts separately (building python packages and ytserver-all)

achulkov2 (changes requested):
The way I see all of this, there are two kinds of artefacts: ones used for query tracker/yql agent only (udf and ydb related stuff), and ones reused by other components (python wheels and ytserver-all). 

In the current CI, artefacts of the second kind are built separately and reused across jobs. Meanwhile, in ya-build packages we use ya package + additional stages in the Dockerfile to build them. 
As there is no plan to change this part of the CI (for now), let's use the same method for building ytserver-all and python wheels as in other ya-build packages/scripts.

As for the first kind of artefacts, building them should be indeed incapsulated into some bash script, which will be reused between build-jobs.yaml and whatever configuration the separate build will use (for example, we could make this script a ya-acceptable target). 

Extra idea worth considering: since we are building stuff within a docker container, why not make the build process part of the main Dockerfile itself? Maybe it would be possible to avoid bash-scripts completely, idk.

Krock21:
Rebased, new CI: https://github.com/ytsaurus/ytsaurus/actions/runs/10816834469

yt/docker/ya-build/query-tracker/build_query_tracker.sh
```diff
@@ -0,0 +1,53 @@
...
+
+# Copy required files from ytsaurus source
+mkdir -p ${OUTPUT_PATH}/yt/python
+cp -a ${YTSAURUS_SOURCE_PATH}/yt/python/. ${OUTPUT_PATH}/yt/python
+mkdir -p ${OUTPUT_PATH}/yt/yt/python
+cp -a ${YTSAURUS_SOURCE_PATH}/yt/yt/python/. ${OUTPUT_PATH}/yt/yt/python
+mkdir -p ${OUTPUT_PATH}/certs
+cp ${YTSAURUS_SOURCE_PATH}/certs/cacert.pem ${OUTPUT_PATH}/certs/cacert.pem
+
+# Build python libraries
```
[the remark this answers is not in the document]
Krock21: As we discussed, it is not present in our multi-stage dockerfile and we can't avoid it. For other packages it is hidden in package.json

yt/docker/ya-build/query-tracker/build_query_tracker.sh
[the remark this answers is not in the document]
Krock21: Done

Krock21:
We discussed with @achulkov2 the best approach and decided the following:

- We want ya package (Because other images in ya-build folder utilize it)
  - We can't use it alone, since it doesn't support multiple repositories (we build artifacts from ydb)
- We want to reuse common parts between ytsaurus and query-tracker. They are extracted to `base` package

This PR:
1. Extracts common part to `base` folder. It consists of ytserver-all, certificates, credits, Dockerfile, python-related stuff
2. Creates query-tracker package.json that copies ydb artifacts from `yql_build` directory and produces a docker image
3. Creates `build_yql_agent.sh` script that builds all required artifacts for yql agent
4. Adjusts CI to use this script

Ci run 1: https://github.com/ytsaurus/ytsaurus/actions/runs/10816834469

yt/docker/ya-build/query-tracker/build_system_python_udfs.sh
```diff
@@ -0,0 +1,32 @@
...
+               3.10 \
+               3.11 \
+               3.12
+do
+  udf_suffix=$(echo "$version" | tr . _)
+  udf_name="python${udf_suffix}"
+  python_name="python${version}"
+
+  python3 $YDB_SOURCE_PATH/ya make $YDB_SOURCE_PATH/ydb/library/yql/udfs/common/python/system_python/${udf_name} \
+    -T ${BUILD_FLAGS} --ignore-recurses -DSTRIP=yes -DOS_SDK=local -DUSE_ARCADIA_PYTHON=no -DUSE_LOCAL_PYTHON -DPYTHON_CONFIG=${python_name}-config -DPYTHON_BIN=${python_name} --output $YQL_BUILD_PATH
```
Krock21, line 30: `-DOS_SDK=local` here allows to avoid some bug in ya, when `/usr/include/pythonX/pyconfig.h` tries to include `ARCH/pythonX/pyconfig.h` and fails. It can be fixed some other way (by creating some symlinks and tricking it), but `-DOS_SDK=local` just works (cause we are in pure ubuntu focal docker with only pythons installed)

achulkov2: I do not like this approach. OS_SDK local is intended to enable other functionality, and considering how little these options are documented it is unwise to use them "just because it works". For example, I am not sure that running this script in the same ubuntu release as the main docker image is what we need, maybe we would need to run it using the same sdk the ya make toolchain uses internally. 

Let's use one of the hacks without OS_SDK=local, and create a ticket for yatool about this bug, if we haven't already.

achulkov2 (changes requested):

Krock21:
Latest CI run is successful and the image is working(tested on internal cluster)

yt/docker/ya-build/query-tracker/package.json
```diff
@@ -0,0 +1,84 @@
...
+    },
+    "params": {
+        "format": "docker",
+        "docker_target": "query-tracker",
+        "docker_build_arg": {
+            "PYTHON_BUILD_BASE": "build-ytsaurus-python-binaries",
+        },
+    },
+    "include": [
+      "yt/docker/ya-build/base/package.json",
```
achulkov2, line 16: I remember we discussed that this ends up breaking the param definitions above. Idk what you figured out here, but anything changed in how `ya package` has to be called (for example, if one is always required to specify docker_target now), please add it to the corresponding READMEs. 

Also, let's create issues in yatool, if something is broken.

achulkov2 (approved):
Generally, LGTM.

yt/docker/ya-build/query-tracker/package.json
Krock21: https://github.com/yandex/yatool/issues/18

Krock21:
Sync were losing executable flag from scripts before
Could you please check that .sh files will preserve executable flag during import?

Krisha11:
> Sync were losing executable flag from scripts before Could you please check that .sh files will preserve executable flag during import?

They were lost. I added as it should be

PR #794, opened 2024-08-22
Fix flaky dynamic table logger test (hopefully)

achulkov2:
We write too many logs and local YT cannot handle it. I left only a few categories for the test to work better. Ran them in a loop for a few hours without failures, so hopefully this will help. 

P.S. We include Concurrency deliberately because we filter it out using rate limiters.


maxim-babenko:
Shipped!

PR #800, opened 2024-08-24
Fix read_query_result always returning raw results.

zlobober:
Currently, read_query_result always acts as if raw = True, since
the implementation does not care to load rows from the specified
format. I fix that similar to how read_table is implemented.

This is an obvious mistake that would not happen if we had tests
on Python SDK for query methods, but we do not have them, and
obtaining them is considerably hard (due to QT not being supported
by yt_local).

I tested the fixed implementation locally, it seems working correctly.


Kontakter (approved):
To run tests in internal CI

PR #802, opened 2024-08-24
Add query tracker to yt_local.

zlobober:
After this change, it becomes possible to specify query-tracker-count
in yt_local.

This change allows proper testing of query-related methods of Python SDK
in the subsequent commits.


Kontakter:
@savnadya – please take a look, I think these changes conflict with your plan of adding query tracker to yt_local

zlobober:
Just in case, this is only the support for query tracker, not for YQL agent. Supporting query tracker like this feels natural enough, as it is already bundled as a program in ytserver-all.

zlobober:
Decided to discard this PR in favor of https://github.com/ytsaurus/ytsaurus/commit/c5c56e195df25201749d2bfc2350877ac44877b8.

PR #803, opened 2024-08-26
Move query tracker docs to a separate section

Krock21:
This is needed for growing documentation about query tracker. DQ and Python UDF articles will be added to this section

Checked locally. All as expected


ogorbacheva (approved):
OK

Krisha11 (approved):

PR #811, opened 2024-08-30
yt-local: look at file name extension for http-proxy config

koct9i:
Historically it was in json format.


Kontakter:
I have already remove this logic with json format. So I'll drop this PR

PR #814, opened 2024-08-30
Add query_tracker custom solomon metrics

Krisha11:


yt/yt/server/query_tracker/handler_base.cpp
```diff
@@ -376,6 +384,31 @@ bool TQueryHandlerBase::TryWriteQueryState(EQueryState state, EQueryState previo
         }
         WaitFor(transaction->Commit())
             .ThrowOnError();
+
+        auto now = TInstant::Now();
+        StateTimes[previousState] += now - LastStateChange_.value_or(record.StartTime);
```
Krock21, line 389: This logic may sometimes give us incorrect information. For example, when a query started, changed state, and then got assigned to a different query tracker. I am not sure if this situation is possible now, but seems like it should be.

Instead of `record.StartTime` it would be more correct to use time, when a query tracker acquired the query

Krock21 (approved):
Tests for it would be nice

yt/yt/server/query_tracker/handler_base.cpp
```diff
@@ -376,6 +384,31 @@ bool TQueryHandlerBase::TryWriteQueryState(EQueryState state, EQueryState previo
...
+        LastStateChange_ = now;
+
+        if (state == EQueryState::Completing || state == EQueryState::Failing) {
```
psushin, line 392: why don't we emit sensor values as soon as they are obtained, and instead firts collect all the  StateTimes in the separate variable until query completion?

psushin:
I also think that a few tests would be helpful.

psushin (changes requested):

yt/yt/server/query_tracker/handler_base.cpp
Krisha11: For most counters, the situation is as you describe. However, for example, in the case of a pending state, a query may enter into this state multiple times (if yqla is overloaded)

yt/yt/server/query_tracker/query_tracker.cpp
```diff
@@ -646,6 +670,24 @@ class TQueryTracker
...
+                .AssignedTracker = activeQueryRecord->AssignedTracker.value_or(NoneQueryTracker),
+            };
+            {
+                auto guard = std::make_optional(Guard(StateTimeProfilingCountersMap_->Lock));
+                auto& stateTimeGauge = GetOrCreateProfilingCounter(
+                    QueryTrackerProfiler,
+                    tags,
+                    StateTimeProfilingCountersMap_->Map,
+                    guard).StateTime;
+                auto now = TInstant::Now();
```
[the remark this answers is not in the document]
Krisha11: If you are talking about this loop, it is useful to recalculate it every iteration because they are too long.
https://github.com/ytsaurus/ytsaurus/blob/b2952c4fd20b002525efd1c862e6f6b264fb954a/yt/yt/server/query_tracker/query_tracker.cpp#L532

yt/yt/server/query_tracker/handler_base.cpp
```diff
@@ -342,7 +347,11 @@ void TQueryHandlerBase::TryWriteProgress()
 bool TQueryHandlerBase::TryWriteQueryState(EQueryState state, EQueryState previousState, const TError& error, const std::vector<TErrorOr<TWireRowset>>& wireRowsetOrErrors)
 {
     try {
-        auto transaction = StartIncarnationTransaction(previousState);
+        ITransactionPtr transaction;
+        TActiveQuery record;
+        std::tie(transaction, record) = StartIncarnationTransaction(previousState);
+
+        auto now = TInstant::Now();
```
Krock21, line 354: Would be nice to tie all these timestams to actual transaction timestamps that change state

Krock21 (approved):

yt/yt/server/query_tracker/handler_base.cpp
Krisha11: Why do we want this?

Now it is consistent with other times (like start_time, finish_time)
https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/server/query_tracker/query_tracker_proxy.cpp#L682

And also transaction timestamps are less precise. (seconds instead of microseconds)
https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/client/transaction_client/public.h#L67

yt/yql/tests/test_simple.py
```diff
@@ -42,6 +43,47 @@ def _test_simple_query(self, query, expected, **kwargs):
...
+
+        create_pool("small", attributes={"resource_limits": {"user_slots": 0}})
+        query = start_query("yql", 'pragma yt.StaticPool = "small"; select a+1 as result from primary.`//tmp/t`')
+
+        wait(lambda: self._exists_pending_stage_in_progress(query))
+
+        def filter_sensor(sensor):
+            return sensor["labels"]["sensor"] == "yt.yql_agent.active_queries"
+
+        def lastActiveQueriesMetric():
```
psushin, line 72: python styleguide violation ;)

yt/yql/tests/test_simple.py
```diff
@@ -42,6 +43,47 @@ def _test_simple_query(self, query, expected, **kwargs):
...
+            sensors = requests.get(f"http://localhost:{yql_agent.yql_agent.configs[0]["monitoring_port"]}/solomon/shard/default").json()["sensors"]
```
psushin, line 73: you can take a look at profiler_factory() in the other integration tests for boilerplate elimination

yt/python/yt/environment/init_query_tracker_state.py
```diff
@@ -677,6 +677,55 @@ def finished_queries_table_by_aco_and_start_time_mapper(row):
...
+    ],
+    [
+        ("engine", "string", {"lock": "client"}),
+        ("query", "string", {"lock": "client"}),
+        ("files", "any", {"lock": "client"}),
+        ("settings", "any", {"lock": "client"}),
+        ("user", "string", {"lock": "client"}),
+        ("access_control_objects", "any", {"lock": "client"}),
+        ("start_time", "timestamp", {"lock": "client"}),
+        ("start_running_time", "timestamp", {"lock": "client"}),
```
psushin, line 692: `run_start_time` or `execution_start_time` can be better options

yt/yt/server/query_tracker/handler_base.h
```diff
@@ -68,6 +69,9 @@ class TQueryHandlerBase
     int ProgressVersion_ = 0;
     int LastSavedProgressVersion_ = 0;
 
+    std::optional<TInstant> LastStateChange_;
+    THashMap<EQueryState, TDuration> StateTimes;
```
psushin, line 73: Missing trailing _

yt/yt/server/query_tracker/profiler.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include <util/digest/multi.h>
+
+namespace NYT::NQueryTracker {
+
+using namespace NProfiling;
+
+////////////////////////////////////////////////////////////////////////////////
+
+TActiveQueriesProfilingCounter::TActiveQueriesProfilingCounter(const NProfiling::TProfiler& profiler)
```
psushin, line 12: excessive namespace qualifier

yt/yt/server/query_tracker/profiler.cpp
```diff
@@ -0,0 +1,31 @@
...
+    : ActiveQueries(profiler.Gauge("/active_queries"))
+{ }
+
+TStateTimeProfilingCounter::TStateTimeProfilingCounter(const NProfiling::TProfiler& profiler)
```
psushin, line 16: excessive namespace qualifier

yt/yt/server/query_tracker/profiler.h
```diff
@@ -0,0 +1,70 @@
+#pragma once
+
+#include <yt/yt/library/profiling/sensor.h>
```
psushin, line 3: yt/yt/client should be the first include, since yt/yt/library is more general than yt/yt/client

yt/yt/server/query_tracker/profiler.h
```diff
@@ -0,0 +1,70 @@
...
+    bool operator<(const TProfilingTags& other) const = default;
+};
+
+////////////////////////////////////////////////////////////////////////////////
+
+} // namespace NYT::NQueryTracker
+
+////////////////////////////////////////////////////////////////////////////////
+
+template <>
```
psushin, line 32: usually we don't break the internals of a namespace, and declare THash in the end of the file

yt/yt/server/query_tracker/query_tracker.cpp
```diff
@@ -220,6 +221,28 @@ class TQueryTracker
...
+        LeakySingleton<TActiveQueriesProfilingCountersMap>()->IterateReadOnly([&](const TProfilingTags& tags, const TActiveQueriesProfilingCounter&) {
+            counts[tags] = 0;
+        });
+        for (const auto& record : queryRecords) {
+            auto tags = TProfilingTags{
+                .State = record.State,
+                .Engine = record.Engine,
+                .AssignedTracker = record.AssignedTracker.value_or(NoneQueryTracker),
+            };
+            counts[tags]++;
```
psushin, line 237: we usually use prefix increment iterator when possible

yt/yt/tests/integration/queries/test_mock.py
```diff
@@ -30,6 +31,69 @@ def expect_queries(queries, list_result, incomplete=False):
...
+    @authors("mpereskokova")
+    def test_active_queries_metrics(self, query_tracker):
+        monitoring_port = query_tracker.query_tracker.configs[0]["monitoring_port"]
+
+        def filter_sensor(sensor):
+            return sensor["labels"]["sensor"] == "yt.query_tracker.active_queries" \
+                and "state" in sensor["labels"] \
+                and sensor["labels"]["state"] == "Running"
+
+        def lastActiveQueriesMetric():
```
psushin, line 48: python naming convention violated)

yt/yt/tests/integration/queries/test_mock.py
```diff
@@ -30,6 +31,69 @@ def expect_queries(queries, list_result, incomplete=False):
...
+
+        def get_filter_sensor_lambda(state):
+            def filter_sensor(sensor):
+                return sensor["labels"]["sensor"] == "yt.query_tracker.state_time.max" \
+                    and "state" in sensor["labels"] \
+                    and sensor["labels"]["state"] == state \
+                    and sensor["value"] > 0
+            return filter_sensor
+
+        def hasStateTimeMetric(state):
```
psushin, line 71: python naming convention violated)

yt/yt/tests/integration/queries/test_mock.py
```diff
@@ -30,6 +31,69 @@ def expect_queries(queries, list_result, incomplete=False):
         raise
 
 
+class TestMetrics(YTEnvSetup):
+    DELTA_DRIVER_CONFIG = {
+        "cluster_connection_dynamic_config_policy": "from_cluster_directory",
+    }
+
+    @authors("mpereskokova")
+    def test_active_queries_metrics(self, query_tracker):
```
psushin, line 40: I would suggest looking at profiler_factory()

yt/yt/server/query_tracker/query_tracker_proxy.cpp
```diff
@@ -869,6 +871,25 @@ void TQueryTrackerProxy::AbortQuery(
         }
         THROW_ERROR error;
     }
+
+    // Save profile counter.
+    auto tags = TProfilingTags{
```
psushin, line 876: it could be handy to extract construction of profiling tags from a record into a helper function.

yt/yt/server/query_tracker/query_tracker.cpp
```diff
@@ -220,6 +221,28 @@ class TQueryTracker
             return;
         }
 
+        // Save profile counters.
+        THashMap<TProfilingTags, int> counts;
```
psushin, line 225: may be activeQueryCounts

psushin (changes requested):

Krisha11:
Cmake build crashes as expected. This is due to the fact that cmake files are not being regenerated.

yt/yt/server/query_tracker/profiler-inl.h
```diff
@@ -0,0 +1,30 @@
...
+template<class TProfilingCounter>
+TProfilingCounter* GetOrCreateProfilingCounter(
+    const NProfiling::TProfiler& profiler,
+    const TProfilingTags& profilingTags)
+{
+    using TCountersMap = NConcurrency::TSyncMap<TProfilingTags, TProfilingCounter>;
+
+    auto [counter, _] = LeakySingleton<TCountersMap>()->FindOrInsert(profilingTags, [&]{
+        return TProfilingCounter(profiler
+                .WithTag("state", ToString(profilingTags.State))
```
psushin, line 20: тут отступы уехали

psushin (approved):
Shipped!

Issue #823, opened 2024-09-03
GetQueryTrackerInfo should return query tracker stage


Krock21:
UI needs to know default query tracker stage in order to fetch default ACO for this specific stage.

This should work like this:

1. User opens Queries page and clicks "New Query"
2. UI sends `GetQueryTrackerInfo()` call and looks for `stage` in response. It will indicate default QT stage for backend for all QT-related APIs
3. UI uses `//sys/@ui_config/query_tracker_default_aco/[stage]` (or user override) for a default ACO

This is needed for https://github.com/ytsaurus/ytsaurus-ui/issues/436

Krock21:
Could you please cherry pick it to 23.2, 24.1, 24.2? Seems like new http proxies are requires for this functionality

PR #825, opened 2024-09-04
Add QueryTrackerStage to GetQueryTrackerInfo response. Pass stage through rpc calls

Krock21:
Issue: #823

### DIFF
--- a/yt/yt/server/query_tracker/proxy_service.cpp
+++ b/yt/yt/server/query_tracker/proxy_service.cpp
@@ -61,6 +61,9 @@ private:
         ToProto(rpcResponse->mutable_query_id(), queryId);
 
         TStartQueryOptions options;
+        if (rpcRequest.has_query_tracker_stage()) {
+            options.QueryTrackerStage = rpcRequest.query_tracker_stage();
+        }
         if (rpcRequest.has_settings()) {
             options.Settings = ConvertToNode(TYsonStringBuf(rpcRequest.settings()));
         }
@@ -106,6 +109,9 @@ private:
         auto queryId = FromProto<TQueryId>(rpcRequest.query_id());
 
         TAbortQueryOptions options;
+        if (rpcRequest.has_query_tracker_stage()) {
+            options.QueryTrackerStage = rpcRequest.query_tracker_stage();
+        }
         options.AbortMessage = rpcRequest.has_abort_message()
             ? std::make_optional(rpcRequest.abort_message())
             : std::nullopt;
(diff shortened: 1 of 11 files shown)


yt/yt_proto/yt/client/api/rpc_proxy/proto/api_service.proto
```diff
@@ -3368,6 +3368,7 @@ message TRspGetQueryTrackerInfo
     required string cluster_name = 1;
     required bytes supported_features = 2; // YSON
     repeated string access_control_objects = 3;
+    required string query_tracker_stage = 4;
```
Krock21, line 3371: Will it work with `required`? If only QT is updated or only proxy

Krisha11: I suppose we may have ones in case of new proxies, but you can check for sure

Krock21: Tested. It doesn't break, just returns `"query_tracker_stage":""`

Funny:
https://protobuf.dev/programming-guides/proto2/#field-labels
Screenshot 2024-09-04 at 13 54 22

I will change it to optional

yt/yt/server/query_tracker/proxy_service.cpp
```diff
@@ -106,6 +109,9 @@ class TProxyService
         auto queryId = FromProto<TQueryId>(rpcRequest.query_id());
 
         TAbortQueryOptions options;
+        if (rpcRequest.has_query_tracker_stage()) {
+            options.QueryTrackerStage = rpcRequest.query_tracker_stage();
+        }
```
Krock21, line 114: It is interesting and not clear why we don't have this for GetQueryResult: https://github.com/ytsaurus/ytsaurus/pull/825/files#diff-3218387956293ee1e6820ce23c18208bfe2bfa199e70a2ef32b69ca8467c12eaR145

It doesn't even have Options and doesn't accept a stage

Krisha11: Just because we don't need them for now. Maybe we have no reason to add everywhere
```
        if (rpcRequest.has_query_tracker_stage()) {
            options.QueryTrackerStage = rpcRequest.query_tracker_stage();
        }
```

Krock21: >Just because we don't need them for now.

For me it was unexpected that this parameter is lost somewhere. I would prefer to have all parameters passed to QT proxy, even if they are unused(like stage in most methods). Just so in the next time (the most realistic case is when we will start storing metadata of queries somewhere) we have it and can start using, without searching where it was lost.

In fact, I only discovered it because I added a test for a different stage. It was very easy to miss without proper tests, which may happen in future

For GetQueryResult it seems fine. It is not possible in code to use stage so it  shouldn't be a source of bugs

Krisha11 (approved):

Krock21:
Could you please cherry pick it to 23.2, 24.1, 24.2? Seems like new http proxies are requires for this functionality

PR #827, opened 2024-09-04
Fixing non-cancelling delayed tasks + Advanced monitoring in MultiYtClient

gvsem:
In real world request must be constrained by a timeout. MultiYtClient did not support strict timeouts for hedging subrequests, which could lead to dangling IO connections.

At first, I implemented support of timeouts and specified default hedging request timeout in MultiYtClient builder. We discussed it with @tinarsky and decided to exclude this API.

So, a fix for non-cancelling delayed tasks in MultiExecutor has been added, and more MultiExecutorMonitoring callbacks were added.


tinarsky:
Hello, I'll check it in the next few days.

PR #829, opened 2024-09-05
Add docker pull error attribute about docker_auth presence

koct9i:
This makes missing docker_auth much more obvious.


gritukan (approved):

maxim-babenko:
Shipped!

PR #831, opened 2024-09-05
Suspend jupyts that were not used for a while

dmi-feo:


dmi-feo:
There is one thing that I find a bit ugly: `resume` api command changes yt operation by itself, while all other modifying commands only change the speclet. That's because it might be the first time when the same attribute (`suspend`) has to be changed both by user and strawberry.

There are two alternatives:
1. Introduce `suspended` attribute to oplet, so that `Pass` will suspend marked operations. It would mean that API has to change strawberry persistent state, which is not good at all.
2. Introduce a new type of speclet configuration, managed both by user and by strawberry - a document, for example. The logic is the same like in 1, but API doesn't modify private strawberry params.

However, I find the current implementation less sophisticated that the options above, and the only disadvantage is the `resume` handler.

yt/chyt/controller/internal/agent/agent.go
```diff
@@ -409,10 +409,70 @@ func (a *Agent) Start() {
...
+								oplet.Identificator(),
+								target.InstanceCount,
+							)
+							if target.InstanceCount == 0 {
+								err = oplet.SuspendOp(context.Background(), target.Reason)
+								if err != nil {
+									a.l.Errorf("Failed to suspend %s: %v", oplet.Identificator(), err)
+								}
+							} else {
+								a.l.Error("Not implemented")
```
buyval01, line 458: Let's note that it is scaling for a non-zero number of instances that is not implemented

yt/chyt/controller/internal/strawberry/oplet.go
```diff
@@ -362,6 +367,19 @@ func (oplet *Oplet) CheckOperationLiveness(ctx context.Context) error {
 		log.String("operation_state", string(ytOp.State)))
 
 	oplet.persistentState.YTOpState = ytOp.State
+
+	// oplet.persistentState.YTOpSuspended = ytOp.Suspend
```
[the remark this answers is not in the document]
dmi-feo: The comment below explains why :)
Actually, since https://github.com/ytsaurus/ytsaurus/commit/f3d2c695fb7ad70cb690b25f782a8f83980490ed has been merged, maybe it's time to uncomment the line - I'll check this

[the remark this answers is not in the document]
dmi-feo: There was no go sdk release since the merge, so this crutch is still necessary :(

yt/chyt/controller/internal/api/api.go
```diff
@@ -840,3 +840,23 @@ func (a *API) SetSecrets(ctx context.Context, alias string, secrets map[string]a
...
+	oplet, err := a.getOpletFromCypress(ctx, alias, nil, a.cfg.AgentInfo)
+	if err != nil {
+		return err
+	}
+
+	if err = oplet.ResumeOp(ctx, "api call"); err != nil {
+		return err
+	}
+	err = oplet.EnsurePersistentStateFlushed(ctx)
+	return err
```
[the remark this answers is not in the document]
dmi-feo: I thought one more time about it, and now I think it's not a good idea since it can start or abort an operation straight from API - something that never happened before, AFAIK

yt/chyt/controller/internal/api/api.go
```diff
@@ -840,3 +840,22 @@ func (a *API) SetSecrets(ctx context.Context, alias string, secrets map[string]a
...
+
+	oplet, err := a.getOpletFromCypress(ctx, alias, nil, a.cfg.AgentInfo)
+	if err != nil {
+		return err
+	}
+
+	if err = oplet.ResumeOp(ctx, "api call"); err != nil {
+		return err
+	}
+	return oplet.Pass(ctx)
```
DimasKovas, line 860: I don't like resuming the op and making the pass here.

We do so for `untracked` operations, because there is no controller who tracks them.

But here the API and the controller both read/write the persistent state and it may lead to conflicts (you may loose part of persistent state).

The `resume` operation is not easily expressible via speclet, but as a workaround I suggest to introduce a separate option like `latest_resume_time` - and just set it to `time.Now()` here. The actual `resume` logic should be moved to the controller based on this value. Another alternative is to introduce the option `resume_stamp` (not sure about the naming) and set it to a random int64 value. Then you can check if the value has been changed in the controller and resume operation.

Moreover, the `resume` command is jupyt-related, let's move it to `jupyt` folder. And actually we'are implementing a `restart` command now. It looks like `restart` can be an alternative to `resume` operation. Do we really need the `resume` command?

yt/chyt/controller/internal/jupyt/speclet.go
```diff
@@ -5,6 +5,7 @@ type Speclet struct {
 	Memory *uint64 `yson:"memory"`
 
 	JupyterDockerImage string `yson:"jupyter_docker_image"`
+	IdleTimeoutSec     uint64 `yson:"idle_timeout_sec"`
```
DimasKovas, line 8: Looks more like `yson.Duration` to me

DimasKovas (changes requested):

yt/chyt/controller/internal/jupyt/controller.go
```diff
@@ -139,6 +162,20 @@ func (c *Controller) DescribeOptions(parsedSpeclet any) []strawberry.OptionGroup
...
+			Title: "Idle timeout",
+			Options: []strawberry.OptionDescriptor{
+				{
+					Title:        "Idle timeout",
+					Name:         "idle_timeout_sec",
+					Type:         strawberry.TypeInt64,
+					CurrentValue: speclet.IdleTimeoutSec,
+					DefaultValue: 0,
+					MinValue:     0,
+					Description:  "Jupyt operation will be suspended in case of no activity for the specified time. 0 means 'never'.",
```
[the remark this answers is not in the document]
dmi-feo: > Why 0 means never? 

Maybe it's not that common, but still used - in `timeout` util, for example. But let's keep it consistent with CHYT, I'll change it to `enable_` + value field.

yt/chyt/controller/internal/strawberry/oplet.go
```diff
@@ -362,6 +367,20 @@ func (oplet *Oplet) CheckOperationLiveness(ctx context.Context) error {
 		log.String("operation_state", string(ytOp.State)))
 
 	oplet.persistentState.YTOpState = ytOp.State
+
+	// oplet.persistentState.YTOpSuspended = ytOp.Suspended
+	// ytOp.Suspend always returns false, so let's use running job count until it's fixed
+	// TODO: wait for https://github.com/ytsaurus/ytsaurus/commit/f3d2c695fb7ad70cb690b25f782a8f83980490ed to be released
+	jobRes, err := oplet.systemClient.ListJobs(ctx, opID, &yt.ListJobsOptions{JobState: &yt.JobRunning})
```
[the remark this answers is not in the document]
dmi-feo: It's not required as long as we use user_slots->0 instead of suspending, but it was the only way to detect whether the operation has been suspended

yt/chyt/controller/internal/jupyt/controller.go
```diff
@@ -40,7 +63,7 @@ func (c *Controller) UpdateState() (changed bool, err error) {
 }
 
 func (c *Controller) buildCommand(speclet *Speclet) (command string, env map[string]string) {
-	cmd := "bash -x start.sh /opt/conda/bin/jupyter lab --ip '*' --port $YT_PORT_0 --LabApp.token='' --allow-root >&2"
+	cmd := c.config.CommandOrDefault()
```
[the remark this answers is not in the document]
dmi-feo: Long story short - we had to add a custom activity handler (that allows to detect idle timeouts) to jupyter server (the default one requires auth - it's not a problem when you use a vanilla jupyter with LabApp.token='', but we have yt-auth based identity enabled), and it turned out that custom plugins don't work unless the jupyter config is specified in the startup command, so we had to modify the command.

I can move it to another PR, if you want.

yt/chyt/controller/internal/jupyt/controller.go
```diff
@@ -1,23 +1,32 @@
...
+	"net/http"
+	"time"
 )
 
 type Config struct {
-	YTAuthCookieName *string           `yson:"yt_auth_cookie_name"`
-	ExtraEnvVars     map[string]string `yson:"extra_env_vars"`
+	YTAuthCookieName   *string           `yson:"yt_auth_cookie_name"`
+	ExtraEnvVars       map[string]string `yson:"extra_env_vars"`
+	LastActivityRelUrl *string           `yson:"last_activity_rel_url"`
```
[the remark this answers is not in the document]
dmi-feo: `rel` means "relative" in this case, but I'll rename it, no problem

yt/chyt/controller/internal/api/api.go
dmi-feo: Well, I thought about this option... Can't say I like it (nor the other options), but you are right, it fits the current strawberry logic better. 

> the resume command is jupyt-related, let's move it to jupyt folder

It's a command for all controller families that support scaling to 0, isn't it? I don't think jupyt will be unique for that matter.

yt/chyt/controller/internal/jupyt/controller.go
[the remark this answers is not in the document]
DimasKovas: I see. No need to move to another PR, it's fine

yt/chyt/controller/internal/agent/agent.go
```diff
@@ -409,10 +409,70 @@ func (a *Agent) Start() {
...
+								oplet.Identificator(),
+								target.InstanceCount,
+							)
+							if target.InstanceCount == 0 {
+								err = oplet.SuspendOp(a.ctx, target.Reason)
+								if err != nil {
+									a.l.Errorf("Failed to suspend %s: %v", oplet.Identificator(), err)
+								}
+							} else {
+								a.l.Error("Scaling to a non-zero instance number is not implemented yet")
```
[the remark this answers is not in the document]
dmi-feo: Is it a common way to set user_slot to 0? I got an error
```
Operation has an ancestor whose specified resource limits are too small to satisfy operation's minimum job resource demand
```
which leads to the operation failure. 

And it seems that changing `resource_limits/user_slots` to 0 on a running operation has no effect - but I found `scheduling_options_per_pool_tree/default/resource_limits/user_slots` in python wrapper tests and it works.

dmi-feo:
I have an issue with user_slots. This script - https://gist.github.com/dmi-feo/c8cbf868c81d8c2ab80486e98e9bd621 - doesn't work for me, number of operation jobs doesn't change. What do I do wrong?

dmi-feo:
Setting `scheduling_options_per_pool_tree/<name>/resource_limits/user_slots` works for num_slots > 0, but when I set it to 0, I eventually get:
```
Error
Operation scheduling hung in all trees[1]
Operation has an ancestor whose specified resource limits are too small to satisfy operation's minimum job resource demand[1]
```

Can it be disabled somehow? Or maybe let's switch back to the suspension?

DimasKovas:
> Setting `scheduling_options_per_pool_tree/<name>/resource_limits/user_slots` works for num_slots > 0, but when I set it to 0, I eventually get:
> 
> ```
> Error
> Operation scheduling hung in all trees[1]
> Operation has an ancestor whose specified resource limits are too small to satisfy operation's minimum job resource demand[1]
> ```
> 
> Can it be disabled somehow? Or maybe let's switch back to the suspension?

Oh, I didn't know about that, sorry.

Yeah, it looks like that handling the zero case separately is the only way to go.

But I would suggest to consider not suspending but actually aborting the operation (e.g. return true in `needsAbort` and setting special flag `stopped_by_scaler` or smth), In that case we won't need all this complex logic about suspending/resuming (and the restart command would fit more nicely to that case without the need for a separate `resume` command)

DimasKovas:
> I have an issue with user_slots. This script - https://gist.github.com/dmi-feo/c8cbf868c81d8c2ab80486e98e9bd621 - doesn't work for me, number of operation jobs doesn't change. What do I do wrong?

You are waiting for only 30 seconds. I guess the scheduler isn't so quick to react

yt/chyt/controller/internal/agent/agent.go
```diff
@@ -463,10 +483,63 @@ func (a *Agent) Start() {
...
+		case <-ticker.C:
+			startedAt := time.Now()
+
+			a.l.Info("Starting scaler")
+
+			workerNumber := a.config.ScaleWorkerNumberOrDefault()
+			var wg sync.WaitGroup
+			wg.Add(workerNumber)
+
+			oplets := <-a.opletBatchCh
```
DimasKovas, line 511: Here there is the same race condition. You may access one oplet from both background and runScaler goroutines.
It is OK for fields like `Alias`, but you also access `Active()`, which surely is a race condition.

I think you need to copy all needed fields to a special struct in the background goroutine (you need only `opID` and probably `alias` I guess) and pass this to controller.GetScalerTarget(...) here instead of the oplet

yt/chyt/controller/internal/agent/agent.go
```diff
@@ -492,3 +565,8 @@ func (a *Agent) OperationNamespace() string {
 func (a *Agent) CheckHealth() error {
 	return a.healthState.Get()
 }
+
+type scalingRequest struct {
+	OpletAlias    string
```
DimasKovas, line 570: Let's add opID here to be sure that we won't apply outdated scalingRequest to a new operation after its restart

yt/chyt/controller/go.mod
```diff
@@ -1,6 +1,6 @@
 module go.ytsaurus.tech/yt/chyt/controller
 
-go 1.20
+go 1.23
```
DimasKovas, line 3: We've fixed the compilation under go 1.20 in the master. Please, rebase the brunch and remove this

yt/chyt/controller/internal/jupyt/controller.go
```diff
@@ -27,6 +36,20 @@ func (c *Config) YTAuthCookieNameOrDefault() string {
 	return DefaultYTAuthCookieName
 }
 
+func (c *Config) LastActivityRelUrlOrDefault() string {
```
DimasKovas, line 39: URLPath

yt/chyt/controller/internal/agent/agent.go
```diff
@@ -385,10 +388,27 @@ func (a *Agent) background() {
 			a.healthState.SetTrackOpsState(nil)
 		case <-ticker.C:
 			a.pass()
+		case a.opletBatchCh <- a.getOpletMappingCopy():
+			continue
+		case scT := <-a.scalingTargetCh:
+			oplet, ok := a.aliasToOp[scT.OpletAlias]
+			if !ok {
+				a.l.Warnf("Oplet %s not found, skipping scaling")
```
DimasKovas, line 396: We usually don't use Warn**f**/Info**f**/etc. Our log is structured, so it's better to write
```
a.l.Warn("Oplet not found, skipping scaling", log.String("alias", scT.OpletAlias))
```

yt/chyt/controller/internal/agent/agent.go
```diff
@@ -492,3 +565,8 @@ func (a *Agent) OperationNamespace() string {
 func (a *Agent) CheckHealth() error {
 	return a.healthState.Get()
 }
+
+type scalingRequest struct {
+	OpletAlias    string
+	InstanceCount int64
```
DimasKovas, line 571: Golang suggests using int:
```
When you need an integer value you should use int unless you have a specific reason to use a sized or unsigned integer type. 
```

yt/chyt/controller/internal/jupyt/controller.go
```diff
@@ -139,6 +162,33 @@ func (c *Controller) DescribeOptions(parsedSpeclet any) []strawberry.OptionGroup
...
+		},
+		{
+			Title: "Idle timeout",
+			Options: []strawberry.OptionDescriptor{
+				{
+					Title:        "Idle timeout",
+					Name:         "idle_timeout",
+					Type:         strawberry.TypeYson,
+					CurrentValue: speclet.IdleTimeout,
+					DefaultValue: 0,
```
DimasKovas, line 186: DefaultIdleTimeout

yt/chyt/controller/internal/strawberry/controller.go
```diff
@@ -46,6 +46,13 @@ type Controller interface {
 	// Given speclet should have suitable type for the specific controller.
 	// Otherwise, GetOpBriefAttributes may panic.
 	GetOpBriefAttributes(parsedSpeclet any) map[string]any
+
+	// GetScalerTarget checks whether YT operation should be scaled and returns required scaling parameters.
+	// Return `nil` is scaling is not required.
+	GetScalerTarget(ctx context.Context, oplet *Oplet) (*ScalerTarget, error)
+
+	// NonRestartingOpSpecletFields returns "set" of fields that should not cause oplet restart when changed.
+	NonRestartingOpSpecletFields() map[string]struct{}
```
DimasKovas, line 55: I don't like this method in the interface.

I suggest to add a tag to the fields (like `yson:"abcd"`, but smth like `requires_restart:false`).
If there is such tag on the field, use it, otherwise assume that it `requires_restart`

yt/chyt/controller/internal/strawberry/controller.go
```diff
@@ -46,6 +46,13 @@ type Controller interface {
 	// Given speclet should have suitable type for the specific controller.
 	// Otherwise, GetOpBriefAttributes may panic.
 	GetOpBriefAttributes(parsedSpeclet any) map[string]any
+
+	// GetScalerTarget checks whether YT operation should be scaled and returns required scaling parameters.
+	// Return `nil` is scaling is not required.
+	GetScalerTarget(ctx context.Context, oplet *Oplet) (*ScalerTarget, error)
```
DimasKovas, line 52: Since we access this method from another goroutine (not from background goroutine), we need to state it in the comment that `this method may be called concurrently ...`

DimasKovas (changes requested):

yt/chyt/controller/internal/agent/agent.go
dmi-feo: I guess the worst thing that could happen is an attempt to check a disabled oplet, but you are right, I don't need the whole oplet, I'll make another struct for it

dmi-feo:
> You are waiting for only 30 seconds. I guess the scheduler isn't so quick to react

Doesn't matter, I've tried waiting for 5 minutes with the same result. Does this script - https://gist.github.com/dmi-feo/98c5928e67c3aeada0f52cce9e9c757b - work for you? Maybe it is caused by some differences in cluster configurations

DimasKovas:
> > You are waiting for only 30 seconds. I guess the scheduler isn't so quick to react
> 
> Doesn't matter, I've tried waiting for 5 minutes with the same result. Does this script - https://gist.github.com/dmi-feo/98c5928e67c3aeada0f52cce9e9c757b - work for you? Maybe it is caused by some differences in cluster configurations

I'll check

dmi-feo:
> But I would suggest to consider not suspending but actually aborting the operation (e.g. return true in needsAbort and setting special flag stopped_by_scaler or smth), In that case we won't need all this complex logic about suspending/resuming (and the restart command would fit more nicely to that case without the need for a separate resume command)

That would make difference between zero and non-zero cases even bigger - while we just update op params in non-zero case, in zero case operation has to be completely stopped. And it brings symmetrical problem - we need to modify from the agent something that is usually controlled by user. 

Another reason - I'm thinking about possibility of saving/restoring state in case os suspension, and it would be easier if we can guarantee the speclet (non-restarting part of it) has not been changed (so runtime is the same). However, it's not a blocker, as far as I can see.

Anyway, I suggest using suspension here, if there is no other reasons to avoid it

yt/chyt/controller/internal/jupyt/controller.go
```diff
@@ -1,23 +1,32 @@
...
-
+	"encoding/json"
+	"fmt"
 	"go.ytsaurus.tech/library/go/core/log"
 	"go.ytsaurus.tech/yt/chyt/controller/internal/strawberry"
 	"go.ytsaurus.tech/yt/go/ypath"
 	"go.ytsaurus.tech/yt/go/yson"
 	"go.ytsaurus.tech/yt/go/yt"
 	"go.ytsaurus.tech/yt/go/yterrors"
+	"io"
```
DimasKovas, line 14: Did you run `go fmt`?
My version sorts the import paths differently (system imports are separated from the `go.ytsaurus.tech` ones)

yt/chyt/controller/internal/jupyt/controller.go
```diff
@@ -193,6 +243,54 @@ func (c *Controller) appendConfigs(ctx context.Context, oplet *strawberry.Oplet,
...
+		LastActivity time.Time `json:"last_activity"`
+	}
+	err = json.NewDecoder(bytes.NewReader(body)).Decode(&jupyterResponse)
+	if err != nil {
+		return nil, err
+	}
+
+	speclet := opletInfo.ControllerSpeclet.(Speclet)
+	sinceLastActivity := time.Now().Sub(jupyterResponse.LastActivity)
+	if speclet.EnableIdleSuspension && sinceLastActivity > speclet.IdleTimeoutOrDefault() {
```
DimasKovas, line 278: If "idle suspension" is disabled we don't need to get endpoint/make the http request - it's a waste of time

yt/chyt/controller/internal/strawberry/oplet.go
```diff
@@ -381,6 +397,19 @@ func (oplet *Oplet) CheckOperationLiveness(ctx context.Context) error {
 		log.String("operation_id", opID.String()),
 		log.String("operation_state", string(ytOp.State)))
 
+	// oplet.persistentState.YTOpSuspended = ytOp.Suspended
+	// ytOp.Suspend always returns false, so let's use running job count until it's fixed
+	// TODO: wait for https://github.com/ytsaurus/ytsaurus/commit/f3d2c695fb7ad70cb690b25f782a8f83980490ed to be released
+	jobRes, err := oplet.systemClient.ListJobs(ctx, opID, &yt.ListJobsOptions{JobState: &yt.JobRunning})
```
DimasKovas, line 403: This call is still very undesirable.
First, `len(jobs) == 0` doesn't equal `suspended`.
Second, I don't understand why we need to change YTOpSuspended here at all. We change it during `suspendOp`/`resumeOp` - that should be enough.
Third, it's expensive.

yt/chyt/controller/internal/strawberry/oplet.go
```diff
@@ -984,6 +1101,14 @@ func (oplet *Oplet) Pass(ctx context.Context, checkOpLiveness bool) error {
 		return err
 	}
 
+	if oplet.strawberrySpeclet.ResumeMarker != oplet.persistentState.ResumeMarker {
+		// Maybe put into persistent state, so resuming will be not interrupted by strawberry restart?
+		oplet.pendingResume = true
+		oplet.persistentState.ResumeMarker = oplet.strawberrySpeclet.ResumeMarker
+		oplet.persistentState.YTOpSpeclet = oplet.specletYson
```
DimasKovas, line 1108: YTOpSpeclet/YTOpSpecletRevision relate to the speclet/revision the operations was started with. It should be changed only during the operation start.

And I don't understand the `pendingResume` flag purpose. We can compare persistentState.ResumeMarker vs speclet.ResumeMarker in `needsResume` and update persistentState.ResumeMarker in `resumeOp` - this looks more reasonable to me.

DimasKovas:
About aborting jobs after decreasing resource_limits - colleagues say that it should work, but it doesn't. They will investigate the issue

DimasKovas (changes requested):

yt/chyt/controller/internal/agent/agent.go
```diff
@@ -463,10 +490,75 @@ func (a *Agent) Start() {
...
+							)
+							continue
+						}
+						if target != nil {
+							a.l.Info(
+								"Oplet must be scaled",
+								log.String("alias", opletInfo.Alias),
+								log.Int("target_instance_count", target.InstanceCount),
+							)
+							a.scalingTargetCh <- scalingRequest{
```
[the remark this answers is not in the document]
dmi-feo: Not intended, I overlooked that background goroutine can get busy for a while

yt/chyt/controller/internal/strawberry/oplet.go
dmi-feo: > First, len(jobs) == 0 doesn't equal suspended.

That's (maybe) the best approach until go sdk fix is released.

> I don't understand why we need to change YTOpSuspended here at all

I wanted to keep it consistent with operations state - we check it in this function. If you think it's not necessary - I'll delete it.

yt/chyt/controller/internal/jupyt/controller.go
dmi-feo: It seems you have `goimports`, which sorts imports the way you expect. I installed it too.

yt/chyt/controller/internal/strawberry/oplet.go
```diff
@@ -888,13 +979,25 @@ func (oplet *Oplet) updateOpParameters(ctx context.Context, reason string) error
...
+	}
+	if oplet.pendingScaling && oplet.targetInstanceCount != 0 {
+		params["scheduling_options_per_pool_tree"] = map[string]any{
+			"default": map[string]any{
+				"resource_limits": map[string]any{
+					"user_slots": oplet.targetInstanceCount,
+				},
+			},
+		}
+		oplet.pendingScaling = false
```
DimasKovas, line 994: To the end of the function (so only update if there is no error in `UpdateOperationParameters` call).
Also it's possible to eliminate `pending` flag at all (e.g. we do not have pending flag for `acl`), but it's not required, you can leave it as is

yt/chyt/controller/internal/strawberry/oplet.go
```diff
@@ -503,6 +533,32 @@ func (oplet *Oplet) needsAbort() (needsAbort bool, reason string) {
 	return false, "up to date"
 }
 
+func (oplet *Oplet) needsSuspend() (needsSuspend bool, reason string) {
```
DimasKovas, line 536: Does it work as expected if we try to suspend already suspended operation, or try to resume not suspended operation?

yt/chyt/controller/internal/strawberry/state.go
```diff
@@ -11,8 +11,11 @@ import (
 // It is written to the cypress when changed and is read from the cypress when the cypress revision is changed.
 // The persistence of the state is needed for fault tolerance.
 type PersistentState struct {
-	YTOpID    yt.OperationID    `yson:"yt_operation_id"`
-	YTOpState yt.OperationState `yson:"yt_operation_state"`
+	YTOpID        yt.OperationID    `yson:"yt_operation_id"`
+	YTOpState     yt.OperationState `yson:"yt_operation_state"`
+	YTOpSuspended bool              `yson:"yt_operation_suspended"`
+
+	ResumeMarker string `yson:"resume_marker"`
```
DimasKovas, line 18: Let's add a comment about what `ResumeMarker` is about

DimasKovas (changes requested):
LGTM
There are only minor issues left

yt/chyt/controller/internal/strawberry/oplet.go
dmi-feo: You mean `suspendOp` and `resumeOp` methods? In python sdk suspending a suspended operation does nothing, while resuming a not suspended operation returns an error. I guess both options are acceptable in our case.

yt/chyt/controller/internal/agent/agent.go
```diff
@@ -407,12 +408,12 @@ func (a *Agent) background() {
 	}
 }
 
-func (a *Agent) getOpletInfoForScaler() []strawberry.OpletInfoForScaler {
-	opletInfos := make([]strawberry.OpletInfoForScaler, 0, len(a.aliasToOp))
+func (a *Agent) getOpletInfoForScaler() []strawberry.OpletInfosForScaler {
```
DimasKovas, line 411: A little bit of misunderstanding here :)
My comment was about `getOpletInfoForScaler` naming.
One oplet info is `OpletInfoForScaler`, withous **s**, but since this function returns an array of `OpletInfo` structs, it should be called `getOpletInfos...` with **s**

dmi-feo: Oh, shame on me. I'll fix it asap

yt/chyt/controller/internal/strawberry/oplet.go
DimasKovas: > You mean `suspendOp` and `resumeOp` methods? In python sdk suspending a suspended operation does nothing, while resuming a not suspended operation returns an error. I guess both options are acceptable in our case.

Yes. It seems to me that if you call the `resume` command on a Jupyter that is not suspended, the `resume_marker` will be changed and the controller will attempt to resume the not-suspended operation indefinitely on every pass.

dmi-feo: Oh, you are right, let's handle this case in `resumeOp`.

DimasKovas (approved):

DimasKovas (approved):

DimasKovas (approved):

DimasKovas (approved):

PR #835, opened 2024-09-08
Add whoami method to Go SDK

### LINKED ISSUE #760: Add whoami method to Go SDK
Sometimes, we need to simply check the validity of a user’s token

wilwell:
Add method `whoami` added to the Go SDK functionality. 

To invoke the method user only need to create a client with correct credentials (Cookie or Token for example). Unlike other yt methods, `whoami `response is JSON not YSON, and uses another handler path: `/auth/` instead of `api/v4`.

Right now the *testing* docker container of YT doesn't support the authentication, so for every request `whoami` we are getting `root` response. When the testing docker container will be fixed, the auth test also could be fixed.

Closes #760

### DIFF
--- a/yt/go/yt/integration/auth_client_test.go
+++ b/yt/go/yt/integration/auth_client_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/sha256"
 	"fmt"
-	"go.ytsaurus.tech/yt/go/yt/ythttp"
 	"os"
 	"testing"
 
@@ -13,6 +12,7 @@ import (
 	"go.ytsaurus.tech/yt/go/guid"
 	"go.ytsaurus.tech/yt/go/ypath"
 	"go.ytsaurus.tech/yt/go/yt"
+	"go.ytsaurus.tech/yt/go/yt/ythttp"
 )
 
 func TestAuthClient(t *testing.T) {
--- a/yt/go/yt/integration/auth_client_test.go
+++ b/yt/go/yt/integration/auth_client_test.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"crypto/sha256"
 	"fmt"
+	"go.ytsaurus.tech/yt/go/yt/ythttp"
+	"os"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -19,6 +21,7 @@ func TestAuthClient(t *testing.T) {
 	suite.RunClientTests(t, []ClientTest{
 		{Name: "SetUserPassword", Test: suite.TestSetUserPassword, SkipRPC: true},
 		{Name: "IssueListRevokeToken", Test: suite.TestIssueListRevokeToken, SkipRPC: true},
+		{Name: "WhoAmI", Test: suite.TestWhoAmI, SkipRPC: true},
 	})
 }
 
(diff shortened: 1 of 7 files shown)


wilwell:
cc @l0kix2

yt/go/yt/integration/auth_client_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"crypto/sha256"
 	"fmt"
+	"go.ytsaurus.tech/yt/go/yt/ythttp"
+	"os"
```
l0kix2, line 8: It should be linters' job to say that, but these imports are not sorted properly. I suggest you configure your IDE for auto sorting (and we also need lint workflow here of course)

yt/go/yt/integration/auth_client_test.go
```diff
@@ -65,6 +68,22 @@ func (s *Suite) TestIssueListRevokeToken(ctx context.Context, t *testing.T, yc y
...
 }
 
+func (s *Suite) TestWhoAmI(ctx context.Context, t *testing.T, yc yt.Client) {
+	user := "user-" + guid.New().String()
+	_ = s.CreateUser(ctx, t, user)
+	token, err := yc.IssueToken(ctx, user, "", nil)
+	require.NoError(t, err)
+	require.NotEmpty(t, token)
+
+	userClient, err := ythttp.NewClient(&yt.Config{Proxy: os.Getenv("YT_PROXY"), Logger: s.Env.L, Token: token})
```
l0kix2, line 78: Offtop, but I don't like this idea of passing proxy for test in env var, but I see that the current test run configuration doesn't have another way for it.

yt/go/yt/internal/call.go
```diff
@@ -56,6 +58,11 @@ func (res *CallResult) decode(value any) (err error) {
 	return
 }
 
+func (res *CallResult) decodeJSON(value any) (err error) {
+	err = json.Unmarshal(res.YSONValue, value)
+	return
+}
```
l0kix2, line 64: Could you provide some context (or maybe in the code comment also) 
1) why CallResult require JSON decoder now and didn't need it before? All other methods are returning YSON?
2)Am I understanding correctly that something called YSONValue contains JSON?

l0kix2 (changes requested):

yt/go/yt/integration/auth_client_test.go
wilwell: Yes, I agree that tis looks ugly, but I copied this method from the client creation in other testing methods.

yt/go/yt/internal/call.go
wilwell: 1. Even if we specify the output format is specified in header as YSON the whoami handler returns the JSON value.
2. YSONValue contains the byte representation of response body, it's inly named YSON

I've added couple more comments here and there.

l0kix2 (approved):
LGTM

yt/go/yt/internal/call.go
```diff
@@ -56,6 +58,12 @@ func (res *CallResult) decode(value any) (err error) {
 	return
 }
 
+func (res *CallResult) decodeJSON(value any) (err error) {
+	// YSONValue contains the body of the response, it's not necessary YSON format, e.g. for whoami request
```
tinarsky, line 62: nit: dot at the end of the comment.

tinarsky (changes requested):

yt/go/yt/internal/call.go
wilwell: Done.

tinarsky (approved):
LGTM

Issue #836, opened 2024-09-09
Go SDK: GetOperation always returns suspend=false


dmi-feo:
How to reproduce: https://gist.github.com/dmi-feo/de83c956f604734a9b9829c643f3620a

```
Operation: 70f933ab-efa2d698-134403e8-8b789332
Waiting for the jobs to start
Num of running jobs: 0
Num of running jobs: 0
Num of running jobs: 1
All jobs have been started
Suspending operation
Operation has been suspended, waiting for jobs to be terminated
Num of running jobs: 1
Num of running jobs: 0
All jobs have been terminated
error: API returned suspended=false
```

PR #837, opened 2024-09-09
Go sdk: fix suspend property of operation status

dmi-feo:


dmi-feo:
resolves https://github.com/ytsaurus/ytsaurus/issues/836

yt/go/yt/interface.go
```diff
@@ -677,7 +677,7 @@ type OperationStatus struct {
 	FullSpec          yson.RawValue              `yson:"full_spec"`
 	StartTime         yson.Time                  `yson:"start_time"`
 	FinishTime        yson.Time                  `yson:"finish_time"`
-	Suspend           bool                       `yson:"suspend"`
+	Suspend           bool                       `yson:"suspended"`
```
zlobober, line 680: Let's rename the field name too. I doubt that there is much code that would break, taking into account that the field was never working.

yt/go/yt/integration/operations_test.go
```diff
@@ -131,6 +132,65 @@ func TestOperationWithStderr(t *testing.T) {
...
+			time.Sleep(3 * time.Second)
+		} else {
+			break
+		}
+	}
+
+	// Check the operation status
+	opStatus, err := env.YT.GetOperation(ctx, opID, &yt.GetOperationOptions{})
+	require.NoError(t, err)
+	assert.Equal(t, yt.StateRunning, opStatus.State)
```
zlobober, line 190: There is a require.Equal, why can't we use it? Is a new dependency really needed?

yt/go/yt/interface.go
dmi-feo: Let's keep both for some time maybe? Even though it was not working, some code could rely on its existence

yt/go/yt/integration/operations_test.go
dmi-feo: My bad, will fix

yt/go/yt/interface.go
dmi-feo: And it seems it worked for rpc client

dmi-feo: okay

yt/go/yt/integration/operations_test.go
dmi-feo: btw, it would have not become a really new dependency in the project, unfortunately: https://github.com/search?q=repo%3Aytsaurus%2Fytsaurus+github.com%2Fstretchr%2Ftestify%2Fassert+path%3A%2F%5Eyt%5C%2Fgo%5C%2Fyt%5C%2Fintegration%5C%2F%2F&type=code

yt/go/yt/integration/operations_test.go
```diff
@@ -131,6 +131,65 @@ func TestOperationWithStderr(t *testing.T) {
...
+				t.Fatal("The job did not stop")
+			}
+			time.Sleep(3 * time.Second)
+		} else {
+			break
+		}
+	}
+
+	// Check the operation status
+	opStatus, err := env.YT.GetOperation(ctx, opID, &yt.GetOperationOptions{})
```
tinarsky, line 187: The third parameter may be nil.

yt/go/yt/integration/operations_test.go
```diff
@@ -131,6 +131,65 @@ func TestOperationWithStderr(t *testing.T) {
...
+				"job_count": 1,
+				"command":   "sleep 5m",
+			},
+		},
+	}
+	opID, err := env.YT.StartOperation(ctx, yt.OperationVanilla, opSpec, nil)
+	require.NoError(t, err)
+
+	// Wait for the job to start.
+	for idx := 0; idx < 10; idx++ {
```
tinarsky, line 155: I suggest waiting here for the operation state to be `Running`. And then we could check that `require.Equal(t, false, opStatus.Suspended)`.

tinarsky (changes requested):

tinarsky (approved):
LGTM

PR #838, opened 2024-09-09
Add System Python UDF documentation

Krock21:
Added Python UDF documentation under Query Tracker section. Checked locally


Krisha11:
LGTM

ogorbacheva (approved):
OK

Krisha11 (approved):

Krisha11:
Why are we putting this in query_tracker chapter and not in the yql one?

Krock21:
> Why are we putting this in query_tracker chapter and not in the yql one?

We agreed that I will put it here as the best place available for external people

YQL documentation is special. It's source of truth is combined from ydb repository and internal Yandex repository. YT specific things are not included in ydb repository so the only thing that left for yt-specific YQL things in internal Yandex repository.

Later this part can be moved to a better place. cc @ogorbacheva @blinkov

PR #840, opened 2024-09-10
Increase default rate_limit for caching_object_service

achulkov2:
The current default of 10 is impractical for master caches. For example, we make 16 * 16 requests to Cypress when listing operations, which would take 30 seconds to complete with this rate limit. We increase it to 1000, which is more commonly used for production YT installations.


maxim-babenko:
Shipped

maxim-babenko:
Shipped!

PR #841, opened 2024-09-10
add auth to yt_local

dmi-feo:


yt/python/yt/environment/yt_env.py
```diff
@@ -66,6 +67,11 @@
 _environment_driver_logging_config = None
 
 
+DEFAULT_ADMIN_USERNAME = "admin"
+DEFAULT_ADMIN_PASSWORD = "admin"
+DEFAULT_ADMIN_TOKEN = "admin"
```
koct9i, line 72: k8s operator uses "admin" "password" "password"

https://github.com/ytsaurus/ytsaurus-k8s-operator/blob/main/pkg/consts/defaults.go

https://github.com/ytsaurus/ytsaurus-k8s-operator/blob/7d512528654291ce841e5f468cd01becde77bc12/pkg/components/master.go#L120

dmi-feo: okay, let it be "password"

Kontakter:
I will read this PR in few days from now

dmi-feo:
upd: now `use_native_client` is ignored only for `create_admin_user` procedure. Not sure if it's real to use non-native client there (start http-proxy without auth, create a user, then restart http-proxy with auth? is it worth it?)

yt/python/yt/environment/yt_env.py
```diff
@@ -609,6 +615,10 @@ def start(self, on_sequoia_started_func=None, on_masters_started_func=None):
             if self.yt_config.replicated_table_tracker_count > 0:
                 self.start_replicated_table_trackers(sync=False)
 
+            if self.yt_config.enable_auth:
+                # The following `synchroize` might use http client, so the user has to be created before that.
```
Kontakter, line 619: `synchronize`

yt/python/yt/environment/yt_env.py
```diff
@@ -609,6 +615,10 @@ def start(self, on_sequoia_started_func=None, on_masters_started_func=None):
...
+                self.create_admin_user()
```
Kontakter, line 620: I think in this point the cluster might not be started (synchronized) yet, so this calls may fail. I think that it should be done _after_ starting masters and initial synchronization, but _before_ starting http proxies

dmi-feo: This means that http proxies should be started after the initial synchronization, but this synchronization itself requires http proxies - unless we start bootstrapping the cluster using native client only, no matter what the value of `use_native_client` is. 

BTW, if we make `create_admin_user` optional, we will face the same issue - http client won't be able to be used as  `cluster_client`. 

Are we going to always use native client as `cluster_client`?

Kontakter:
We have discussed the plan:
1. `use_native_client` is wrong name of the option, proper name is `native_client_supported` (I have renamed it yesterday)
2. There is a lot of internal tests that do not want to rely on native_client (but auth is not needed for them)
3. In yt local docker image we can suppose that `native_client_supported` is always true and in this case logic gets simpler
4. If `enable_auth` is true we want to require `native_client_supported` to be also true

Kontakter:
Related changes:
1. Check that all internal usages of client in YtEnvironment use `_create_cluster_client`
2. Simplify this `if` statement https://a.yandex-team.ru/arcadia/yt/python/yt/environment/yt_env.py?rev=r15023221#L1826, check for `cypress_proxy_count` looks weird

yt/docker/local/run_local_cluster.sh
```diff
@@ -13,7 +13,7 @@ script_name=$0
 proxy_port=8000
 docker_hostname=localhost
 interface_port=8001
-yt_version=stable
+yt_version=dev
```
dmi-feo, line 16: not sure about this change, but `stable` seems pretty outdated, it doesn't even support the native client setting

yt/docker/local/run_local_cluster.sh
```diff
@@ -212,6 +219,10 @@ if [ ${enable_debug_logging} == "true" ]; then
     params="--enable-debug-logging"
 fi
 
+if [ ${with_auth} == "true" ]; then
+  params="${params} --enable-auth --create-admin-user"
```
Kontakter, line 223: Use 4 spaces for identation

Kontakter (approved):
LGTM

Kontakter:
Some style errors:
  * `yt/python/yt/local/commands.py:180: [E124] closing bracket does not match visual indentation`
  * `yt/python/yt/environment/yt_env.py:701: [E231] missing whitespace after ','`

yt/docker/local/run_local_cluster.sh
Kontakter: @savnadya – what do you think about this change? (stable -> dev)

savnadya: I don't think it's a good idea. `dev` image is updated after each commit, and it can break easily. You can just push a new stable version of the image.

dmi-feo: I doubt I have access to push a new image - who can do this?

dmi-feo:
I fixed the style errors and rolled back yt_version to stable. Now we need to:
1. Update stable image before merging - to make `native-client-supported` work
2. Update stable image after merging - to  make auth params in yt_local work

Who can do that?

savnadya:
> I fixed the style errors and rolled back yt_version to stable. Now we need to:
> 
> 1. Update stable image before merging - to make `native-client-supported` work
> 2. Update stable image after merging - to  make auth params in yt_local work
> 
> Who can do that?

Hi! It looks like you have all the necessary rights to do this.

dmi-feo:
```
$ docker push ghcr.io/ytsaurus/local:stable
The push refers to repository [ghcr.io/ytsaurus/local]
4430078076e5: Layer already exists
9e7c2ced5c2a: Layer already exists
eb8fae5482f5: Layer already exists
0df8c6eac500: Layer already exists
5f70bf18a086: Layer already exists
36e5ce8792b6: Layer already exists
denied: permission_denied: write_package
```
Nope

Kontakter (approved):

PR #842, opened 2024-09-11
yt-local: add option to choose node flavors

koct9i:


yt/python/yt/environment/api/__init__.py
```diff
@@ -106,6 +106,7 @@ class LocalYtConfig(object):
     scheduler_count = attr.ib(1)
     controller_agent_count = attr.ib(1)
     node_count = attr.ib(1)
+    node_flavors = attr.ib([])
```
Kontakter, line 109: This option looks twisted, I propose an alternative way.

Let add subconfig `YtNodeGroup` with attributes: `flavors` and `count`. And add an option to specify list of groups here. In the CLI group can be specified by yson for example.

Kontakter (changes requested):
Look my previous comments

PR #844, opened 2024-09-13
Improve/fix rowlet size estimation for operations archive

achulkov2:
Previously, TJobReport had a single `EstimateSize` method, which didn't account all its fields. This made rate limiting in archive reporter ineffective when dealing with large stderrs/fail contexts. This PR implements separate estimation methods for each job table, which will handle these issues correctly.

### DIFF
--- a/yt/yt/server/lib/misc/estimate_size_helpers-inl.h
+++ b/yt/yt/server/lib/misc/estimate_size_helpers-inl.h
@@ -8,10 +8,35 @@ namespace NYT {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+template <typename I>
+    requires std::integral<I>
+size_t EstimateSize(const I& /*value*/)
+{
+    return EstimatedValueSize;
+}
+
+template <typename F>
+    requires std::floating_point<F>
+size_t EstimateSize(const F& /*value*/)
+{
+    return EstimatedValueSize * 2;
+}
+
 template <typename T>
 size_t EstimateSize(const std::optional<T>& v)
 {
     return v ? EstimateSize(*v) : 0;
+}
+
+template <typename C>
+    requires std::ranges::range<C>
+size_t EstimateSize(const C& value)
+{
+    size_t result = EstimatedValueSize;
+    for (const auto& elem : value) {
+        result += EstimateSize(elem);
+    }
+    return result;
 }
 
 template <typename E>
--- a/yt/yt/server/lib/misc/estimate_size_helpers.cpp
+++ b/yt/yt/server/lib/misc/estimate_size_helpers.cpp
@@ -7,6 +7,11 @@ namespace NYT {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+size_t EstimateSize(const std::string& value)
+{
+    return EstimatedValueSize + value.size();
+}
+
 size_t EstimateSize(const TString& value)
 {
     return EstimatedValueSize + value.size();
@@ -30,6 +35,21 @@ size_t EstimateSize(TGuid value)
 size_t EstimateSize(TInstant /*value*/)
 {
     return EstimatedValueSize;
+}
+
+size_t EstimateSize(TDuration /*value*/)
+{
+    return EstimatedValueSize;
+}
+
+size_t EstimateSize(double /*value*/)
+{
+    return EstimatedValueSize * 2;
+}
+
+size_t EstimateSize(const ::google::protobuf::Message& value)
+{
+    return value.ByteSizeLong();
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/yt/yt/server/lib/misc/estimate_size_helpers.cpp
+++ b/yt/yt/server/lib/misc/estimate_size_helpers.cpp
@@ -22,11 +22,6 @@ size_t EstimateSize(const NYson::TYsonString& value)
     return value ? EstimatedValueSize + value.AsStringBuf().size() : 0;
 }
 
-size_t EstimateSize(i64 /*value*/)
-{
-    return EstimatedValueSize;
-}
-
 size_t EstimateSize(TGuid value)
 {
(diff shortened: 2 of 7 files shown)


yt/yt/server/lib/misc/job_reporter.cpp
```diff
@@ -74,7 +75,35 @@ class TJobRowlet
...
+            Report_.TaskName(),
+            Report_.TreeId(),
+            Report_.MonitoringDescriptor(),
+            Report_.Type(),
+            Report_.Error(),
+            Report_.InterruptionInfo(),
+            Report_.Statistics(),
+            Report_.Events(),
+            Report_.Spec().has_value(),
+            i64{0}, // FailContextSize.
```
eshcherbin, line 98: I'd suggest `/* failContextSize */ i64{0}`.

yt/yt/server/lib/misc/estimate_size_helpers.cpp
```diff
@@ -32,6 +37,21 @@ size_t EstimateSize(TInstant /*value*/)
...
 
+size_t EstimateSize(TDuration /*value*/)
+{
+    return EstimatedValueSize;
+}
+
+size_t EstimateSize(double /*value*/)
+{
+    return EstimatedValueSize * 2;
+}
```
eshcherbin, line 48: I don't believe we need a special instantiation for `double` and `i64` now.

yt/yt/server/lib/misc/job_reporter.cpp
achulkov2: Afaik we use `/*smth*/` without spaces now? Otherwise, sure.

Kontakter (approved):
To run tests in internal CI

yt/yt/server/lib/misc/job_reporter.cpp
```diff
@@ -326,7 +367,13 @@ class TJobProfileRowlet
...
     size_t EstimateSize() const override
     {
-        return Report_.EstimateSize();
+        return ::NYT::EstimateSizes(
+            Report_.OperationId().Underlying(),
+            Report_.JobId().Underlying(),
+            Report_.Profile().value_or(NJobAgent::TJobProfile{}).Type,
+            /*partIndex*/ int{0},
+            Report_.Profile().value_or(NJobAgent::TJobProfile{}).Blob,
+            Report_.Profile().value_or(NJobAgent::TJobProfile{}).ProfilingProbability);
```
omgronny, line 376: Maybe we can implement `EstimateSize` for `NJobAgent::TJobProfile`?

omgronny (changes requested):

yt/yt/server/lib/misc/job_reporter.cpp
achulkov2: When I was writing the code I thought `estimate_size_helpers` are in regular generic misc, but now I see that they are actually in `server/lib/misc`, same as the rest of the code. 

I do not have a strong opinion, I guess we can dedicate a section to specific types.

yt/yt/server/lib/misc/estimate_size_helpers-inl.h
```diff
@@ -8,12 +8,37 @@ namespace NYT {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+template <typename I>
+    requires std::integral<I>
+size_t EstimateSize(const I& /*value*/)
+{
+    return EstimatedValueSize;
+}
```
eshcherbin, line 16: Why do we estimate integers as 16 bytes instead of 8? Furthermore, why are floating points 32 bytes?

omgronny (approved):

yt/yt/server/lib/misc/job_reporter.cpp
```diff
@@ -74,7 +75,35 @@ class TJobRowlet
...
     size_t EstimateSize() const override
     {
-        return Report_.EstimateSize();
+        return ::NYT::EstimateSizes(
+            Report_.OperationId().Underlying(),
+            Report_.JobId().Underlying(),
+            Report_.State(),
+            Report_.StartTime(),
+            Report_.FinishTime(),
+            /*updateTime*/ TInstant::Now().MicroSeconds(),
```
eshcherbin, line 84: Why `MicroSeconds`?

yt/yt/server/lib/misc/job_reporter.cpp
eshcherbin: I believe we can define an `EstimateSize` specialization for `TJobProfile` next to the definition of `TJobProfile`?

yt/yt/server/lib/misc/job_reporter.cpp
achulkov2: For simplicity, same as value used for `.UpdateTime` below.

yt/yt/server/lib/misc/estimate_size_helpers-inl.h
achulkov2: Idk, this code is quite ancient: https://github.com/ytsaurus/ytsaurus/commit/a2a80a90bcec869fffd63b7d9a89bbc26e21d929. 
I just went with the spirit of things, since `TGuid` is also estimated as 32 bytes even though it is only 16 bytes in reality. 

Frankly, I do not like the way this is implemented at all. The best way to estimate the size of an unversioned row is to call `GetDataWeight` on the actual row. Is there any reason we cannot materialize the unversioned rows earlier and throw out all of this code? The rows will be materialized later on anyway. I guess it might be a bit cumbersome to ensure that rows we drop are deallocated at the right moment, but it should be doable.
In any case, that is obviously a more substantial refactoring that I do not want to pursue from a non code-owning point of view.

achulkov2: So in this case I am more for keeping coherence with the existing logic.

PR #845, opened 2024-09-13
Split attribute update and removal validation

gritukan:
Prior to this change, `ValidateCustomAttributeUpdate` was responsible both for the attribute update and removal validation. This is error-prone since having null `newValue` in the `Validate...Update` function may be unexpected. This issue happened in the `//sys/@cluster_name` and `//sys/@cluster_connection` attributes whose removal led to master crash prior to this change.

To address these issues, `ValidateCustomAttributeUpdate` is split here into `ValidateCustomAttributeUpdate` that is called only during attribute set or update and thus new value is always non-null and `ValidateCustomAttributeRemoval`. Also, `oldValue` is not passed to `ValidateCustomAttributeUpdate` after this change since stateful validation may be tricky and is not used for now.


maxim-babenko:
Shipped!

PR #849, opened 2024-09-16
Add cli params to docker respawn decorator

thenno:


yt/python/yt/wrapper/py_wrapper.py
```diff
@@ -803,6 +803,7 @@ def __init__(
             target_platform,
             docker,
             python,
+            cli_args=None,
```
Kontakter, line 806: Two remarks:
1. This default value does not seem intuitive, it looks like empty list is more natural (but I may be wrong, since I am not the user of this feature)
2. Using `None` for this non-trivial behaviour of taking `sys.argv[1:]` is not a good solution, it is better to use some other marker of flag here (for example make a separate boolean option `inherit_sys_argv`)

thenno: I based this decisions on use cases (which we encountered last week):
1. If the user does not specify `cli_args`, we should pass the original launch's arguments. `main_script_path` (==`argv[0]`) is determined separately, because most often user only needs to override it.
2. The user might want to pass a different arguments:
  1. A fixed set of arguments.
  2. Extend the existing set or filter current `argv[1:]` somehow (e.g., `[arg for arg in sys.argv[1:] if arg != "--strange-flag"` or `sys.argv[1:] + ["--magic-arg", "1"]`.

Using `None` as the default value instead of `[]` is just a good practice in Python to avoid accidentally modifying a mutable list inside a function.

I considered a flag like inherit_sys_argv, but there are two problems:
1. If only `inherit_sys_argv` is implemented, there is no simple way to override arguments in the user's script.
2. If both options are implemented, it's unclear how the code should behave with provided `cli_args` and `inherit_sys_argv` (fail, merge them, append `cli_args`, etc.).

I think with the current approach, 99% of users will simply use the defaults and get an equivalent restart in the docker container, while 1% will be able to easily override args as they prefer.

Kontakter: > Using None as the default value instead of [] is just a good practice in Python to avoid accidentally modifying a mutable list inside a function.

But None is not equivalent to `[]`, and it is not intuitive

thenno: > But None is not equivalent to [], and it is not intuitive
I mean the following
https://docs.python.org/3/reference/compound_stmts.html#function-definitions
paragraph starting from line "Default parameter values are evaluated from left to right when the function definition is executed".

In our case, the semantics of an empty `list` and `None` are different:
1. None is a default value (== sys.argv[1:])
2. `[]` is just an empty list (==forece set "no cli args")

We want to use `sys.argv[1:]` as the default value ==> by default all cli args should be passed to the docker container.

On the one hand we can write something like
```python
import sys

def foo(cli_args=sys.argv[1:]:
    pass
```

but on the other hand it sets `cli_args default` value in the import time

```
In [4]: def f(a=time.time()):
   ...:     print(a)
   ...:

In [5]: f()
1726505645.2767649

In [6]: f()
1726505645.2767649
```

In case the user changes `sys.argv` for some reason at runtime this solution won't work well. In my opinion default should be `None`.

Maybe this code

```python
if self._cli_args is None:
    self._cli_args = sys.argv[1:]
```

should be moved to `respawn_in_docker` for improving readability, I don't have a strong opinion here.

thenno: agreed to use an object `_INHERIT_SYS_ARGV = object()` as a default value.

Kontakter (approved):

PR #851, opened 2024-09-16
Make possible to ban yqla\qt

Krisha11:
This PR makes it possible to ban qt\yqla. This may be useful for rolling updates, as it will allow us to avoid running new queries on old instances.

YqlAgent in banned state stops accepting new queries, but continues to process old ones.
QueryTracker in banned state stops accepting incoming API calls and stops taking new queries from Active Queries table, while still processing old ones.


Krock21:
> YqlAgent in banned state stops accepting new queries, but continues to process old ones.

Our queries can be as long as 24 hours :( And we may increase this limit soon

> QueryTracker in banned state stops accepting incoming API calls

But proxies still try to send API requests to this QT? Do they fallback(how fast?) to other QTs or do they fail?

yt/yt/ytlib/yql_client/public.h
```diff
@@ -16,6 +16,7 @@ DEFINE_ENUM(EExecuteMode,
 
 YT_DEFINE_ERROR_ENUM(
     ((Throttled)            (40100))
+    ((YqlAgentBanned)       (40101))
```
Krock21, line 19: What http/rpc proxy does when it is banned? Does it have a special error code for it?
If yes, then it would be nice to have one for QT too

yt/python/yt/environment/components/query_tracker/__init__.py
```diff
@@ -63,11 +63,16 @@ def run(self):
             "stages": {
                 "production": {
                     "channel": {
+                        # NB(mpereskokova): Required for the bans mechanics;
+                        # In case of disabled balancing on single address discovery,
+                        # discovery requests are not sent and bans are not executed
+                        "disable_balancing_on_single_address": False,
```
Krock21, line 69: Should we do the same in ytsaurus-k8s-operator?

Krisha11: Yes, after finishing this PR I will do it

yt/yt/ytlib/yql_client/public.h
Krisha11: Yes it has its own error code, but we don't need it in qt. Maybe in YQL Agent it is redundant too. 

I have added an override for the IsUp function, which tells clients that if it returns false, this specific instance is unavailable. As a result, the client will stop seeing it and will stop sending requests to it.

I added this error to yqla because after the ban flag has been set to true, we have a short amount of time before the next isUp function call, during which we still accept new queries. These queries are too long, and we may want to stop accepting them as quickly as possible. In qt, we don't want this at all.

yt/yt/tests/integration/queries/test_mock.py
```diff
@@ -202,6 +205,44 @@ def test_rows_limit(self, query_tracker):
...
+    def test_query_tracker_ban(self, query_tracker):
+        addresses = query_tracker.query_tracker.addresses
+        set(f"//sys/query_tracker/instances/{addresses[0]}/@banned", True)
+
+        with raises_yt_error() as err:
+            wait(TestQueryTrackerBan.query_fails)
+        assert err[0].contains_text("No alive peers found")
+        time.sleep(3)  # NB(mpereskokova): Sleep for multiple active_query_acquisition_periods, to ensure, that no new queries will be acquired
+
+        guid = "eb334625-4a579e0c-b99d5757-78866429"
```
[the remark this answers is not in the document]
Krisha11: I just didn't find a way to create a correct guid in python - so I used a concrete one. OK, I will add a comment with this information

Krisha11:
> > YqlAgent in banned state stops accepting new queries, but continues to process old ones.
> 
> Our queries can be as long as 24 hours :( And we may increase this limit soon
> 
This scheme makes available only really long rolling updates, but it is better than nothing

Krisha11:
> > QueryTracker in banned state stops accepting incoming API calls
> 
> But proxies still try to send API requests to this QT? Do they fallback(how fast?) to other QTs or do they fail?

No, they don't. In some moment proxies will call IsUp function in qt - which tells, that it stops working - and stops sending queries to it. It takes seconds to make system understand that qt is banned

yt/yql/tests/test_simple.py
```diff
@@ -105,6 +105,58 @@ def test_calc(self, query_tracker, yql_agent):
...
+        write_table("//tmp/t", rows)
+
+        create_pool("small", attributes={"resource_limits": {"user_slots": 0}})
+        long_query = start_query("yql", 'pragma yt.StaticPool = "small"; select a+1 as result from primary.`//tmp/t`')
+        wait(lambda: long_query.get()["state"] == "running")
+
+        address = yql_agent.yql_agent.addresses[0]
+        set(f"//sys/yql_agent/instances/{address}/@banned", True)
+
+        with raises_yt_error():
```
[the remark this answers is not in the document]
Krisha11: I checked it here 
https://github.com/ytsaurus/ytsaurus/pull/851/files#diff-468cf55eb218c90102bd15a8ea9853a96d4a9cb89971480bcfad9c007e13a81eR132
And I thought, that it might be unnecessary in other place, but OK, I will add it

yt/yt/tests/integration/queries/test_mock.py
```diff
@@ -202,6 +205,44 @@ def test_rows_limit(self, query_tracker):
...
+    QUERY_TRACKER_DYNAMIC_CONFIG = {"state_check_period": 2000}
+
+    def query_fails():
+        start_query("mock", "run_forever")
+        return False
+
+    @authors("mpereskokova")
+    def test_query_tracker_ban(self, query_tracker):
+        addresses = query_tracker.query_tracker.addresses
+        set(f"//sys/query_tracker/instances/{addresses[0]}/@banned", True)
```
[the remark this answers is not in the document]
Krisha11: It is unnecessary, because the next wait can include this time in its 30s limit

yt/yt/server/lib/state_checker/state_checker.cpp
```diff
@@ -0,0 +1,74 @@
...
+{
+    StateCheckerExecutor_->Start();
+}
+
+void TStateChecker::SetPeriod(TDuration stateCheckPeriod)
+{
+    StateCheckerExecutor_->SetPeriod(stateCheckPeriod);
+}
+
+bool TStateChecker::IsComponentBanned()
```
psushin, line 39: const

yt/yt/server/lib/state_checker/state_checker.cpp
```diff
@@ -0,0 +1,74 @@
...
+    });
+
+    auto options = NApi::TGetNodeOptions{
+        .Attributes = NYTree::TAttributeFilter({NApi::BannedAttributeName}),
+    };
+
+    try {
+        auto yson = WaitFor(NativeClient_->GetNode(InstancePath_, options))
+            .ValueOrThrow();
+        auto instance = ConvertTo<NYTree::INodePtr>(yson);
```
psushin, line 62: almost no use of namespaces in cpp, instead move to using namespace (here, and in other places)

yt/yt/server/lib/state_checker/state_checker.cpp
```diff
@@ -0,0 +1,74 @@
...
+    return Banned_.load();
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+void TStateChecker::DoCheckState()
+{
+    VERIFY_INVOKER_AFFINITY(Invoker_);
+
+    YT_LOG_DEBUG("Checking component state started");
```
psushin, line 50: Started checking / Finished checking

yt/yt/server/lib/state_checker/state_checker.h
```diff
@@ -0,0 +1,38 @@
...
+
+namespace NYT::NStateChecker {
+
+////////////////////////////////////////////////////////////////////////////////
+
+class TStateChecker
+    : public TRefCounted
+{
+public:
+    TStateChecker() = delete;
```
psushin, line 13: we have a "normal" ctor, do we need this explicit delete?

yt/yql/agent/yql_service.cpp
```diff
@@ -49,6 +55,11 @@ class TYqlService
         context->SetRequestInfo("QueryId: %v, Async: %v, BuildRowsets: %v, RowCountLimit: %v", queryId, request->async(), request->build_rowsets(), request->row_count_limit());
         context->SetResponseInfo("QueryId: %v", queryId);
 
+        if (StateChecker_->IsComponentBanned()) {
+            YT_LOG_INFO("Yql agent was banned, failing query (QueryId: %v, User: %v)", queryId, impersonationUser);
```
psushin, line 59: is banned

yt/yql/agent/yql_service.cpp
```diff
@@ -49,6 +55,11 @@ class TYqlService
...
+            THROW_ERROR_EXCEPTION(NYqlClient::EErrorCode::YqlAgentBanned, "Yql agent was banned");
```
psushin, line 60: is banned

yt/yql/tests/test_simple.py
```diff
@@ -105,6 +105,58 @@ def test_calc(self, query_tracker, yql_agent):
         """, [[{"column0": 1}], [{"column0": 2}], [{"a": 1, "b": 2}]])
 
 
+class TestYqlAgentBan(TestQueriesYqlBase):
+    NUM_YQL_AGENTS = 1
+    YQL_AGENT_DYNAMIC_CONFIG = {"state_check_period": 2000}
+
+    def query_fails():
```
psushin, line 112: a bit weird name for a method, looks like a verb is missing

yt/yql/tests/test_simple.py
```diff
@@ -105,6 +105,58 @@ def test_calc(self, query_tracker, yql_agent):
...
+        address = yql_agent.yql_agent.addresses[0]
+        set(f"//sys/yql_agent/instances/{address}/@banned", True)
+
+        with raises_yt_error() as err:
+            wait(TestYqlAgentBan.query_fails)
+        assert err[0].contains_text("No alive peers found")
+
+        set(f"//sys/yql_agent/instances/{address}/@banned", False)
+
+        wait(TestYqlAgentBan.query_completes)
```
psushin, line 136: is it intentional, that we don't check result value here (whether it failed or not)?

yt/yt/server/lib/state_checker/state_checker.cpp
psushin, line 62: ConvertToNode

yt/yt/server/lib/state_checker/state_checker.h
```diff
@@ -0,0 +1,38 @@
...
+    TStateChecker(IInvokerPtr invoker, NApi::IClientPtr nativeClient, TString instancePath, TDuration stateCheckPeriod);
+
+    void Start();
+    void SetPeriod(TDuration stateCheckPeriod);
+
+    bool IsComponentBanned();
```
psushin, line 19: const

yt/yt/server/lib/state_checker/state_checker.cpp
```diff
@@ -0,0 +1,74 @@
...
+    auto options = NApi::TGetNodeOptions{
+        .Attributes = NYTree::TAttributeFilter({NApi::BannedAttributeName}),
+    };
+
+    try {
+        auto yson = WaitFor(NativeClient_->GetNode(InstancePath_, options))
+            .ValueOrThrow();
+        auto instance = ConvertTo<NYTree::INodePtr>(yson);
+
+        auto banned = instance->Attributes().Get<bool>(NApi::BannedAttributeName);
```
psushin, line 64: What if we don't have an attribute (since it's a user attribute, this is quite possible)?
Should we have a default fallback?

yt/yt/server/lib/state_checker/state_checker.h
```diff
@@ -0,0 +1,38 @@
...
+namespace NYT::NStateChecker {
+
+////////////////////////////////////////////////////////////////////////////////
+
+class TStateChecker
+    : public TRefCounted
+{
+public:
+    TStateChecker() = delete;
+    TStateChecker(IInvokerPtr invoker, NApi::IClientPtr nativeClient, TString instancePath, TDuration stateCheckPeriod);
```
psushin, line 14: TString -> TYPath?

yt/yt/tests/integration/queries/test_mock.py
```diff
@@ -202,6 +205,44 @@ def test_rows_limit(self, query_tracker):
...
+
+        with raises_yt_error() as err:
+            wait(TestQueryTrackerBan.query_fails)
+        assert err[0].contains_text("No alive peers found")
+        time.sleep(3)  # NB(mpereskokova): Sleep for multiple active_query_acquisition_periods, to ensure, that no new queries will be acquired
```
psushin, line 227: I strongly object writing tests with sleep.
Let's export StateChecker state via orchid, and wait until the banned flag is updated from the cypress.

psushin (changes requested):

yt/yt/tests/integration/queries/test_mock.py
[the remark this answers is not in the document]
psushin: look for generate_uuid() in test_portals.py for example

yt/yql/tests/test_simple.py
Krisha11: I used "fail" as a verb, like "complete" in the next one

yt/yql/tests/test_simple.py
Krisha11: In case of failure, it throws an error during testing - so if it completes successfully, I can be sure that there are no errors.

yt/yql/tests/test_simple.py
psushin: next one also looks like a place for improvement)

method names usually start with verb

I would use names like  
 * _test_query_fails / _test_query_completes (with "test" like in test_and_set, not a separate integration tets)
 * check_query_fails / check_query_completes
 * run_query_fail_check / run_query_complete_check

yt/yql/tests/test_simple.py
psushin: now I got it!

yt/yt/server/lib/state_checker/state_checker.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+        auto previousBanned = Banned_.exchange(banned);
+        if (previousBanned != banned) {
+            auto orchidBannedPath = InstancePath_ + "/orchid/@banned";
+
+            YT_LOG_DEBUG("Updating orchid node (Path: %v, Banned: %v)", orchidBannedPath, banned);
+            WaitFor(NativeClient_->SetNode(orchidBannedPath, ConvertToYsonString(banned)))
+                .ThrowOnError();
+            YT_LOG_DEBUG("Orchid node updated");
+        }
```
Krock21, line 79: Why do we need to do it? Seems like `orchid` should belong to a node and should be managed by a node.
Also we already have this attribute on `instance/@banned`, I don't think we need an additional one on `instance/orchid/@banned`

Krisha11: It was made to reduce sleeping in tests.
https://github.com/ytsaurus/ytsaurus/pull/851#discussion_r1763027352

This attribute is different. It means that component already knows about ban and won't continue to work.

Krock21: I see, it is used only in yql agent and query tracker itself, so they are still responsible for their `orchid`

You said in some reply:
> In some moment proxies will call IsUp function in qt - which tells, that it stops working - and stops sending queries to it.

But isUp exists only in StateChecker which should exist only in yql agent and query tracker
Also I don't see any IsUp usages in the PR. Could you please help me find them?

Krisha11: It is a common RPC mechanic in Discover requests
https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/core/rpc/service_detail.cpp#L2735

Krock21: I see. My misunderstanding was here:
> isUp exists only in StateChecker

isUp is a separate API that forwards the request to StateChecker

Thanks

Krisha11:
Shipped!

yt/yt/server/lib/state_checker/public.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <library/cpp/yt/memory/ref_counted.h>
```
psushin, line 3: you can include core/misc/common.h instead

yt/yt/server/lib/state_checker/state_checker.cpp
```diff
@@ -0,0 +1,88 @@
...
+TStateChecker::TStateChecker(IInvokerPtr invoker, IClientPtr nativeClient, TYPath instancePath, TDuration stateCheckPeriod)
+    : Logger(StateCheckerLogger())
+    , Invoker_(std::move(invoker))
+    , NativeClient_(std::move(nativeClient))
+    , InstancePath_(std::move(instancePath))
+{
+    Banned_.store(false);
+    StateCheckerExecutor_ = New<TPeriodicExecutor>(
+        Invoker_,
+        BIND(&TStateChecker::DoCheckState, MakeStrong(this)),
```
psushin, line 27: Perhaps MakeWeak is enough and helps to prevent reference cycle.

yt/yt/server/lib/state_checker/state_checker.cpp
```diff
@@ -0,0 +1,88 @@
...
+    };
+
+    try {
+        auto yson = WaitFor(NativeClient_->GetNode(InstancePath_, options))
+            .ValueOrThrow();
+        auto instance = ConvertToNode(yson);
+
+        auto banned = false;
+        if (instance->Attributes().Contains(BannedAttributeName)) {
+            banned = instance->Attributes().Get<bool>(BannedAttributeName);
```
psushin, line 68: Alternatively you could write it with Get method (with a default value).

yt/yt/server/lib/state_checker/state_checker.cpp
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        auto previousBanned = Banned_.exchange(banned);
+        if (previousBanned != banned) {
+            auto orchidBannedPath = InstancePath_ + "/orchid/@banned";
```
psushin, line 73: Orchid usually works as a virtual YTree for diagnostics.
What you've done here with direct sets into cypress is really confusing.

You can create a pluggable orchid service in StateChecker, that will be incorporated into ochid on yql agent and QT.

Look for example at methods GetOrchidService.

yt/yt/tests/integration/queries/test_mock.py
```diff
@@ -202,6 +205,44 @@ def test_rows_limit(self, query_tracker):
...
+        guid = generate_uuid()
+        insert_rows("//sys/query_tracker/active_queries", [{
+            "query_id": guid,
+            "engine": "mock",
+            "query": "run_forever",
+            "incarnation": 0,
+            "state": "pending",
+            "settings": {}
+        }])
+        time.sleep(3)  # NB(mpereskokova): Sleep for multiple active_query_acquisition_periods, to ensure, that inserted query won't be launched
```
psushin, line 238: Even though this is fine with me, we could expose the counter of query acquisition iterations into the orchid and wait for a required number of iterations to pass.

psushin (changes requested):

Krisha11:
Shipped!

Krisha11:
Shipped!

PR #858, opened 2024-09-20
Fix building odin in opensource

achulkov2:
This PR makes various improvements around the open source odin build stack:
* Fixes `yt/python/packages/setup.py` to work correctly with python 3.12+.
* Fixes `yt/python/packages/setup.py` to declare its own dependencies.
* Strips ytsaurus client requirements from odin: they take a while to install and are unnecessary. We can add them singularly as need be.
* Implements odin build as part of the shared ytsaurus multi-stage dockerfile, which required some refactoring.
* Switches to yt python libraries from pip for the new odin dockerfile: incompatibilities here are rare, we can implement a source-based version later.

The main reason for this change is that the current Dockerfile in `yt/odin` doesn't actually work, especially after the clang 14 -> 16 change in the toolchain. We spent quite some time trying to get past through these problems and ended up building from an older version of the repository (with the older toolchain).

I feel like it is best to scrap the current dockerfile and move towards sharing more layers with the existing images. I do not think it is necessary to build python libraries from source for odin checks, so I did not implement it. But it is possible to improve the configuration so that this behaviour is configurable.

Big thanks to @thenno for helping with setup.py debug!


yt/docker/ytsaurus/Dockerfile
```diff
@@ -27,16 +55,18 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt-get install
...
   telnet \
   tini \
   unzip \
-  zstd
+  zstd \
+  google-perftools \
+  procps \
+  jq \
+  htop \
+  && rm -rf /var/lib/apt/lists/*
```
Kontakter, line 69: It seems that the proper way is to call command `apt-cache clean` 

Also I suppose to split visually separate commands:
```
RUN apt-get update && \
    DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt-get install -y \
        curl \
        dnsutils \
    ....
        htop && \
    apt-cache clean
```

The exact syntax is discussable

yt/docker/ytsaurus/Dockerfile
```diff
@@ -228,3 +258,75 @@ FROM scratch AS ytsaurus-server-override
...
+ARG ODIN_CHECKS_DIR
+ARG ODIN_CHECKS_DATA_DIR
+
+ARG SOURCE_ROOT="${ODIN_ROOT}/ytsaurus"
+ARG ODIN_SOURCE_ROOT="${SOURCE_ROOT}/yt/odin"
+ARG BUILD_ROOT="${ODIN_ROOT}/build"
+ARG PYTHON_ROOT="${ODIN_ROOT}/python"
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
```
Kontakter, line 275: Why `DEBIAN_FRONTEND=noninteractive` is omitted here?

yt/docker/ytsaurus/Dockerfile
```diff
@@ -228,3 +258,75 @@ FROM scratch AS ytsaurus-server-override
...
+    && mkdir ${PYTHON_ROOT} \
+    && cd ${SOURCE_ROOT} && pip install -e yt/python/packages \
+    && cd ${PYTHON_ROOT} \
+    && python ${ODIN_SOURCE_ROOT}/packages/prepare_python_modules.py --source-root "$SOURCE_ROOT" --output-path "$PYTHON_ROOT" \
+    && cp "${ODIN_SOURCE_ROOT}/packages/setup.py" ./ && python setup.py bdist_wheel --universal
+
+RUN . ${ODIN_VIRTUAL_ENV}/bin/activate \
+    && pip install -r ${ODIN_SOURCE_ROOT}/requirements.txt \
+    && pip install ytsaurus-client ytsaurus-yson ytsaurus-rpc-driver \
+  && pip install $PYTHON_ROOT/dist/*
```
Kontakter, line 299: Wrong indentation

yt/docker/ytsaurus/Dockerfile
```diff
@@ -228,3 +258,75 @@ FROM scratch AS ytsaurus-server-override
...
+
+RUN mkdir -p ${ODIN_CHECKS_DIR} \
+    && mkdir -p ${ODIN_CHECKS_DATA_DIR} \
+    && for CHECK in $(find ${ODIN_SOURCE_ROOT}/checks/bin -mindepth 1 -maxdepth 1 -type d -not -name "ya.make"); do \
```
Kontakter, line 303: Maybe use here `;` instead of `&&` ?

Kontakter: Why `-not -name "ya.make"` needed? It is not a directory already

yt/odin/requirements.txt
```diff
@@ -3,8 +3,6 @@ Flask==2.0.3
 Werkzeug==2.2.2
 cheroot==10.0.0
 
-# Required by some of the checks
+# Required by some of the checks.
 requests==2.31.0
 tabulate==0.9.0
-
--r ../yt/scripts/python_libraries_requirements.txt
```
Kontakter, line 10: Why is it removed?

achulkov2: Some time ago we were trying to build some working newer build of odin and these requirements had weird compatibility issues. It was not me building it, so I do not know the exact details.

In this new setup, these requirements do not seem to case any trouble (afair I built with them at some point), but frankly I do not see why we need all of them. They take quite a while to install, clog up the image, etc.

Let's put it this way: either a library is necessary for yt client to work properly, then it should be a dependency of the library itself, so that it is auto-installed by pip. If it is somehow optional, let's list such libraries actually needed for odin checks to work. Do we really need numpy, pandas and confluent-kafka for odin checks? 

@savnadya says that a lot of these libraries are required by YT tests, which IMO should be separate from odin

yt/docker/ytsaurus/Dockerfile
achulkov2: I think && is better, so we do not rely on the fact that the last statement should be erroneous in case any of the first mkdirs fail (otherwise docker would swallow the error).

As for `-not -name "ya.make"` I think it is not neccessary, removed it.

yt/docker/ytsaurus/Dockerfile
achulkov2: No good reason. I was copy and pasting from old Odin dockerfile. Added it to all apt-get install commands in this file.

yt/docker/ytsaurus/Dockerfile
achulkov2: Ok, I will try it.

achulkov2: I did some googling, and I can't find the `apt-cache clean` command, but there is `apt-get clean`. However, it [does not](https://askubuntu.com/questions/1370546/what-is-the-difference-between-these-package-cache-clearing-methods) clean apt/lists, but it cleans caches. So it seems best to do both.

Kontakter:
Sorry for the delay, I will return to this PR in the beginning of the next week

yt/docker/ytsaurus/Dockerfile
```diff
@@ -4,39 +4,75 @@
 ARG BASE_REPOSITORY="ghcr.io/ytsaurus/ytsaurus-nightly"
 ARG BASE_IMAGE="latest"
 
+# Args for odin.
+ARG PYTHON_RUNTIME_VERSION="3.12"
```
Kontakter, line 8: It looks inconsistent with python3.11 below

yt/docker/ytsaurus/Dockerfile
```diff
@@ -4,39 +4,75 @@
...
+
+##########################################################################################
+
+FROM mirror.gcr.io/ubuntu:focal AS ubuntu-base
+
+RUN apt-get update \
+    && DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt-get install -y \
+      python3-pip \
+      python3.8  \
+      python3.8-distutils \
```
Kontakter, line 28: Why we install python3.8 here?

yt/docker/ytsaurus/Dockerfile
```diff
@@ -238,3 +274,75 @@ FROM scratch AS ytsaurus-server-override
 COPY --from=ytsaurus-server-override-bloated / /
 
 ##########################################################################################
+
+FROM python-base as odin-installer
```
Kontakter, line 278: What the idea to include odin docker to common ytsaurus Dockerfile?

yt/docker/ytsaurus/Dockerfile
achulkov2: I had some issues with python3.11, something was incompatible, I will try again and mention them here.

yt/docker/ytsaurus/Dockerfile
achulkov2: Isn't this our "default" system python that we have had for a while now?

yt/docker/ytsaurus/Dockerfile
achulkov2: Ok, I tried 3.11 and it seems to build fine now. Will try it on a real installation later, but for now let's use 3.11.

savnadya:
Shipped!

PR #859, opened 2024-09-20
Fallback to generic cpu for llvm codegen when running in weird environments

achulkov2:
This PR addresses some, but not all of the issues described in #4. When running ytsaurus-local in docker desktop, the process would fail with `LLVM ERROR: 64-bit code requested on a subtarget that doesn't support it!` whenever the first dynamic table was mounted. It is due to the fact that in such an environment the cpu derived by llvm is `athlon-xp`, which is actually 32-bit, while the request target triple is provided at compile time and is obviously x86_64. This is "fixed" by falling back on cpu = "generic" when the detected cpu is incompatible with the target triple.

There are still some remaining issues when running with orbstack, but the failures happen in random components and the traces are very different, so my current stance is that they are unrelated.

### DIFF
--- a/yt/yt/library/codegen/module.cpp
+++ b/yt/yt/library/codegen/module.cpp
@@ -30,6 +30,7 @@
 #include <llvm/Support/raw_ostream.h>
 #include <llvm/Support/SmallVectorMemoryBuffer.h>
 #include <llvm/CodeGen/Passes.h>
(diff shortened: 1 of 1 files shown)


AlexeyLukyanchikov (approved):

Kontakter (approved):
To run tests in internal CI

PR #862, opened 2024-09-24
Fix clang logical-op-parentheses warning

abogutskiy:


yt/yt/core/concurrency/periodic_executor_base.h
```diff
@@ -15,12 +15,12 @@ namespace NYT::NConcurrency::NDetail {
 //! and have a function to process given result.
 template <class T>
 concept CCallbackResultProcessor =
-    requires {
+    (requires {
```
E1pp, line 18: This ordering is wrong: we want to always make sure that given alias is present. And then we check that we either have `void` alias and no argument callable method or non-`void` alias and callable member method with an argument of the aliased type.

Before it wasn't an actual bug because the first constraint is always implied in every other constraint.

Please move parentheses so that `||` takes place first

E1pp (changes requested):

PR #865, opened 2024-09-27
Export odin check statuses in prometheus format

dmi-feo:


yt/odin/bin/yt_odin_webservice/__main__.py
```diff
@@ -151,6 +167,31 @@ def solomon_duration():
...
+    stop_timestamp = int(calendar.timegm(now.utctimetuple()))
+    db_records = extract_all_records(start_timestamp, stop_timestamp)
+
+    last_records_per_service = {}
+    for record in db_records:
+        if (
+                record.service not in last_records_per_service
+                or last_records_per_service[record.service].timestamp < record.timestamp
+        ):
+            last_records_per_service[record.service] = record
```
dmi-feo, line 184: I believe it's better to do this grouping on python backend side considering the small amount of data here.

yt/odin/bin/yt_odin_webservice/__main__.py
```diff
@@ -135,6 +137,20 @@ def create_solomon_timed_out_sensors(records):
     return sensors
 
 
+def get_prometheus_metric_line(
+        metric_name: str, metric_value: float, timestamp: Optional[int], labels: Optional[Dict[str, str]] = None
```
dim-an, line 141: `int | None`? `dict[str,str] | None` ?

yt/odin/bin/yt_odin_webservice/__main__.py
```diff
@@ -151,6 +167,31 @@ def solomon_duration():
     return jsonify({"sensors": duration_sensors + timed_out_sensors})
 
 
+@app.route("/prom")
```
dim-an, line 170: Is it default path for `prometheus` ? I wouldn't use abbreviation here.

yt/odin/bin/yt_odin_webservice/__main__.py
```diff
@@ -151,6 +167,31 @@ def solomon_duration():
...
+def prometheus():
+    now = datetime.datetime.now(datetime.timezone.utc)
+    period = datetime.timedelta(seconds=int(request.args.get("period", 60)))
+    start_timestamp = int(calendar.timegm((now - period).utctimetuple()))
+    stop_timestamp = int(calendar.timegm(now.utctimetuple()))
```
dim-an, line 175: Can't we use `datetime.timestamp()` here ?

yt/odin/bin/yt_odin_webservice/__main__.py
dmi-feo: Would be nice. What is the minimal python version that should be supported by odin?

yt/odin/bin/yt_odin_webservice/__main__.py
dmi-feo: Default path is `/metrics`, but it can be easily changed in scraper config. Let's use `/prometheus` then? I think it would be not bad to underline the fact this handler is for prometheus

yt/odin/bin/yt_odin_webservice/__main__.py
```diff
@@ -151,6 +166,31 @@ def solomon_duration():
     return jsonify({"sensors": duration_sensors + timed_out_sensors})
 
 
+@app.route("/prometheus")
+def prometheus():
+    now = datetime.datetime.now(datetime.timezone.utc)
+    period = datetime.timedelta(seconds=int(request.args.get("period", 90)))
```
achulkov2, line 172: If we are implementing this ourselves, can we put some link to docs in the comments? I actually do not know which HTTP request args prometheus handlers are supposed to support. For example, is there a parameter for "now"?

achulkov2 (changes requested):

yt/odin/bin/yt_odin_webservice/__main__.py
```diff
@@ -135,6 +136,20 @@ def create_solomon_timed_out_sensors(records):
     return sensors
 
 
+def get_prometheus_metric_line(
```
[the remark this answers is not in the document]
dmi-feo: Of course there is an official library: https://prometheus.github.io/client_python/getting-started/three-step-demo/

It is designed to collect and expose application metrics, our case (when you want to expose some metrics that have been already gathered not via the client and you don't want to use built-in magic flask integration) probably can be supported by the client, but if we want to keep it simple and understandable (taking into account how simple prometheus format is) I'd say it's better to implement the formatting function here.

yt/odin/bin/yt_odin_webservice/__main__.py
dmi-feo: It's not that hard to google these docs, but okay, mijn vriend, I'll do that!)
Prometheus doesn't use any params itself, but you can specify `metrics_path` and `params` in the scaper config, so it can look like:
```
metrics_path: "/prometheus"
    params:
      period: ["120"]
```

yt/odin/bin/yt_odin_webservice/__main__.py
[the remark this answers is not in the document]
zlobober: And taking into account we might need to get it under yamake/arcadia...
I agree with Dima that it is justified to do it manually as it takes less than 10 lines of code.
(and btw, similar private integration with Yandex-internal Juggler is also made in form of a manual HTTP request, and not a PEERDIR to their SDK library)

yt/odin/bin/yt_odin_webservice/__main__.py
achulkov2: Ok! While I am being picky, let's also add a comment about what `period` means logically, since it is basically an external configuration option. One should not have to read the code to understand what value should be provided for it.

yt/odin/bin/yt_odin_webservice/__main__.py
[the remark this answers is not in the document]
achulkov2: Ok.

yt/odin/bin/yt_odin_webservice/__main__.py
dim-an: I think they are not specified. Internally we use 3.12 and don't have tests for other versions.

I think we can use these type hints.

achulkov2 (approved):

dim-an:
Shipped!

PR #870, opened 2024-09-30
Fix bug in http proxy liveness during read-only introduced after refactoring

achulkov2:
Fun fact: if all orchids return error, we will consider masters to be in read-only and keep the proxies alive for longer. I don't really see too much harm in that.


don-dron (approved):
LGTM

PR #871, opened 2024-10-01
Do not fail on completed job stderr get if node if offline

gritukan:
Prior to this change, `get_job_stderr` would fail for a completed job if the corresponding exec node is offline because we try to get job stderr from the node first.


Kontakter (approved):
LGTM

PR #882, opened 2024-10-08
Allow to filter speclets in list handler

dmi-feo:


Kontakter:
Dmitry is on vacation, I hope that he will read this PR in the beginning of the next week

yt/chyt/controller/internal/api/helpers.go
```diff
@@ -92,6 +93,55 @@ func transformToStringSlice(value any) (any, error) {
...
+	filterVals, ok := value.(map[string]any)
+	if !ok {
+		typeName := reflect.TypeOf(value).String()
+		return nil, unexpectedTypeError(typeName)
+	}
+
+	allowedTypes := []string{"string", "bool"}
+	newMap := make(map[string]any, len(filterVals))
+	for k, v := range filterVals {
+		if k == "state" {
```
DimasKovas, line 110:
```go
if k == "state" {
    newMap[k] = strawberry.OpletState(v)
} else if k == "health" {
    newMap[k] = strawberry.OpletHealth(v)
} else {
    newMap[k] = v
}
```

DimasKovas (changes requested):

yt/chyt/controller/internal/api/helpers.go
```diff
@@ -92,6 +93,30 @@ func transformToStringSlice(value any) (any, error) {
 	return transformedAttributes, nil
 }
 
+func transformFilterValues(value any) (any, error) {
+	if value == nil {
+		return nil, nil
```
DimasKovas, line 98: `return nil` and `return map[string]any(nil)` are different in this context - the first one doesn't store the type info and `x.(map[string]any)` would cause panic (I'll fix it during the import)

yt/chyt/controller/internal/api/helpers.go
```diff
@@ -92,6 +93,30 @@ func transformToStringSlice(value any) (any, error) {
...
+	filterVals, ok := value.(map[string]any)
+	if !ok {
+		typeName := reflect.TypeOf(value).String()
+		return nil, unexpectedTypeError(typeName)
+	}
+
+	newMap := make(map[string]any, len(filterVals))
+	for k, v := range filterVals {
+		if k == "state" {
+			newMap[k] = strawberry.OpletState(v.(string))
```
DimasKovas, line 110: Need to validate the type of the `v` parameter before accessing it with `.(string)`
I'll fix it during the import

DimasKovas (approved):

PR #888, opened 2024-10-14
Require native client only in case of enabled auth

dmi-feo:


Kontakter (approved):

Kontakter:
Discard in favor of proper fix by Nadya Savchenko

PR #892, opened 2024-10-17
Import ydb/library/yql/public/result_format from Arcadia

Tony-Romanov:
For get rid of SKIFF in favor of yson in YQL-YT interop.

### DIFF
--- a/yt/yql/agent/ya.make
+++ b/yt/yql/agent/ya.make
@@ -45,6 +45,8 @@ PEERDIR(
     yt/yt/library/query/row_comparer
 
     yt/yql/plugin/bridge
+
+    ydb/library/yql/public/result_format
 )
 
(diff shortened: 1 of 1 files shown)


Kontakter (approved):
Just for internal tests.

Please do not commit before we discuss the necessity of this new dependency on ydb/

Kontakter:
We are waiting sync of this commit  https://github.com/ydb-platform/ydb/commit/9865b795dc9b9df3b001ef9c20696f598df3fe00 to Arcadia, I think it will be synced in about 1 week from now

Kontakter:
Unfortunately above commit isn't synced yet :(

Kontakter:
We implement it internally with updated location of YQL in Arcadia

PR #894, opened 2024-10-17
Couple fixes for clangd language server

koct9i:
- yt/client: include async_stream.h in file_reader.h
  NApi::IFileReader needs NConcurrency::IAsyncZeroCopyInputStream.
  
- Declare yt/yt/client/api/rpc_proxy/protocol_version_variables.h using CONFIGURE_FILE
  Otherwise resulting file does not materialize after:
  ya make --build-all --force-build-depends --replace-result --add-result=.h yt/yt/client/
  
  Proper source file generation is required at least for making clangd happy.

### DIFF
--- a/yt/yt/client/ya.make
+++ b/yt/yt/client/ya.make
@@ -198,6 +198,11 @@ SRCS(
     kafka/requests.cpp
 )
 
+CONFIGURE_FILE(
+    api/rpc_proxy/protocol_version_variables.h.in
+    api/rpc_proxy/protocol_version_variables.h
+)
+
 SRCS(
     ${YT_SRCS}
     yt/yt/client/api/rpc_proxy/protocol_version_variables.h.in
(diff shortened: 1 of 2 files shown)


yt/yt/client/ya.make
```diff
@@ -198,6 +198,11 @@ SRCS(
     kafka/requests.cpp
 )
 
+CONFIGURE_FILE(
```
Kontakter, line 201: It seams to be deducted by ya.make from SRCS below. Why do you think that this `CONFIGURE_FILE` is needed?

koct9i: As I mentioned - otherwise resulting file does not materialize after:
ya make --build-all --force-build-depends --replace-result --add-result=.h yt/yt/client/

Kontakter (approved):
To run build in internal CI

PR #897, opened 2024-10-17
Fix error message

gritukan:
`No such "tablet" cell bundle "default"` -> `No such tablet cell bundle "default"`

### DIFF
--- a/yt/yt/server/master/cell_server/tamed_cell_manager.cpp
+++ b/yt/yt/server/master/cell_server/tamed_cell_manager.cpp
@@ -1093,7 +1093,7 @@ public:
         if (!cellBundle) {
             THROW_ERROR_EXCEPTION(
                 NYTree::EErrorCode::ResolveError,
-                "No such %Qlv cell bundle %Qlv",
+                "No such %lv cell bundle %Qlv",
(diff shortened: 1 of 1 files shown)


Kontakter (approved):

yt/yt/server/master/cell_server/tamed_cell_manager.cpp
```diff
@@ -1093,7 +1093,7 @@ class TTamedCellManager
         if (!cellBundle) {
             THROW_ERROR_EXCEPTION(
                 NYTree::EErrorCode::ResolveError,
-                "No such %Qlv cell bundle %Qlv",
+                "No such %lv cell bundle %Qlv",
```
maxim-babenko, line 1096: Why Qlv for name?

gritukan: I think there is no reason, it should be %Qv. WIll fix in a separate PR.

gritukan: https://github.com/ytsaurus/ytsaurus/pull/937

PR #901, opened 2024-10-21
[yt/python] Include prefix "YT_" for names of SDK environment variables

koct9i:
Make grep-friendly mentions of all known environment variables.

Most lines of patch are generated by this:
sed -iE 's/    "\([A-Z0-9_]\+\)"/    "YT_\1"/' yt/python/yt/wrapper/default_config.py

### DIFF
--- a/yt/python/yt/wrapper/default_config.py
+++ b/yt/python/yt/wrapper/default_config.py
@@ -720,87 +720,87 @@ def get_default_config():
 
 
 FORCED_SHORTCUTS = {
-    "BASE_LAYER" : "operation_base_layer",
+    "YT_BASE_LAYER" : "operation_base_layer",
 }
 
 
 SHORTCUTS = {
-    "PROXY": "proxy/url",
-    "PROXY_SUFFIX": "proxy/default_suffix",
-    "PROXY_URL_ALIASING_CONFIG": "proxy/aliases",
-    "TOKEN": "token",
-    "TOKEN_PATH": "token_path",
-    "USE_TOKEN": "enable_token",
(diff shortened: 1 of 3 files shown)


yt/python/yt/wrapper/default_config.py
```diff
@@ -720,87 +720,87 @@ def get_default_config():
 
 
 FORCED_SHORTCUTS = {
-    "BASE_LAYER" : "operation_base_layer",
+    "YT_YT_BASE_LAYER" : "operation_base_layer",
```
gritukan, line 723: YT_YT?

yt/python/yt/wrapper/default_config.py
```diff
@@ -865,7 +865,6 @@ def _get(d, key):
         if not key.startswith(prefix):
```
gritukan, line 865: Do we need this code after the change?

Kontakter:
Please, read the code carefully, there are some other places that should be accurately fixed after this change: at least this one https://github.com/ytsaurus/ytsaurus/blob/main/yt/python/yt/wrapper/default_config.py#L988

Kontakter (changes requested):
See my and @gritukan  comments

yt/python/yt/wrapper/default_config.py
koct9i: Looks like good sanity-check and optimization - all must start with "YT_"

koct9i:
done

koct9i:
found one more place with obscure environment references

Kontakter (approved):
To run test in internal CI

PR #908, opened 2024-10-25
Use cell leader address as sanitized host

achulkov2:
This PR addresses multiple issues:
* Host sanitization was dependent on cell peer hostnames having the same length. The fallback was the local host name, which is different for different peers and can lead to snapshot mismatch.
* The sanitized host attribute is only **serialized** when running under error sanitizer guard. However, response errors aren't serialized in `Automaton`, so right now all such errors are returned to the user without the host attribute. Conversely, PRK serialization is performed in `Automaton`, so we were storing the potentially differing host names :)

The new approach uses the address of the leading peer in the cell as the sanitized host name, and a fixed placeholder if it is not possible to determine it for some reason. This way we don't need additional options for cellar nodes, since the leader is handling all requests and sanitization is not poisoning any errors.

Additionally, the host attribute is always serialized into errors, if available, same as datetime.

Also deletes `SanitizeLocalHostName` helper and corresponding unittest as obsolete.


maxim-babenko:
Shipped!

Kontakter:
Configuration phase tells: `cannot find source file: hostname_sanitization_ut.cpp`

achulkov2:
> Configuration phase tells: cannot find source file: hostname_sanitization_ut.cpp

Forgot to remove from ya.make, done.

Kontakter (approved):

maxim-babenko:
Shipped!

maxim-babenko:
Shipped!

maxim-babenko:
Shipped!

yt/yt/server/lib/hydra/unittests/hostname_sanitization_ut.cpp
```diff
@@ -57,7 +56,95 @@ TEST(TLocalHostNameSanitizer, CrossDataCenterSanitizing)
...
+        EXPECT_EQ("abc.cX*", sanitizedHost->ToStringBuf());
+    }
+}
+
+TEST(TLocalHostNameSanitizer, PeerEqualToCommonSuffix)
+{
+    THashSet<TString> clusterPeers = {
+        "abc.cX.cZ",
+        "cX.cZ",
+        "e.cX.cZ"
```
maxim-babenko, line 98: Trailing comma please

maxim-babenko:
Shipped!

PR #916, opened 2024-11-04
Fixed grafana sensors node names

jasminetealover:
In k8s operator nodes can have suffix in `service` label based on instance group name. For example `yt-tablet-node-my-bundle` for tablet node instance group with name `my-bundle`.


Kontakter (approved):
Approved by ifsmirnov@ in personal conversation

PR #928, opened 2024-11-06
[queues] YT-23208: Queue static table exports refactor

ItIsApachee:
Mirroring PR for reviewing from github


ItIsApachee:
Need to add small fix to how `accumulatedMinFragmentUnixTs` is calculated. Currently it would break everything if some data is flushed that corresponds to unix ts which has already been exported. I'll do it tomorrow.

achulkov2 (changes requested):
Left some comments here and there, generally I think the code became a bit too confusing and we should take some time to make this PR cleaner.

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -160,7 +189,7 @@ class TQueueExportTask
             .ValueOrThrow()
             .NodeId;
 
-        // Take it for guarantee that only one queue agent instance do this export.
+        // Take it for granted that only one queue agent instance do this export.
```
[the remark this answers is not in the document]
ItIsApachee: It is not possible in a sense that there can't be 2 concurrent transactions doing the same export since we take the same shared lock. It might be possible if leader switch and config change occurred at the same time, but I don't think that is a problem.
Do you agree with that? Or am I missing something?
If you do agree, then I'll re-write the comment to reflect this behavior.

yt/yt/server/queue_agent/config.cpp
```diff
@@ -88,6 +92,14 @@ void TQueueControllerDynamicConfig::Register(TRegistrar registrar)
 
 ////////////////////////////////////////////////////////////////////////////////
 
+void TQueueStaticTableExportManagerDynamicConfig::Register(TRegistrar registrar)
+{
+    registrar.Parameter("max_concurrent_exports", &TThis::MaxConcurrentExports)
```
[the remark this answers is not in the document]
ItIsApachee: I added TQueueStaticTableExportManager, but forgot to stage new files for github. Fixed it already as you can see from the latest commit.

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -241,45 +316,378 @@ class TQueueExportTask
...
+        const NNative::IClientPtr Client;
+        const NNative::IConnectionPtr Connection;
+
+        //! Corresponds to the queue being exported.
+        const TYPath Queue;
+        const TUserObject& QueueObject;
+        const TTableSchemaPtr& QueueSchema;
+        const TMasterTableSchemaId QueueSchemaId;
+
+        const TQueueStaticExportConfig ExportConfig;
```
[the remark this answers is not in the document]
ItIsApachee: Well, I thought that some values are better of copied rather than stored as a pointer or a reference to the actual data, e.g. primitive types. To be completely honest, I'm not that sure how far this argument extends, and at what point do we prefer references or pointers rather than a copy.
In the case of `TQueueStaticExportConfig` I thought that maybe reference/pointer would be better, but at this point should this really be a lite yson struct? If we care enough to save some space for this config, shouldn't we make it ref-counted yson struct?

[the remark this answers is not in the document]
ItIsApachee: Are you against making a copy of `TMasterTableSchemaId` (just a guid)? And did you meaningfully add this comment for the first line? Is there a problem with `const TTableSchemaPtr& QueueSchema;`?

yt/yt/server/queue_agent/queue_static_table_export_manager.cpp
```diff
@@ -0,0 +1,71 @@
...
+    TQueueStaticTableExportManagerDynamicConfig DynamicConfig_;
+    IBoundedConcurrencyInvokerPtr Invoker_;
+};
+
+////////////////////////////////////////////////////////////////////////////////
+
+TQueueStaticTableExportManager::TQueueStaticTableExportManager(
+    TQueueStaticTableExportManagerDynamicConfig dynamicConfig,
+    const IInvokerPtr& underlyingInvoker)
+    : DynamicConfig_(dynamicConfig)
```
savnadya, line 35: std::move

yt/yt/server/queue_agent/queue_static_table_export_manager.cpp
```diff
@@ -0,0 +1,71 @@
...
+    , Invoker_(CreateBoundedConcurrencyInvoker(underlyingInvoker, DynamicConfig_.MaxConcurrentExports))
```
savnadya, line 36: MaxConcurrentExportCount?

yt/yt/server/queue_agent/config.h
```diff
@@ -159,6 +180,9 @@ class TQueueAgentDynamicConfig
     //! Configuration of queue controllers.
     TQueueControllerDynamicConfigPtr Controller;
 
+    //! Configuration of queue static table export manager.
+    TQueueStaticTableExportManagerDynamicConfig QueueStaticTableExportManager;
```
savnadya, line 184: Let's store it as intrusive ptr?
`TQueueStaticTableExportManagerDynamicConfigPtr QueueStaticTableExportManager`

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -211,17 +278,25 @@ class TQueueExportTask
...
 
-        return newExportProgress;
+        QueueExportProgress_ = newExportProgress;
+    }
+
+    TError GetLastSubtaskError() const {
+        return LastSubtaskError_;
+    }
+
+    TQueueExportProgressPtr GetExportProgress() const {
```
savnadya, line 288: And here

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -241,45 +316,378 @@ class TQueueExportTask
...
+        //! Options used for Cypress requests.
+        TTransactionalOptions Options;
+        //! Corresponds to the actual output table created.
+        TUserObject DestinationObject;
+        TTransactionPtr UploadTransaction;
+
+        //! Data statistics collected from attaching chunks.
+        TDataStatistics DataStatistics;
+
+        TQueueExportSubtask(
```
savnadya, line 359: `struct` in our code base usually don't have a constructor or public methods. It looks like a `class`.

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -241,45 +316,378 @@ class TQueueExportTask
...
+            , QueueSchemaId(queueSchemaId)
+            , ExportConfig(exportConfig)
+            , ExportFragmentUnixTs(exportFragmentUnixTs)
+            , ExportInstant(exportInstant)
+            , ChunkSpecsToExport(chunkSpecsToExport)
+            , ParentTransaction(parentTransaction)
+            , Logger(logger)
+        { }
+
+        TError Run() {
```
savnadya, line 386: Move `{` to the next line

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -241,45 +316,378 @@ class TQueueExportTask
...
+            try {
+                GuardedRun();
+            } catch (const std::exception& ex) {
+                return TError("Queue export subtask failed")
+                    << ex;
+            }
+            return TError();
+        }
+
+        void GuardedRun() {
```
savnadya, line 400: And here

savnadya (changes requested):

savnadya (changes requested):

yt/yt/server/queue_agent/queue_static_table_export_manager.cpp
```diff
@@ -0,0 +1,76 @@
...
+IThroughputThrottlerPtr TQueueStaticTableExportManager::GetExportThrottler() const
+{
+    return Throttler_;
+}
+
+void TQueueStaticTableExportManager::OnDynamicConfigChanged(
+    const TQueueStaticTableExportManagerDynamicConfigPtr& /*oldConfig*/,
+    const TQueueStaticTableExportManagerDynamicConfigPtr& newConfig)
+{
+    Throttler_->Reconfigure(TThroughputThrottlerConfig::Create(newConfig->ExportRateLimit));
```
achulkov2, line 62: Let's log the change.

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -143,8 +593,9 @@ class TQueueExportTask
     //! In any case, we will only produce an output table if its unix ts is strictly greater than the one of the last exported table.
     //! If the physical time diverges from the cluster time, the only effect is that some chunks might end up being
     //! exported into later tables, which is perfectly fine.
-    TQueueExportProgressPtr Run()
+    void DoRun()
     {
+        // XXX(apachee): Do we need this now?
```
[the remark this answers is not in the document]
achulkov2: I am not sure removing this will not break tests :) 
Interesting why I didn't write a comment though, I barely remember the reasoning.

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -770,63 +1039,76 @@ TQueueExporter::TQueueExporter(
...
+    auto traceContextGuard = TTraceContextGuard(TTraceContext::NewRoot("QueueExporterPass"));
 
-void TQueueExporter::Export()
-{
-    VERIFY_INVOKER_AFFINITY(Invoker_);
+    auto now = TInstant::Now();
+
+    auto guard = Guard(Lock_);
+    auto exportUnixTs = GetExportFragmentUnixTs(now.Seconds(), ExportConfig_->ExportPeriod);
+    guard.Release();
```
achulkov2, line 1103: Let's introduce some method for getting the whole config under lock, and use it here and in other places to avoid manually releasing guards. But you will need to deep-copy, since dynamic config changes can happen.

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -838,24 +1120,88 @@ void TQueueExporter::Export()
...
+        } else {
+            // NB(apachee): It is possible that task failed before creating updated export progress.
+            // XXX(apachee): Should we assert that exportTaskError is not OK in this case?
+            THROW_ERROR_EXCEPTION("Export task failed to create updated export progress, internal progress stayed the same")
+                << TErrorAttribute("TaskError", exportTaskError)
+                << TErrorAttribute("LastSubtaskError", exportTask->GetLastSubtaskError());
+        }
     }
+
+    guard.Release();
```
achulkov2, line 1165: My personal preference with guards is scoping. If you are not doing that, you might as well use plain `Lock_.Acquire()` and `Lock_.Release()`.
But I think we tend to prefer this:
```
{
    auto guard = Guard(Lock_);
    // Do something.
}
```

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -838,24 +1120,88 @@ void TQueueExporter::Export()
     AlertCollector_->PublishAlerts();
 }
 
-void TQueueExporter::GuardedExport()
+void TQueueStaticTableExporter::GuardedExport(ui64 exportUnixTs)
 {
-    auto config = GetConfig();
+    // XXX(apachee): Throttle inside export task?
+    auto throttler = QueueStaticTableExportManager_->GetExportThrottler();
+    WaitFor(throttler->Throttle(1))
```
achulkov2, line 1127: I think we should account 1 per each exported table.

achulkov2 (changes requested):
First batch of comments.

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -750,18 +967,70 @@ TQueueExportProfilingCounters::TQueueExportProfilingCounters(const TProfiler& pr
...
+
+    const NHiveClient::TClientDirectoryPtr ClientDirectory_;
+    const IInvokerPtr Invoker_;
+    const IQueueStaticTableExportManagerPtr QueueStaticTableExportManager_;
+    const NAlertManager::IAlertCollectorPtr AlertCollector_;
+    const TQueueExportProfilingCountersPtr ProfilingCounters_;
+
+    const NConcurrency::TPeriodicExecutorPtr Executor_;
+
+    const NLogging::TLogger Logger;
```
ItIsApachee, line 1015: Re-order fields such that const fields are grouped together

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -838,24 +1120,88 @@ void TQueueExporter::Export()
...
+        auto guard = Guard(Lock_);
+
+        exportConfig = ExportConfig_;
+
+        exportTask = New<TQueueExportTask>(
+            ClientDirectory_->GetClientOrThrow(Queue_.Cluster),
+            Invoker_,
+            ProfilingCounters_,
+            Queue_.Path,
+            ExportConfig_,
```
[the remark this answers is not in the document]
ItIsApachee: Do we really need to deep-copy this config? I think it is okay to assume that the only changes to config are those  which replace one pointer for another, without changing the value pointers point to.
Maybe this invariant should somehow be enforced? In some code we have `TIntrusivePtr<const TType>`, maybe that can be used here?

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -207,79 +681,43 @@ class TQueueExportTask
...
+        YT_LOG_DEBUG("Finished running subtasks (PreparedSubtaskCount: %v, CompletedSubtaskCount: %v, FailedSubtaskCount: %v)",
+            QueueExportSubtasks_.size(),
+            completedSubtaskCount,
+            failedSubtaskCount);
+
         ProfilingCounters_->ExportedRows.Increment(diff.RowCount);
         ProfilingCounters_->ExportedChunks.Increment(diff.ChunkCount);
-        ProfilingCounters_->ExportedTables.Increment();
+        ProfilingCounters_->ExportedTables.Increment(completedSubtaskCount);
+        ProfilingCounters_->UncompletedExports.Increment(failedSubtaskCount);
```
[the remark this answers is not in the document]
ItIsApachee: What if we use "skipped" terminology here to clearly communicate what is happening?

yt/yt/server/queue_agent/queue_static_table_exporter.h
```diff
@@ -69,6 +75,8 @@ struct TQueueExportProfilingCounters final
     NProfiling::TCounter ExportedRows;
     NProfiling::TCounter ExportedChunks;
     NProfiling::TCounter ExportedTables;
+    NProfiling::TCounter SkippedTables;
```
achulkov2, line 78: Let's add comments for what these mean.

yt/yt/tests/integration/queues/test_queue_agent.py
```diff
@@ -3563,30 +3613,34 @@ def test_export_iteration_with_no_data(self):
             },
         })
 
+        # COMPAT(apachee): We check "last_successful_export_iteration_instant" until third parties rely on this field.
```
achulkov2, line 3616: stop relying?

yt/yt/tests/integration/queues/test_queue_agent.py
```diff
@@ -3235,6 +3146,142 @@ def test_multicell_export(self, queue_external_cell_tag):
...
+        tx = start_transaction()
+        lock(export_dir, mode="shared", tx=tx, attribute_key="queue_static_exporter")
+
+        insert_rows("//tmp/q", [{"data": "sample"}] * 2)
+        self._flush_table("//tmp/q")
+
+        self._sleep_until_next_export_instant(export_period_seconds)
+        time.sleep(export_period_seconds)
+
+        # XXX(apachee): Is it really enough to check that retry is scheduled way before the next export iteration?
```
achulkov2, line 3219: let's add a test where we wait longer, write/flush some more data and then see how it has to retry creating the first table and also create a new one

yt/yt/tests/integration/queues/test_queue_agent.py
```diff
@@ -3235,6 +3146,142 @@ def test_multicell_export(self, queue_external_cell_tag):
...
+
+        wait(lambda: len(ls(export_dir)) == 1)
+        self._check_export(export_dir, [["42"] * 2])
+
+        self._flush_table("//tmp/q")
+        wait(lambda: len(ls(export_dir)) == 2)
+        self._check_export(export_dir, [["42"] * 2, ["13"] * 2])
+
+        self.remove_export_destination(export_dir)
+
```
achulkov2, line 3284: More test ideas:
* Check behaviour of partial export tasks, where only some exports are successful, both to some natural causes (idk how to do it yet, maybe set some quota limits somehow? like set a separate account and a low node quota for it) and configuration limits (max exported tables per task). And also check multiple iterations here, so that tables are not re-exported for example. Also check alerts are propagated correctly here.
* Check ttl depending on how you implement it.
* Check throttler? It is a bit annoying to do, but I recently did it for master r/w user throttlers and I think it is a good practice to write tests for throttlers where possible. 

First one I am strongly requesting.

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -209,79 +363,124 @@ class TQueueExportTask
...
+        YT_LOG_DEBUG("Finished creating exported tables (PreparedTableCount: %v, ExportedTableCount: %v, SkippedTableCount: %v)",
+            TaskParts_.size(),
+            exportedTableCount,
+            skippedTableCount);
+
         ProfilingCounters_->ExportedRows.Increment(diff.RowCount);
         ProfilingCounters_->ExportedChunks.Increment(diff.ChunkCount);
-        ProfilingCounters_->ExportedTables.Increment();
+        ProfilingCounters_->ExportedTables.Increment(exportedTableCount);
+        ProfilingCounters_->SkippedTables.Increment(skippedTableCount);
```
achulkov2, line 374: Can we add a gauge for number of "pending" exports? Say, if exports are failing for 5 export instants already, it should be equal to 5. 

~queue_size for a lot of our services/throttlers

yt/yt/server/queue_agent/queue_agent.cpp
```diff
@@ -7,6 +7,7 @@
 #include "object.h"
 #include "queue_controller.h"
 #include "queue_static_table_exporter.h"
+#include "queue_static_table_export_manager.h"
```
savnadya, line 10: Wrong order

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -764,112 +958,231 @@ TQueueExportProfilingCounters::TQueueExportProfilingCounters(const TProfiler& pr
...
+class TQueueStaticTableExporter
+    : public virtual IQueueStaticTableExporter
 {
-    Executor_->Start();
-    Executor_->SetInterval(DynamicConfig_.Enable
-        ? std::optional(ExportConfig_.ExportPeriod)
-        : std::nullopt);
-}
+public:
+    TQueueStaticTableExporter(
```
savnadya, line 971: Let's name it `TQueueExporter`?
`TQueueExporterDynamicConfig` and `TQueueExportTask` are without `StaticTable`

yt/yt/server/queue_agent/queue_static_table_exporter.cpp
```diff
@@ -764,112 +958,231 @@ TQueueExportProfilingCounters::TQueueExportProfilingCounters(const TProfiler& pr
...
+    TSpinLock Lock_;
+    TQueueStaticExportConfigPtr ExportConfig_;
+    TQueueExporterDynamicConfig DynamicConfig_;
+    TQueueExportProgressPtr ExportProgress_;
+
+    //! Export unix timestamp corresponding to the last successful export.
+    ui64 LastSuccessfulExportUnixTs_ = 0;
+
+    const TString ExportName_;
+    const NQueueClient::TCrossClusterReference Queue_;
```
savnadya, line 1047: Namespace isn't needed here, and below

ItIsApachee:
Sadly, due to refactoring, mirroring diff to github was problematic, and file renaming was not properly recognized, even though I had this in git status:
```
        renamed:    yt/yt/server/queue_agent/queue_static_table_export_manager.h -> yt/yt/server/queue_agent/queue_export_manager.h
        renamed:    yt/yt/server/queue_agent/queue_static_table_exporter.cpp -> yt/yt/server/queue_agent/queue_exporter.cpp
        renamed:    yt/yt/server/queue_agent/queue_static_table_exporter.h -> yt/yt/server/queue_agent/queue_exporter.h
```

yt/yt/server/queue_agent/queue_exporter.cpp
```diff
@@ -0,0 +1,1221 @@
...
+    }
+
+    void ProfileExport(const TQueueStaticExportConfigPtr& exportConfig, int exportTaskErrors)
+    {
+        auto now = TInstant::Now();
+        auto exportUnixTsUpperBound = GetExportUnixTsUpperBound(now, exportConfig->ExportPeriod);
+
+        ui64 timeLagSeconds = 0;
+        auto exportProgress = GetExportProgress();
+        if (exportProgress->LastExportUnixTs <= exportUnixTsUpperBound) {
```
achulkov2, line 1162: Is it possible for `exportProgress->LastExportUnixTs` to be larger than `exportUnixTsUpperBound`?

yt/yt/tests/library/yt_queue_agent_test_base.py
```diff
@@ -91,6 +92,24 @@ class ObjectAlertHelper:
     def __init__(self, alerts):
         self.alerts = alerts
 
+    def check_matching(self, category, text=None, attributes=None):
```
achulkov2, line 95: I would prefer it if we shared code with the helper below. Either put assert vs raise with some callback, or maybe just return False in these cases and outward assert will fail?

achulkov2 (approved):
LGTM, left 2 small comments, you can just leave TODOs for them, if you don't want to fix.

yt/yt/server/queue_agent/queue_exporter.cpp
ItIsApachee: I think it can be the case if there are some issues with system time, but, realistically, I just guarded against unsigned integer overflow to prevent weird values for metrics.

achulkov2: Maybe `YT_LOG_ALERT` for such cases?

Issue #926, opened 2024-11-06
Implement OAuth usernames transformations


l0kix2:
We have OAuth support in the server code. 
[Here](https://github.com/l0kix2/ytsaurus/blob/a60801d86879c27da10f1a9e24e35e0ecb0d9c22/yt/yt/library/auth_server/oauth_service.cpp#L127-L127) our OAuth service makes request in a remote identity server to exchange token from cookie for username. 
The issue here is that identity server may have usernames which are not suitable for ytsaurus users as is. The real world example is username if form `username@company.com` which are problematic in ytsaurus since `@` has special meaning (attribute) in ytsaurus and therefore it is hard (if even possible) to use such usernames/cypress nodes in cli/sdk/api commands, a lot of escaping issues comes up.

The proposed change is to add extra field in [TOAuthServiceConfig](https://github.com/l0kix2/ytsaurus/blob/a60801d86879c27da10f1a9e24e35e0ecb0d9c22/yt/yt/library/auth_server/config.h#L231-L231) which controls transformation info for the username field received from OAuth server and use that setting in `TOAuthService` to apply such transformation.

There can be different approaches to describe simple string manipulations in text config. In [ytsaurus-identity-sync ](https://github.com/tractoai/ytsaurus-identity-sync/blob/main/azure_config.example.yaml#L3-L7) we use `username_replacements` field with list of simple string replacements. But maybe something more generic can be used here (regex match string for example).

achulkov2:
Implemented in #932, reassigning to me and adding Nadya as a reviewer.

PR #932, opened 2024-11-07
Add login transformation ability to OAuth server

achulkov2:
Implements #926. OAuth server can now be configured with `login_transformation`, which will replaces all occurrences of a regex-matcher with the specified replacement pattern.


yt/yt/tests/integration/proxies/test_oauth.py
```diff
@@ -19,7 +19,11 @@ def auth_config(port):
             "user_info_endpoint": "user_info",
             "user_info_login_field": "login",
             "user_info_subject_field": "sub",
-            "user_info_error_field": "error"
+            "user_info_error_field": "error",
+            "login_transformation": {
+                "match_pattern": "(.*)@(.*)\\.(.*)",
+                "replacement": "\\1-\\2",
```
l0kix2, line 25:
```suggestion
                "match_pattern": r'(.*)@(.*)\.(.*)',
                "replacement": r'\1-\2',
```
You may want to use python raw strings to avoid double escaping so regexes would look like ones user would put in text config.

yt/yt/library/auth_server/config.h
```diff
@@ -232,6 +250,12 @@ class TOAuthServiceConfig
     std::optional<TString> UserInfoSubjectField;
     std::optional<TString> UserInfoErrorField;
 
+    //! If specified, configures a global regex replace transformation on the
+    //! contents of the login field. It will replace all non-overlapping matching
+    //! strings with the replacement string. You can use \1-\9 for captured match
+    //! groups in the replacement string, and \0 for the whole match.
+    TStringReplacementConfigPtr LoginTransformation;
```
l0kix2, line 257: Maybe we should give user a hint what type of regex standard (PCRE, POSIX, etc) will be used to interpret his regex in our lib, so he wouldn't have to guess?

zlobober, line 257: I would suggest making this field `std::vector<>` from the very beginning to support various replacements applied in order.

yt/yt/library/auth_server/config.cpp
```diff
@@ -160,6 +170,9 @@ void TOAuthServiceConfig::Register(TRegistrar registrar)
         .Optional();
     registrar.Parameter("user_info_error_field", &TThis::UserInfoErrorField)
         .Optional();
+
+    registrar.Parameter("login_transformation", &TThis::LoginTransformation)
+        .Default();
```
l0kix2, line 175: This change would solve my issue, but I'm not sure that it would be flexible enough in the future.

For example if I want to have transformations like that:
```
user1@acme.com -> user1
user2@othercompany.com -> user2-othercompany
```
I'm not sure how to do it in one regex.

Maybe we can have a list of replacements, so we wouldn't have to make incompatible change in the future since configs are kinda public api?

yt/yt/tests/integration/proxies/test_oauth.py
```diff
@@ -131,3 +135,10 @@ def test_http_user_create(self):
     @authors("kaikash", "ignat")
     def test_http_proxy_oauth_server_error(self):
         assert self._check_deny(cookie="retry_token")
+
+    @authors("achulkov2")
+    def test_login_transformation(self):
+        rsp = self._check_allow(token="good_token", user="achulkov2@nosihvost.de")
```
zlobober, line 141: nosikhvost -> nosokhvost :)

l0kix2: image

yt/yt/library/auth_server/config.h
achulkov2: It's RE2 [syntax](https://github.com/google/re2/wiki/Syntax). Judging from this page, it is "a subset of that accepted by PCRE, roughly speaking, and with various [caveats](https://swtch.com/~rsc/regexp/regexp3.html#caveats)."

I will mention it in the comment.

yt/yt/tests/integration/proxies/test_oauth.py
achulkov2: Suggested changes, much wow.

achulkov2:
Changed to list of transformations, as requested.

yt/yt/tests/integration/proxies/test_oauth.py
```diff
@@ -131,3 +155,17 @@ def test_http_user_create(self):
...
+    def test_login_transformation(self):
+        def check_user(user, expected_login):
+            rsp = self._check_allow(token="good_token", user=user)
+            data = rsp.json()
+            assert data["login"] == expected_login
+            assert data["realm"] == "oauth:token"
+
+        check_user("achulkov2@one-company.de", "achulkov2")
+        check_user("max42@second-company.be", "be-nosokhvost-max42")
+        check_user("john@@third-company.fr", "third-company-john")
```
l0kix2, line 169: what a wierd email john has

l0kix2 (approved):

l0kix2:
> @l0kix2 has triggered the import of a pull request, but it hasn't started. If you are a Yandex employee, you can view [the log](https://nda.ya.ru/t/cH_bSVIJ79Va9E).

hope I haven't merge it or smth

yt/yt/library/auth_server/oauth_service.cpp
```diff
@@ -23,6 +23,8 @@
 
 #include <yt/yt/core/rpc/dispatcher.h>
 
+#include <yt/yt/library/re2/re2.h>
```
savnadya, line 26: same

savnadya (changes requested):

yt/yt/library/auth_server/oauth_service.cpp
```diff
@@ -122,15 +124,33 @@ class TOAuthService
             THROW_ERROR(error);
         }
 
-        const auto& formattedResponose = jsonResponseChecker->GetFormattedResponse()->AsMap();
+        const auto& formattedResponse = jsonResponseChecker->GetFormattedResponse()->AsMap();
+        auto login = formattedResponse->GetChildValueOrThrow<TString>(Config_->UserInfoLoginField);
+        for (const auto& transformation : Config_->LoginTransformations) {
+            YT_VERIFY(transformation->MatchPattern);
```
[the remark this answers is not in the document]
achulkov2: True, added `NonEmpty` validator there.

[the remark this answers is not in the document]
achulkov2: Actually, that is not the same thing, so changed to `Postprocessor`.

PR #942, opened 2024-11-12
Fix various issues related to Decimal and Arrow

zlobober:
* Support parsing of nested decimal128/decimal256 in Arrow.
* Fix bug in decimal256 unversioned value representation --
  before this change, Arrow parser for decimal256(n, p) was
  always emitting strings of length 256 bits, event for n <= 38,
  which is incorrect in the representation of YT decimal(n, p)
  type. Now it produces strings of variadic length (32, 64, 128
  or 256 bits) depending on n, similar to decimal128(n, p).


achulkov2 (approved):

nadya002:
Shipped!

yt/yt/library/formats/unittests/arrow_parser_ut.cpp
```diff
@@ -306,6 +306,26 @@ std::string MakeDecimalArrows(std::vector<TString> values, std::vector<std::tupl
     return MakeOutputFromRecordBatch(recordBatch);
 }
 
+std::string MakeDecimalListArrow(std::vector<TString> values)
```
savnadya, line 309: TString -> std::string

PR #943, opened 2024-11-12
Make wire protocol reader validation limits configurable

Krock21:
Currently, WireProtocol**Writer** can write string values of arbitrary size (It does not do validation), but WireProtocol**Reader** validates string lengths while reading, and it uses constant values for dynamic tables for it.

WireProtocol is used to pass results between QueryTracker and engines (YQL, for example), and they do not have to obey dynamic tables limits. I think it is used in RpcProxies too, while working with static tables.

This PR adds ability to configure validation limits for wire protocol reader, so we can configure it in QueryTracker and RpcProxies to work with big values

You can see how it will be used in QueryTracker in this PR: https://github.com/ytsaurus/ytsaurus/pull/944

### DIFF
--- a/yt/yt/client/table_client/wire_protocol.cpp
+++ b/yt/yt/client/table_client/wire_protocol.cpp
@@ -536,8 +536,9 @@ class TWireProtocolReader
     : public IWireProtocolReader
 {
 public:
-    explicit TWireProtocolReader(TSharedRef data, TRowBufferPtr rowBuffer)
+    explicit TWireProtocolReader(TSharedRef data, TRowBufferPtr rowBuffer, TWireProtocolOptions options)
         : RowBuffer_(rowBuffer ? std::move(rowBuffer) : New<TRowBuffer>(TWireProtocolReaderTag(), ReaderBufferChunkSize))
+        , Options_(std::move(options))
         , Data_(std::move(data))
         , Current_(Data_.Begin())
     { }
@@ -781,9 +782,11 @@ public:
 
 private:
     const TRowBufferPtr RowBuffer_;
+    const TWireProtocolOptions Options_;
 
     TSharedRef Data_;
     TIterator Current_;
+
 
     void ValidateSizeAvailable(size_t size)
     {
@@ -857,15 +860,15 @@ private:
 
     void DoReadStringData(EValueType type, ui32 length, const char** result, bool captureValues)
     {
-        ui32 limit = 0;
+        i64 limit = 0;
         if (type == EValueType::String) {
-            limit = MaxStringValueLength;
+            limit = Options_.MaxStringValueLength;
         }
         if (type == EValueType::Any) {
-            limit = MaxAnyValueLength;
+            limit = Options_.MaxAnyValueLength;
         }
         if (type == EValueType::Composite) {
-            limit = MaxCompositeValueLength;
+            limit = Options_.MaxCompositeValueLength;
         }
         if (length > limit) {
             THROW_ERROR_EXCEPTION("Value of type %Qlv is too long: length %v, limit %v",
@@ -994,10 +997,10 @@ private:
 
     void ValidateVersionedRowTimestampCount(const TVersionedRowHeader& rowHeader)
     {
-        if (rowHeader.WriteTimestampCount > MaxTimestampCountPerRow) {
+        if (rowHeader.WriteTimestampCount > Options_.MaxTimestampCountPerRow) {
             THROW_ERROR_EXCEPTION("Too many write timestamps in a versioned row");
         }
-        if (rowHeader.DeleteTimestampCount > MaxTimestampCountPerRow) {
+        if (rowHeader.DeleteTimestampCount > Options_.MaxTimestampCountPerRow) {
             THROW_ERROR_EXCEPTION("Too many delete timestamps in a versioned row");
         }
     }
@@ -1005,10 +1008,10 @@ private:
     void ValidateVersionedRowDataWeight(TVersionedRow row)
     {
         auto dataWeight = GetDataWeight(row);
-        if (dataWeight > MaxServerVersionedRowDataWeight) {
+        if (static_cast<i64>(dataWeight) > Options_.MaxVersionedRowDataWeight) {
             THROW_ERROR_EXCEPTION("Versioned row data weight is too large: %v > %v",
                 dataWeight,
-                MaxServerVersionedRowDataWeight)
+                Options_.MaxVersionedRowDataWeight)
                 << TErrorAttribute("key", ToOwningKey(row));
         }
     }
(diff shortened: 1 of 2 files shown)


yt/yt/client/table_client/wire_protocol.h
```diff
@@ -280,13 +280,24 @@ struct IWireProtocolReader
 
 ////////////////////////////////////////////////////////////////////////////////
 
+struct TWireProtocolReaderLimits {
```
achulkov2, line 283: I think it is better to name this class `TWireProtocolOptions`. It is closer to how we usually name such classes, and it is more generic if we happen to reuse this somewhere else in the stack.

yt/yt/client/table_client/wire_protocol.h
```diff
@@ -280,13 +280,24 @@ struct IWireProtocolReader
...
+    ui32 MaxStringValueLength = NYT::NTableClient::MaxStringValueLength;
+    ui32 MaxAnyValueLength = NYT::NTableClient::MaxAnyValueLength;
+    ui32 MaxCompositeValueLength = NYT::NTableClient::MaxCompositeValueLength;
+    ui32 MaxTimestampCountPerRow = NYT::NTableClient::MaxTimestampCountPerRow;
+    size_t MaxServerVersionedRowDataWeight = NYT::NTableClient::MaxServerVersionedRowDataWeight;
```
achulkov2, line 288: I think `Server` is unnecessary here. The name of the limit should reflect what it is limiting. The value of the constant specifies to what the constant is attributed (server-side code).

yt/yt/client/table_client/wire_protocol.cpp
```diff
@@ -859,13 +862,13 @@ class TWireProtocolReader
...
         if (type == EValueType::Any) {
-            limit = MaxAnyValueLength;
+            limit = Limits_.MaxAnyValueLength;
         }
         if (type == EValueType::Composite) {
-            limit = MaxCompositeValueLength;
+            limit = Limits_.MaxCompositeValueLength;
         }
         if (length > limit) {
             THROW_ERROR_EXCEPTION("Value of type %Qlv is too long: length %v, limit %v",
```
achulkov2, line 874: Can we improve this error while we are here? It would be helpful, if we could print some information about the position of the value we are reading within the row. Is it possible?

yt/yt/client/table_client/wire_protocol.cpp
```diff
@@ -785,6 +786,8 @@ class TWireProtocolReader
     TSharedRef Data_;
     TIterator Current_;
 
+    const TWireProtocolReaderLimits Limits_;
```
achulkov2, line 789: We declare constant fields first.

achulkov2 (changes requested):

yt/yt/client/table_client/wire_protocol.h
```diff
@@ -280,13 +280,24 @@ struct IWireProtocolReader
 
 ////////////////////////////////////////////////////////////////////////////////
 
+struct TWireProtocolReaderLimits {
+    ui32 MaxStringValueLength = NYT::NTableClient::MaxStringValueLength;
```
achulkov2, line 284: Also, maybe we want to make all of these values signed? We actually discourage using unsigned types for non-negative types (similar to Google C++ code style), and the types of these constants are just legacy. Unsigned types should be used for bitwise-arithmetic-related things.

Krock21: These values are compared to other values in the code. I keep types the same

`MaxStringValueLength` is `ui32` because variable we compare to is `ui32`
https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/client/table_client/wire_protocol.cpp#L860

Same for others

yt/yt/client/table_client/wire_protocol.cpp
Krock21: I believe adding a value index to the error is not that hard, but it will require more code changes than in this PR and I am not sure that it is a useful information

I would try to avoid doing it in the same PR

achulkov2: Ok.

yt/yt/client/table_client/wire_protocol.h
achulkov2: Meh, ok, fine.

yt/yt/client/table_client/wire_protocol.h
```diff
@@ -280,13 +280,24 @@ struct IWireProtocolReader
 
 ////////////////////////////////////////////////////////////////////////////////
 
+struct TWireProtocolOptions {
```
Kontakter, line 283: Place `{` on the new line according style guide

yt/yt/client/table_client/wire_protocol.h
```diff
@@ -280,13 +280,24 @@ struct IWireProtocolReader
...
+    ui32 MaxStringValueLength = NYT::NTableClient::MaxStringValueLength;
+    ui32 MaxAnyValueLength = NYT::NTableClient::MaxAnyValueLength;
+    ui32 MaxCompositeValueLength = NYT::NTableClient::MaxCompositeValueLength;
+    ui32 MaxTimestampCountPerRow = NYT::NTableClient::MaxTimestampCountPerRow;
+    size_t MaxVersionedRowDataWeight = NYT::NTableClient::MaxServerVersionedRowDataWeight;
```
Kontakter, line 288: Why `size_t`, the original type of this constant is `i64`, let use it

yt/yt/client/table_client/wire_protocol.h
```diff
@@ -280,13 +280,24 @@ struct IWireProtocolReader
 
 ////////////////////////////////////////////////////////////////////////////////
 
+struct TWireProtocolOptions {
+    ui32 MaxStringValueLength = NYT::NTableClient::MaxStringValueLength;
```
Kontakter, line 284: Original constant also has type `i64`, let follow original types (check every parameter here, please)

Kontakter:
Shipped!

yt/yt/client/table_client/wire_protocol.h
Krock21: Because we use it to compare to a [size_t variable](https://github.com/ytsaurus/ytsaurus/pull/943/files#diff-47ef756916285c1a2b1d1319972480fef75bdd221e592430c9740f3ec35644fbR1010). [GetDataWeight](https://github.com/ytsaurus/ytsaurus/blob/6a32092f9311afdcc0ee7bb46c292cda97775411/yt/yt/client/table_client/versioned_row.cpp#L24) returns `size_t`

Same for other variables

Originally I did i64 everywhere, but it will require casting one of them for comparison. either `size_t` to `i64`, or the other way

Krock21: Changed all `ui32` to `i64`

Kept `size_t`. It would require casting either `size_t` or `i64` on [this line](https://github.com/ytsaurus/ytsaurus/pull/943/files#diff-47ef756916285c1a2b1d1319972480fef75bdd221e592430c9740f3ec35644fbR1011)

Kontakter: I suppose to use i64 for all these fields

Kontakter (approved):

PR #944, opened 2024-11-12
Configure wire protocol reader in QueryTracker to work with big string values

Krock21:
Issue: https://github.com/ytsaurus/ytsaurus/issues/945

Currently, a simple query `select * from table` can run indefinitely in QueryTracker because QueryTracker will be failing to save it's results with and error `Failed to write query state, backing off. Value of type "string" is too long: length 32000000, limit 16777216`, if the table has a string value longer than 16MB. See new tests for reproduce.

This happens because wire protocol reader validates lengths of string values while reading, and fails. This PR configures wire protocol reader in query tracker to work with bigger values

### DIFF
--- a/yt/yt/server/query_tracker/handler_base.cpp
+++ b/yt/yt/server/query_tracker/handler_base.cpp
@@ -79,33 +79,59 @@ TSharedRef TruncateRowset(TSharedRange<NTableClient::TUnversionedRow>& rows)
     return TSharedRef::MakeEmpty();
 }
 
-void ProcessRowset(TFinishedQueryResultPartial& newRecord, TWireRowset wireSchemaAndSchemafulRowset)
+void ProcessRowset(TFinishedQueryResultPartial& newRecord, TWireRowset wireSchemaAndSchemafulRowset, i64 ValueLengthLimit)
 {
-    // Process schema.
-    auto reader = CreateWireProtocolReader(wireSchemaAndSchemafulRowset.Rowset);
-    auto schema = reader->ReadTableSchema();
-    auto schemaNode = ConvertToNode(schema);
-    // Values in tables cannot have top-level attributes, but we do not need them anyway.
-    schemaNode->MutableAttributes()->Clear();
-    newRecord.Schema = ConvertToYsonString(schemaNode);
+    try {
+        YT_LOG_DEBUG("Processing wire rowset");
+        // Process schema.
+        YT_LOG_DEBUG("Reading schema");
+        TWireProtocolOptions readerOptions;
+        readerOptions.MaxStringValueLength = ValueLengthLimit;
+        readerOptions.MaxAnyValueLength = ValueLengthLimit;
+        readerOptions.MaxCompositeValueLength = ValueLengthLimit;
+        auto reader = CreateWireProtocolReader(wireSchemaAndSchemafulRowset.Rowset, TRowBufferPtr(), readerOptions);
+        auto schema = reader->ReadTableSchema();
+        auto schemaNode = ConvertToNode(schema);
+        // Values in tables cannot have top-level attributes, but we do not need them anyway.
+        schemaNode->MutableAttributes()->Clear();
+        newRecord.Schema = ConvertToYsonString(schemaNode);
 
-    // Process schemaful rowset.
-    auto rowset = reader->Slice(reader->GetCurrent(), reader->GetEnd());
-    auto schemaData = IWireProtocolReader::GetSchemaData(schema);
-    auto rows = reader->ReadSchemafulRowset(schemaData, /*captureValues*/ false);
-    TDataStatistics dataStatistics;
-    dataStatistics.set_row_count(rows.size());
-    dataStatistics.set_data_weight(GetDataWeight(rows));
-    newRecord.DataStatistics = ConvertToYsonString(dataStatistics);
-    newRecord.IsTruncated = wireSchemaAndSchemafulRowset.IsTruncated;
-    if (rowset.Size() <= MaxStringValueLength) {
-        // Fast path. Copy full rowset.
-        newRecord.Rowset = TString(rowset.ToStringBuf());
-    } else {
-        // Slow path. Truncate rowset.
+        // Process schemaful rowset.
+        YT_LOG_DEBUG("Reading schemaful rowset");
+        auto rowset = reader->Slice(reader->GetCurrent(), reader->GetEnd());
+        auto schemaData = IWireProtocolReader::GetSchemaData(schema);
+        TSharedRange<NTableClient::TUnversionedRow> rows;
+        try {
+            rows = reader->ReadSchemafulRowset(schemaData, /*captureValues*/ false);
+        } catch (const std::exception& ex) {
+            THROW_ERROR_EXCEPTION("Failed to read resulting rowset. Try using INSERT INTO to save result") << ex;
+        }
+        TDataStatistics dataStatistics;
+        dataStatistics.set_row_count(rows.size());
+        dataStatistics.set_data_weight(GetDataWeight(rows));
+        newRecord.DataStatistics = ConvertToYsonString(dataStatistics);
+        newRecord.IsTruncated = wireSchemaAndSchemafulRowset.IsTruncated;
+        if (rowset.Size() <= MaxStringValueLength) {
+            // Fast path. Copy full rowset.
+            YT_LOG_DEBUG("Copying full rowset of size %v", rowset.Size());
+            newRecord.Error = TError();
+            newRecord.Rowset = TString(rowset.ToStringBuf());
+        } else {
+            // Slow path. Truncate rowset.
+            YT_LOG_DEBUG("Truncating rowset of size %v", rowset.Size());
+            newRecord.IsTruncated = true;
+            auto truncatedRowset = TruncateRowset(rows);
+            YT_LOG_DEBUG("Copying truncated rowset of size %v", truncatedRowset.Size());
+            newRecord.Error = TError();
+            newRecord.Rowset = TString(truncatedRowset.ToStringBuf());
+        }
+    } catch (const std::exception& ex) {
+        YT_LOG_WARNING(ex, "Failed to save rowset");
+        newRecord.Schema = ConvertToYsonString(TString());
+        newRecord.DataStatistics = ConvertToYsonString(TDataStatistics());
         newRecord.IsTruncated = true;
-        auto truncatedRowset = TruncateRowset(rows);
-        newRecord.Rowset = TString(truncatedRowset.ToStringBuf());
+        newRecord.Error = TError("Failed to save rowset") << ex;
+        newRecord.Rowset = TString();
     }
 }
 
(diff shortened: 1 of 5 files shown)


yt/yt/client/table_client/wire_protocol.h
```diff
@@ -280,13 +280,24 @@ struct IWireProtocolReader
 
 ////////////////////////////////////////////////////////////////////////////////
 
+struct TWireProtocolOptions {
+    ui32 MaxStringValueLength = NYT::NTableClient::MaxStringValueLength;
+    ui32 MaxAnyValueLength = NYT::NTableClient::MaxAnyValueLength;
+    ui32 MaxCompositeValueLength = NYT::NTableClient::MaxCompositeValueLength;
+    ui32 MaxTimestampCountPerRow = NYT::NTableClient::MaxTimestampCountPerRow;
```
Krisha11, line 287: NYT is redundant, isn't it?

yt/yt/server/query_tracker/handler_base.cpp
```diff
@@ -81,15 +81,22 @@ TSharedRef TruncateRowset(TSharedRange<NTableClient::TUnversionedRow>& rows)
 
 void ProcessRowset(TFinishedQueryResultPartial& newRecord, TWireRowset wireSchemaAndSchemafulRowset)
 {
+    YT_LOG_DEBUG("Processing wire rowset");
     // Process schema.
-    auto reader = CreateWireProtocolReader(wireSchemaAndSchemafulRowset.Rowset);
+    YT_LOG_DEBUG("Reading schema");
+    TWireProtocolOptions readerOptions;
+    readerOptions.MaxStringValueLength = MaxRowWeightLimit;
+    readerOptions.MaxAnyValueLength = MaxRowWeightLimit;
```
Krisha11, line 89: Why do we use MaxRowWeightLimit which is marked as "Only for static tables"
https://github.com/ytsaurus/ytsaurus/blob/ccc0a106bda90562675784da9492e08ec5997c3e/yt/yt/client/table_client/public.h#L107

yt/yt/client/table_client/wire_protocol.h
Krisha11, line 287: Why do we need last 2 fields here?

yt/yt/server/query_tracker/handler_base.cpp
Krock21: Before we used `Only for dynamic tables` limit.

For query tracker I see it as a reasonable limit (`select * from static_table`). I can't image real use case where longer values would become even longer (like `select row + "a" from static_table`)

I can put there `numeric_limits<i64>::max()`, if you want. But would prefer to get some estimation of what we want to support and what QueryTracker can handle (let's not give him 100GB lines)

Feel free to propose a better constant in code, or a new one

yt/yt/client/table_client/wire_protocol.h
Krock21: We (In QueryTracker) probably don't. At least we didn't get a case where they matter.

Other wire protocol users may need them, as they may work with different type of data.

>This PR has 2 commits:
>
>1. Adding configurable limits to wire protocol reader
>2. Adding QueryTracker tests and configure limits for wire protocol reader used in QueryTracker
>
>First commit has a separate PR: https://github.com/ytsaurus/ytsaurus/pull/943

Krisha11: It seems strange that we specify only these five parameters instead of configuring only the required ones or all
https://github.com/ytsaurus/ytsaurus/blob/ccc0a106bda90562675784da9492e08ec5997c3e/yt/yt/client/table_client/public.h#L93

yt/yt/server/query_tracker/handler_base.cpp
Krisha11: Maybe we can add a fallback for such cases (like select row + "a" from static_table). If wire protocol reader fails with something like `Value of type "string" is too long` we can at least shorten the result by nothing instead of an infinite loop

yt/yt/client/table_client/wire_protocol.h
Krock21: Not all of them are used in wire protocol
We configure everything that is used in wire protocol (`wire_protocol.cpp`) and may require adjusting

Or is there anything else that may need to be configured?

yt/yt/server/query_tracker/handler_base.cpp
achulkov2: 1. Whatever limit we use, we should not allow the same behavior that happened to us in the first place, with the query just hanging infinitely. We need to handle this validation error, either here on the receiver side, or on the senders' sides, and fail the query.
2. With that in mind, I think we should set our own custom limit based on common sense. Like, idk, 2 gigabytes.
3. This whole behavior is kind of stupid: I am running a query that produces 100 gigabytes, why can't the system give me the first 16MB? It should be able to truncate it. Let's think about implementing truncation in the engines themselves, at least in the future.

yt/yt/tests/integration/queries/test_ql.py
```diff
@@ -109,42 +117,35 @@ def test_types(self, query_tracker):
...
-        new_rows = [{"key": i, "value": ''.join(['a' for _ in range(value_size)])} for i in range(16, 17)]
+        new_rows = [{"key": i, "value": ''.join(['d' for _ in range(value_size)])} for i in range(16, 17)]
         insert_rows("//tmp/t", new_rows)
         rows += new_rows
-        q = start_query("ql", "* from [//tmp/t]", settings=settings)
-        q.track()
-        assert q.get()["result_count"] == 1
-        assert_items_equal(q.read_result(0), rows[:15])
-        assert q.get_result(0)["is_truncated"] == yson.YsonBoolean(True)
+        self._assert_select_result("//tmp/t", settings, rows[:15], True)
```
achulkov2, line 148: Let's add tests for values larger than the allowed limit.

achulkov2 (changes requested):

Krock21:
Rebased, made a limit configurable with a default of 1GB, capture all exceptions when processing rowset with failing a query (without changing state, it stays `completed`, but with an error and empty result)

It means we assume that any exception during rowset processing is non-retryable

yt/yt/tests/integration/queries/test_ql.py
Krock21: Added. In YQL tests

yt/yt/server/query_tracker/handler_base.cpp
```diff
@@ -79,33 +79,54 @@ TSharedRef TruncateRowset(TSharedRange<NTableClient::TUnversionedRow>& rows)
...
-        // Fast path. Copy full rowset.
-        newRecord.Rowset = TString(rowset.ToStringBuf());
-    } else {
-        // Slow path. Truncate rowset.
+    try {
+        YT_LOG_DEBUG("Processing wire rowset");
+        // Process schema.
+        YT_LOG_DEBUG("Reading schema");
+        TWireProtocolOptions readerOptions;
+        readerOptions.MaxStringValueLength = ValueLengthLimit;
```
achulkov2, line 89: This is still a limit only on QT side. I am still very concerned that you can run a query with a huge result that will be piped from YQL agent to QT, only to produce an error. 

Let's at least test (locally and/or on testing clusters) a few things:
1. Is it possible to produce outputs larger than 128M? Maybe there will be another limit somewhere based on the max row size for both static and dynamic tables.
2. What happens when you run a query with a huge output, like 10G, how do the other queries fare?

If you and @Krisha11 are collectively okay with this behavior, I am fine with merging this, but I think it is dangerous.

achulkov2 (changes requested):
A few minor things and a generic concern.

yt/yt/server/query_tracker/handler_base.cpp
Krock21: >This is still a limit only on QT side. I am still very concerned that you can run a query with a huge result that will be piped from YQL agent to QT, only to produce an error.

Assume that there is a separate limit on each engine side. Engine-specific

>Is it possible to produce outputs larger than 128M? Maybe there will be another limit somewhere based on the max row size for both static and dynamic tables.

It is. YQL team said that they have some limit like 2GB (For a single value). I am not sure if it applies to us

>What happens when you run a query with a huge output, like 10G, how do the other queries fare?

It is a separate issue that should be tested and resolved separately. Let's assume that Query Tracker dies in an unexpected way that requires investigation and solution

Huge output is possible already. Example is a table with 8 string columns, each of 16 mb, with 10000 rows. It produces 1.28TB of output. YQL query `SELECT *` will face no issues in the code in this PR

>I am fine with merging this, but I think it is dangerous.

Can you elaborate? This PR doesn't change amount of data transferred between YQL Agents and Query Trackers. It fixes validation error and freezing a query in an "active" state

yt/yt/server/query_tracker/handler_base.cpp
```diff
@@ -79,33 +79,54 @@ TSharedRef TruncateRowset(TSharedRange<NTableClient::TUnversionedRow>& rows)
...
+            newRecord.Rowset = TString(truncatedRowset.ToStringBuf());
+        }
+    } catch (const std::exception& ex) {
+        YT_LOG_WARNING(ex, "Failed to process rowset");
+        newRecord.Schema = ConvertToYsonString(TString());
+        newRecord.DataStatistics = ConvertToYsonString(TDataStatistics());
         newRecord.IsTruncated = true;
-        auto truncatedRowset = TruncateRowset(rows);
-        newRecord.Rowset = TString(truncatedRowset.ToStringBuf());
+        newRecord.Error = TError("Failed to save rowset") << ex;
```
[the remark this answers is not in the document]
achulkov2: The errors might be different though, it could be easy to miss other problems this way, because clients will not report them. I always vote for more categorization/introspection.

yt/yt/server/query_tracker/handler_base.cpp
achulkov2: You are right that there are other ways to transfer huge amounts of data, that's exactly my point — we should forbid it. But ok, you are probably right that it is best to do it in a separate PR.

Krock21: Tested `select *` on a table with 100 16mb rows (1.56gb). YQL agent failed with OOM. Query failed with `Socket read error. Connection reset by peer`

Krock21:
Fixed comments and rebased

yt/yt/server/query_tracker/handler_base.cpp
```diff
@@ -79,33 +79,59 @@ TSharedRef TruncateRowset(TSharedRange<NTableClient::TUnversionedRow>& rows)
...
+            // Slow path. Truncate rowset.
+            YT_LOG_DEBUG("Truncating rowset of size %v", rowset.Size());
+            newRecord.IsTruncated = true;
+            auto truncatedRowset = TruncateRowset(rows);
+            YT_LOG_DEBUG("Copying truncated rowset of size %v", truncatedRowset.Size());
+            newRecord.Error = TError();
+            newRecord.Rowset = TString(truncatedRowset.ToStringBuf());
+        }
+    } catch (const std::exception& ex) {
+        YT_LOG_WARNING(ex, "Failed to process rowset");
```
achulkov2, line 129: nit: log message and error message differ

achulkov2 (approved):
LGTM, but I think @Krisha11 should be the main approver. (and we need her approve to merge anyway)

Krisha11:
```
yt/yql/tests/agent/test_simple.py:505: [E302] expected 2 blank lines, found 1
yt/yql/tests/agent/test_simple.py:555: [E303] too many blank lines (2)
yt/yql/tests/agent/test_simple.py:559: [F541] f-string is missing placeholders
yt/yql/tests/agent/test_simple.py:581: [F541] f-string is missing placeholders
yt/yql/tests/agent/test_simple.py:587: [F541] f-string is missing placeholders
yt/yql/tests/agent/test_simple.py:590: [F841] local variable 'err' is assigned to but never used
```

Krisha11 (approved):

PR #956, opened 2024-11-15
Rework authentication commands in CLI over getpass.

zlobober:
Currently, authentication commands in CLI accept current/new user
passwords as command-line arguments, which is commonly considered
as a bad practice (as password typically leak to the shell history,
are publicly visible during typing, and for other obvious reasons).

For each method among set_user_password, issue_token, revoke_token,
list_user_tokens we introduce a similar command with suffix
"_interactive", which we use in a handler of a corresponding CLI
method. The interactive version reads secrets (passwords, tokens)
via getpass() method. Also, it is aware of server-side logic which
allows omitting the current user's password, if the authenticated user
has administrative permissions over a target user.

Despite being a backwards-incompatible change, it is justified
considering the dangerous behavior of the current CLI implementation.

### DIFF
--- a/yt/python/yt/wrapper/auth_commands.py
+++ b/yt/python/yt/wrapper/auth_commands.py
@@ -11,7 +11,7 @@ def set_user_password(user, new_password, current_password=None,
                       client=None):
     """Updates user password."""
     params = {"user": user, "new_password_sha256": encode_sha256(new_password)}
-    if current_password:
+    if current_password is not None:
         params["current_password_sha256"] = encode_sha256(current_password)
     return make_request(
         "set_user_password",
@@ -36,6 +36,8 @@ def revoke_token(user, password=None, token=None, token_sha256=None,
                  client=None):
     """Revokes user token."""
     params = {"user": user}
+    if not token_sha256 and not token:
+        raise ValueError("Either token or token_sha256 must be provided")
     if not token_sha256:
         token_sha256 = encode_sha256(token)
     params["token_sha256"] = token_sha256
--- a/yt/python/yt/wrapper/cli_impl.py
+++ b/yt/python/yt/wrapper/cli_impl.py
@@ -1,8 +1,10 @@
-import yt.wrapper as yt
-import yt.logger as logger
 from .http_helpers import get_user_name
 from .common import YtError
 from .auth_commands import encode_sha256
+
+import yt.wrapper as yt
+import yt.logger as logger
+
 from getpass import getpass
 
 
--- a/yt/python/yt/wrapper/cli_impl.py
+++ b/yt/python/yt/wrapper/cli_impl.py
@@ -1,5 +1,9 @@
 import yt.wrapper as yt
 import yt.logger as logger
+from .http_helpers import get_user_name
+from .common import YtError
+from .auth_commands import encode_sha256
+from getpass import getpass
 
 
 def _set_attribute(path, name, value, recursive, client=None):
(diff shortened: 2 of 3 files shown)


Kontakter:
I'll try to read and accept this PR on Monday

yt/python/yt/wrapper/auth_commands.py
```diff
@@ -57,3 +120,9 @@ def list_user_tokens(user, password=None,
         "list_user_tokens",
         params=params,
         client=client)
+
+
+def list_user_tokens_interactive(user, client=None):
+    """Lists sha256-encoded user tokens, reading password interactively and securely from console."""
+    password = maybe_request_password(user, client=client)
+    return list_user_tokens(user, password=password, client=client)
```
gritukan, line 128: \n

yt/python/yt/wrapper/auth_commands.py
```diff
@@ -19,19 +54,38 @@ def set_user_password(user, new_password, current_password=None,
...
 
+def set_user_password_interactive(user, client=None):
+    """Updates user password, reading it and optionally a current password
+    interactively and securely from console."""
+
+    current_password = maybe_request_password(user, client=client)
+    new_password = getpass("New password: ")
+    new_password_retyped = getpass("Retype new password: ")
+    if new_password != new_password_retyped:
+        raise YtError("Passwords do not match.")
```
Kontakter, line 65: Drop dot at the end of the sentense

yt/python/yt/wrapper/auth_commands.py
```diff
@@ -19,19 +54,38 @@ def set_user_password(user, new_password, current_password=None,
...
     return make_request(
         "issue_token",
         params=params,
         client=client)
 
 
+def issue_token_interactive(user, client=None):
+    """Issues a new token for user, reading password interactively and securely from console."""
+    password = maybe_request_password(user, client=client)
+    res = issue_token(user, password=password, client=client)
```
Kontakter, line 85: Just `return`, additional variable is not needed

BTW: is this code written by some AI?

yt/python/yt/wrapper/auth_commands.py
```diff
@@ -57,3 +120,9 @@ def list_user_tokens(user, password=None,
         "list_user_tokens",
         params=params,
         client=client)
+
+
+def list_user_tokens_interactive(user, client=None):
```
Kontakter, line 125: It is not a strong opinion, but let consider placing all of these `*_interactive` methods to `cli_impl.py`, since interactive commands look relevant for CLI scenarios only

Kontakter (changes requested):

Kontakter:
There are many test failures, for example: test_yql_agent_overload (mpereskokova)

```
setup failed:
contrib/python/flaky/py3/flaky/flaky_pytest_plugin.py:146: in <lambda>
    lambda: ihook(item=item, **kwds), when=when, reraise=reraise
yt/yql/tests/conftest.py:68: in yql_agent
    with YqlAgent(cls.Env, count, libraries) as yql_agent:
yt/yql/tests/conftest.py:27: in __init__
    self.yql_agent.prepare(env, config={
yt/python/yt/environment/components/yql_agent/__init__.py:49: in prepare
    yql_agent_token = yson.loads(self.client.issue_token(self.USER_NAME))
E   RuntimeError: Object 'None' is not bytes or unicode string
```

yt/python/yt/wrapper/auth_commands.py
zlobober: Me + GH copilot in some variadic proportion

yt/python/yt/wrapper/auth_commands.py
zlobober: Moved to cli_impl.py.

zlobober:
> There are many test failures, for example: test_yql_agent_overload (mpereskokova)
> 
> ```
> setup failed:
> contrib/python/flaky/py3/flaky/flaky_pytest_plugin.py:146: in <lambda>
>     lambda: ihook(item=item, **kwds), when=when, reraise=reraise
> yt/yql/tests/conftest.py:68: in yql_agent
>     with YqlAgent(cls.Env, count, libraries) as yql_agent:
> yt/yql/tests/conftest.py:27: in __init__
>     self.yql_agent.prepare(env, config={
> yt/python/yt/environment/components/yql_agent/__init__.py:49: in prepare
>     yql_agent_token = yson.loads(self.client.issue_token(self.USER_NAME))
> E   RuntimeError: Object 'None' is not bytes or unicode string
> ```

Ok, reverted back the output_format part, will think what to do with that separately (or, more realistically, will not think about that any more...)

yt/python/yt/wrapper/cli_impl.py
```diff
@@ -1,5 +1,9 @@
 import yt.wrapper as yt
 import yt.logger as logger
+from .http_helpers import get_user_name
+from .common import YtError
+from .auth_commands import encode_sha256
+from getpass import getpass
```
Kontakter, line 6: We usually order imports from local one to global one and separate groups within newline, in this case:
```
from .http_helpers import get_user_name
from .common import YtError
from .auth_commands import encode_sha256

import yt.wrapper as yt
import yt.logger as logger

from getpass import getpass
```

Kontakter (approved):

yt/python/yt/wrapper/cli_impl.py
zlobober: Done

Kontakter (approved):

PR #969, opened 2024-11-26
[odin] Stop using transaction for singular insert in dynamic table commands check; add ability to configure driver logging for odin server

achulkov2:
This PR does multiple small things:
* Removes client transaction from the `insert_rows` invocation in the `dynamic_table_commands` check. This allows client retries to work. Transaction is still created by proxies, so nothing changes logically.
* Adds table_id to `dynamic_table_commands` check logs, so that it is easier to debug problems with tables that were already deleted.
* Makes `driver_logging_config` configurable for odin server.


Kontakter (approved):
LGTM

PR #971, opened 2024-11-27
Build YQL python udfs by default

Krock21:
[This commit](https://github.com/ytsaurus/ytsaurus/commit/e047f03f6468bd23d25540db358bb13d50f6b896) changed default behaviour of `build_yql_agent.sh` (building python udf -> not building python udf)

I think by default such scripts should build all released functionality

We used this script to build our own images, and then suddenly lost python udfs. It is possible that someone else did the same

Adjusting default value for building python udfs in this script and consumers who relied to it


savnadya:
Shipped!

PR #994, opened 2024-12-13
Add extra credentials for YQL plugin.

Tony-Romanov:
https://github.com/ydb-platform/ydb/issues/9401


yt/yql/agent/yql_agent.cpp
```diff
@@ -378,8 +378,9 @@ class TYqlAgent
             auto refreshTokenExecutor = New<TPeriodicExecutor>(ControlInvoker_, BIND(&RefreshToken, user, token, queryClients), Config_->RefreshTokenPeriod);
             refreshTokenExecutor->Start();
 
+            const THashMap<TString, THashMap<TString, TString>> credentials = {{"yt_default", {{"category", "yt"}, {"content", token}}}};
```
Krisha11, line 381: We are not adding any new credentials here, we are simply changing the way existing ones are provided. Will this be implemented in the next PR?

Krisha11 (approved):

yt/yql/agent/yql_agent.cpp
Krisha11, line 381: I'm going to change it to "default_yt" as it was and merge, because it's required for other PRs
https://github.com/ydb-platform/ydb/pull/12584/files#diff-baedb7407d57fe1aebc64aed1bf0550aba089d8491969f87b7c38f6fe7ac6a18L468
It works incorrect with this variant.

PR #996, opened 2024-12-16
[24.1] Fix account profiling incumbency

achulkov2:
Current implementation is racy for two reasons.

Firstly, the following scenario is possible:
1) Peer is a leader and profiles metrics
2) Leader is switched, profiling invoker is destroyed 3) Dynamic config changes, StartAccountsProfiling is called and new (epoch) invoker in created 4) Peer becomes leader again, invoker is not recreated while it is still bound to the finished epoch

Secondly, invoker is always stopped when some incumbency finishes, while it is possible that only a part of incumbencies should be finished (i.e. during the change of number of alive followers).

This PR fixes these problems and also does some code cleanup.

---
Co-authored-by: Grigory Reznikov <gritukan@tracto.ai>
7dca0641dbbcf55beb2bdf44bc91678f49af33b5

Pull Request resolved: https://github.com/ytsaurus/ytsaurus/pull/701

--

* Changelog entry
   Type: fix
   Component: master

Fixed issues with accounts profiling disabling itself during maintenance/master leader switches.

### DIFF
--- a/yt/yt/server/master/security_server/security_manager.cpp
+++ b/yt/yt/server/master/security_server/security_manager.cpp
@@ -4067,48 +4067,54 @@ private:
     void OnIncumbencyStarted(int shardIndex) override
     {
         TShardedIncumbentBase::OnIncumbencyStarted(shardIndex);
-        StartAccountsProfiling();
+        MaybeToggleAccountsProfiling();
     }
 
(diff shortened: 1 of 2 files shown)


raundn:
Shipped!

PR #998, opened 2024-12-16
Make solomon scrape options externally configurable and add some new options for prometheus

achulkov2:
This PR adds some new behavior for handling metric scraps and makes them more externally configurable.

Some new options are added:
* `sensor_component_delimiter`: allows specifying the delimiter, which should replace the `/` symbol used to separate metric "categories" used within the codebase. E.g. the sensor `yt/resource_tracker/total_cpu` turns into `yt.resource_tracker.total_cpu` by default (existing behavior) and into `yt-resource_tracker-total_cpu` if `sensor_component_delimiter`=`"-"` (new feature).
* `convert_sensor_component_names_to_camel_case`: if true, sensor component names like `resource_tracker` and `total_cpu` are converted from underscore case to camel case.
* `add_metric_type_label`: if true, sensors are instrumented with an additional `metric_type` label, which contains the type of the time series from the perspective of our internal monitoring library (e.g. `GAUGE`, `RATE`, `COUNTER`, `HIST`, etc).

Additionally, most scraping options can now be provided as GET parameters when fetching metrics, e.g. `GET .../solomon/all?convert_counters_to_rate_gauge=true`. These parameters are parsed as YSON and shared with the static solomon exporter configuration (but have precedence over it). 

Also added some comments here and there across eporter.cpp/cube.cpp for better understanding of the implementation, and documented all scrape options in detail.

* Changelog entry
Type: feature
Component: misc-server

Configurable scrape options for solomon exporter. New scrape options to allow configuration of delimiters between components in sensor names and potential conversion of component names to CamelCase. New scrape option to instrument RATE time series.

### DIFF
--- a/yt/yt/library/profiling/solomon/cube.h
+++ b/yt/yt/library/profiling/solomon/cube.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "public.h"
 #include "private.h"
 #include "tag_registry.h"
 
@@ -23,18 +24,15 @@ struct TReadOptions
 
     std::function<bool(const std::string&)> SensorFilter;
 
-    bool ConvertCountersToRateGauge = false;
-    bool ConvertCountersToDeltaGauge = false;
-    bool RenameConvertedCounters = true;
+    // NB: Be careful when modifying this field, as you might be modifying
+    // a shared object. If you need to modify it, make a deep copy first.
+    // TODO(achulkov2): Ideally, it should be a lite YSON struct, but I do
+    // not know how to use such structs as bases for ref-counted YSON structs.
+    TScrapeOptionsPtr ScrapeOptions;
+
     double RateDenominator = 1.0;
+
     bool EnableHistogramCompat = false;
-
-    bool EnableSolomonAggregationWorkaround = false;
-
-    // Direct summary export is not supported by solomon, yet.
-    ESummaryPolicy SummaryPolicy = ESummaryPolicy::Default;
-
-    bool MarkAggregates = false;
 
     std::optional<std::string> Host;
 
@@ -45,6 +43,8 @@ struct TReadOptions
     bool DisableSensorsRename = false;
     bool DisableDefault = false;
 
+    ESummaryPolicy SummaryPolicy = ESummaryPolicy::Default;
+
     int LingerWindowSize = 0;
 
     // Used only in ReadRecentSensorValue.
@@ -52,9 +52,6 @@ struct TReadOptions
 
     // Only makes sense with ExportSummaryAsMax and ReadAllProjections.
     bool SummaryAsMaxForAllTime = false;
-
-    // Drop all prefix before last '/'.
-    bool StripSensorsNamePrefix = false;
 };
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/yt/yt/tests/integration/misc/test_profiling.py
+++ b/yt/yt/tests/integration/misc/test_profiling.py
@@ -1,0 +1,277 @@
+from yt_env_setup import YTEnvSetup
+
+from yt_commands import (authors, get, raises_yt_error)
+
+from yt.common import YtResponseError
+
+import pytest
+import requests
+import simplejson
+import json
+
+##################################################################
+
+
+def try_parse_yt_error_headers(rsp):
+    if "X-YT-Error" in rsp.headers:
+        assert "X-YT-Framing" not in rsp.headers
+        raise YtResponseError(json.loads(rsp.headers.get("X-YT-Error")))
+    rsp.raise_for_status()
+
+
+##################################################################
+
+
+class Sensor:
+    def __init__(self, name, value, timestamp, kind=None, labels=None):
+        self.name = name
+        self.value = value
+        self.timestamp = timestamp
+        self.kind = kind
+        self.labels = labels or {}
+
+    def from_json(json):
+        try:
+            return Sensor(
+                json["labels"]["sensor"],
+                float(json["value"]) if "value" in json else json["hist"],
+                json["ts"],
+                json["kind"],
+                json["labels"],
+            )
+        except KeyError as e:
+            raise ValueError(f"Invalid sensor JSON: {json}") from e
+
+    # This might not work for histograms or some other types, but we don't check them (for now).
+    def from_prometheus_line(line, kind):
+        # yt_http_connections_accepted{yt_aggr="1", another_label="5"} 1 1733941661000
+        # yt_tablet_server_tablet_cell_decommissioner_check_orphans_max 0.000003 1733943237000
+        try:
+            labels = {}
+            if "{" in line:
+                name, rem = line.split("{")
+                labels_raw, rem = rem.split("} ")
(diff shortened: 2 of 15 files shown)


Kontakter:
Shipped!

Kontakter:
(make shipped command to run tests in internal CI)

Kontakter:
In arcadia there are external usages of deleted options: `EnableSolomonAggregationWorkaround`, `ConvertCountersToRateForSolomon`, `ConvertCountersToDeltaGauge`

We should investigate the details

achulkov2:
> In arcadia there are external usages of deleted options: EnableSolomonAggregationWorkaround, ConvertCountersToRateForSolomon, ConvertCountersToDeltaGauge
We should investigate the details

They just moved in a somewhat non-obvious way via aliases: https://github.com/ytsaurus/ytsaurus/pull/998/files#diff-c461e77a566c227be09a3c45403c014069d9e55c96e6555705ac656fa000e210R24. So nothing should fail, and if it does, I missed something.

Kontakter:
Error:
```
safe_stats_ut.cpp:103:17: error: no member named 'EnableSolomonAggregationWorkaround' in 'NYT::NProfiling::TReadOptions'
  103 |         options.EnableSolomonAggregationWorkaround = false;
```

Code snippet (copied from this file):
```
...
NYT::NProfiling::TReadOptions options; 
options.EnableSolomonAggregationWorkaround = false; 
options.Times = {{{registry->IndexOf(i)}, TInstant::Zero()}};
...
```

achulkov2:
```...
NYT::NProfiling::TReadOptions options; 
options.EnableSolomonAggregationWorkaround = false; 
options.Times = {{{registry->IndexOf(i)}, TInstant::Zero()}};
...```

I see. Frankly, there is no way around manual refactoring by the Yandex reviewer here, because this looks like a usage of the internal API provided by `TSolmonExporter` for unit tests specifically, and I changed the structure of `TReadOptions` a bit.

**Comment by achulkov2** [comment]
```2024-12-24 15:01:33,808042	D	Http	Received HTTP request (ConnectionId: 460051ee-8bcbf3e3-300ca617-fc1f3d8c, RequestId: 9fb75c77-b8bd90e3-3a0df060-8d53d72f, Method: Get, Path: /solomon/all, L7RequestId: <null>, L7RealIP: <null>, UserAgent: Solomon-Fetcher/38470)	HttpMonFS:0	fffeecece1705b89	
2024-12-24 15:01:33,808447	D	Solomon	Failed to export sensors (Shard: <null>)\nError occurred while parsing YSON\n    origin          man0-0460.hw.nebius.yt (pid 1, thread ProfOffload:0, fid fffeecece165bee5)\n    datetime        2024-12-24T15:01:33.808359Z\n    context_pos     0\n    context         2024-12-24T15:01:30Z\n\n  Unexpected "T" in numeric literal\n      origin          man0-0460.hw.nebius.yt (pid 1, thread ProfOffload:0, fid fffeecece165bee5)\n      datetime        2024-12-24T15:01:33.808303Z\n      offset          10\n	ProfOffload:0fffeecece165bee5	```
Fun errors while we still have some solomon installations. Will need to fix that somehow.

File: yt/yt/tests/integration/misc/test_profiling.py [hunk 1]
```diff
@@ -0,0 +1,277 @@
...
+        def is_converted_gauge(sensor):
+            return sensor.kind == "GAUGE" and not (rename_converted_counters ^ sensor.name.endswith(conversion_type))
+
+        def is_rate(sensor):
+            return sensor.kind == "RATE" and not sensor.name.endswith(conversion_type)
+
+        sensors = self.get_sensors(params={
+            f"convert_counters_to_{conversion_type}_gauge": "true",
+            "rename_converted_counters": "true" if rename_converted_counters else "false",
+            "add_metric_type_label": "true"})
```
achulkov2, line 218: Somehow I forgot to test the actual output values of this label.

eshcherbin (changes requested):

yt/yt/library/profiling/solomon/cube.h
```diff
@@ -23,18 +24,15 @@ struct TReadOptions
...
-    bool RenameConvertedCounters = true;
-    double RateDenominator = 1.0;
-    bool EnableHistogramCompat = false;
-
-    bool EnableSolomonAggregationWorkaround = false;
+    // NB: Be careful when modifying this field, as you might be modifying
+    // a shared object. If you need to modify it, make a deep copy first.
+    // TODO(achulkov2): Ideally, it should be a lite YSON struct, but I do
+    // not know how to use such structs as bases for ref-counted YSON structs.
+    TScrapeOptionsPtr ScrapeOptions;
```
eshcherbin, line 31: Naming-wise, I would prefer not to use the term "scrape" here, even though it is well-known in the industry. The reason is we already have a [probably not the best] term for this operation: "read sensors". So it feels natural to name this new struct `TReadOptions`.

The problem is this name is already taken by the outer struct. Wrongfully taken, I believe, as essentially this outer struct is a bunch of parameters which we've consolidated to not have methods with millions of arguments. We call these structs "contexts" in our codebase, so I suggest the following renaming: `TReadOptions -> TReadSensorsContext`, `TScrapeOptions -> TReadOptions`.

achulkov2: Nice names, I think I can live with them just fine!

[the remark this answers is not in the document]
achulkov2: > Frankly, I find current implementation choice of copying readOptions and overwriting copied values to be problematic. 
Me too, definitely. I have to take some time to reacquaint myself with the issues, right now I am not sure I understand what the proposed change actually is.

Did I correctly interpret that you want to make config part of the context unmodifiable (as they usually are, and usually just don't bother with storing const, I don't remember if it causes any trouble), and stop doing refcounted copying in favour of modifying test code? 

I will attempt it and get back to you.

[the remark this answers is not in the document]
eshcherbin: > Me too, definitely. I have to take some time to reacquaint myself with the issues, right now I am not sure I understand what the proposed change actually is.

Current `TReadOptions` consists of two types of fields: (1) parameters that are defined once in the exporter and then are not modified throughout all of the following nested functions, (2) variables that are used somewhere deeper inside and are placed in this struct for convenience, so that we don't need to pass more arguments in some methods. 

Copying this struct makes it easier to modify fields of the second type (we don't need to return them back to the original state), but it may corrupt a developer into thinking that everything is modifiable, which could lead to bad design choices. I propose to enforce a single instance of this struct which will be passed by pointer, not copied. A "context" of the current read sensors invocation. This would require careful thinking in the places where we want to modify something, e.g. eventually removing `InstanceTags` that are added at some places. But it would have the benefit of having unmodifiable fields by default.

I don't suggest doing this particular change in the current PR, but I think it's something that should be done at some point.

Regarding your second question:

> Did I correctly interpret that you want to make config part of the context unmodifiable (as they usually are, and usually just don't bother with storing const, I don't remember if it causes any trouble), and stop doing refcounted copying in favour of modifying test code?

Yes.

PR #1001, opened 2024-12-16
added max_inline_hunk_size in formats/extension.proto

jasminetealover:
We are using these extensions (from yt-cpp-sdk). not sure where in C++ code this should be added.


Kontakter (approved):

Kontakter:
It was committed separately by me, sorry, I should discard this PR

PR #1004, opened 2024-12-20
Introduce temporary tokens for operations

achulkov2:
This PR adds two new spec options available for all operations. If `issue_temporary_token` is set to true, the scheduler will issue a token for the authenticated user of the operation via the Cypress tokens mechanism. The token will be placed into the operation's secure vault with `temporary_token_environment_variable_name` used as key. 

The token is removed by operations cleaner. In order for that to work, we persist the Cypress node id of the issued token in the `temporary_token_node_id` operation node attribute. Such dependent nodes are removed strictly before removing the operation node.

To improve performance, instead of issuing a token transactionally with creating the operation node, we issue it with an expiration timeout first, and then remove the expiration timeout after the operation node is created. If scheduler fails after token is issued but before the operation node is actually created, the expiration timeout will ensure that the token is eventually deleted.

I chose this approach over more typical lease extension mechanisms mainly because I feel that the proposed solution is simpler than introducing new periodic logic to scheduler/CA. There is also the concern of tokens expiring during job revival, which is unlikely if the lease timeout is large enough, but still makes me like the other solution a bit less.

* Changelog entry
Type: feature
Component: scheduler

Add the ability to issue temporary tokens valid during the lifetime of an operation and place them into the operation's secure vault.

### DIFF
--- a/yt/yt/client/api/security_client.h
+++ b/yt/yt/client/api/security_client.h
@@ -89,6 +89,9 @@ struct TIssueTemporaryTokenOptions
 struct TIssueTokenResult
 {
     TString Token;
+    //! Cypress node corresponding to issued token.
+    //! Deleting this node will revoke the token.
+    NCypressClient::TNodeId TokenNodeId;
 };
 
 struct TRefreshTemporaryTokenOptions
--- a/yt/yt/client/api/security_client.h
+++ b/yt/yt/client/api/security_client.h
@@ -91,7 +91,7 @@ struct TIssueTokenResult
     TString Token;
     //! Cypress node corresponding to issued token.
     //! Deleting this node will revoke the token.
-    NCypressClient::TNodeId TokenNodeId;
+    NCypressClient::TNodeId NodeId;
 };
 
 struct TRefreshTemporaryTokenOptions
--- a/yt/yt/server/scheduler/operations_cleaner.cpp
+++ b/yt/yt/server/scheduler/operations_cleaner.cpp
@@ -55,6 +55,7 @@
 namespace NYT::NScheduler {
 
 using namespace NConcurrency;
+using namespace NCypressClient;
 using namespace NYTree;
 using namespace NYson;
 using namespace NObjectClient;
@@ -112,6 +113,7 @@ const std::vector<TString>& TArchiveOperationRequest::GetAttributeKeys()
         "experiment_assignments",
         "controller_features",
         "provided_spec",
+        "temporary_token_node_id",
     };
 
     return attributeKeys;
@@ -598,7 +600,7 @@ public:
         : Config_(std::move(config))
         , Bootstrap_(bootstrap)
         , Host_(host)
-        , RemoveBatcher_(New<TNonblockingBatcher<TOperationId>>(
+        , RemoveBatcher_(New<TNonblockingBatcher<TRemoveOperationRequest>>(
             TBatchSizeLimiter(Config_->RemoveBatchSize),
             Config_->RemoveBatchTimeout))
         , ArchiveBatcher_(New<TNonblockingBatcher<TOperationId>>(
@@ -682,15 +684,15 @@ public:
             }).Via(GetCancelableInvoker()));
     }
 
-    void SubmitForRemoval(std::vector<TOperationId> operationIds)
+    void SubmitForRemoval(std::vector<TRemoveOperationRequest> requests)
     {
         if (!IsEnabled()) {
             return;
         }
 
-        GetCancelableInvoker()->Invoke(BIND([this, this_ = MakeStrong(this), operationIds = std::move(operationIds)] {
-            for (auto operationId : operationIds) {
-                EnqueueForRemoval(operationId);
+        GetCancelableInvoker()->Invoke(BIND([this, this_ = MakeStrong(this), requests = std::move(requests)] {
+            for (auto request : requests) {
+                EnqueueForRemoval(request);
             }
         }));
     }
@@ -780,6 +782,8 @@ public:
             result.FullSpec = initializationAttributes->FullSpec;
         }
 
+        result.DependentNodeIds = operation->GetDependentNodeIds();
+
         return result;
     }
 
@@ -835,6 +839,10 @@ public:
         result.TaskNames = attributes.FindYson("task_names");
         result.ControllerFeatures = attributes.FindYson("controller_features");
 
+        if (auto temporaryTokenNodeId = attributes.Find<TNodeId>("temporary_token_node_id")) {
+            result.DependentNodeIds = std::vector{*temporaryTokenNodeId};
+        }
+
         return result;
     }
 
@@ -860,9 +868,13 @@ private:
     std::multimap<TInstant, TOperationId> ArchiveTimeToOperationIdMap_;
     THashMap<TOperationId, TArchiveOperationRequest> OperationMap_;
 
-    TIntrusivePtr<TNonblockingBatcher<TOperationId>> RemoveBatcher_;
+    // Removals might be issued for operations without an entry in OperationMap_,
+    // so we need to store the complete removal request in the queue.
+    TIntrusivePtr<TNonblockingBatcher<TRemoveOperationRequest>> RemoveBatcher_;
+    // We can store plain operation ids here because operations going
+    // through archivation have a corresponding entry in OperationMap_.
     TIntrusivePtr<TNonblockingBatcher<TOperationId>> ArchiveBatcher_;
-    std::deque<std::pair<TOperationId, TInstant>> LockedOperationQueue_;
+    std::deque<std::pair<TRemoveOperationRequest, TInstant>> LockedOperationQueue_;
 
     std::deque<TOperationAlertEvent> OperationAlertEventQueue_;
     TInstant LastOperationAlertEventSendTime_;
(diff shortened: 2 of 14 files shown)


koct9i:
I don't like term "temporarary" here.

By owning this is "operation-token",
by meaning this is "delegation-token".

yt/yt/client/api/security_client.h
```diff
@@ -89,6 +89,9 @@ struct TIssueTemporaryTokenOptions
 struct TIssueTokenResult
 {
     TString Token;
+    //! Cypress node corresponding to issued token.
+    //! Deleting this node will revoke the token.
+    NCypressClient::TNodeId TokenNodeId;
```
Kontakter, line 94: I think it is more convenient to name this field just `NodeId`

yt/yt/server/scheduler/scheduler.cpp
```diff
@@ -2659,6 +2660,29 @@ class TScheduler::TImpl
             return;
         }
 
+        if (operation->Spec()->IssueTemporaryToken) {
+            try {
+                // We check this first, so that we don't create a token in vain.
+                auto secureVault = operation->GetSecureVault();
+                if (secureVault && secureVault->FindChild(operation->Spec()->TemporaryTokenEnvironmentVariableName)) {
+                    THROW_ERROR_EXCEPTION("Temporary token environment variable %Qv already exists in secure vault", operation->Spec()->TemporaryTokenEnvironmentVariableName);
```
Kontakter, line 2668: Move the second argument on the new line

Kontakter (approved):
I only make a high-level look by now 

Approved to run tests in internal CI

Kontakter (commented):
Small remarks

Kontakter:
I observe many failures around `clean_operations`

One of examples:
```
yt/yt/tests/integration/controller/test_features.py:140: in test_controller_features
    clean_operations()
yt/yt/tests/library/yt_commands.py:1748: in clean_operations
    wait(lambda: not operation_nodes_exist())
yt/python/yt/common.py:881: in wait
    raise WaitFailed(error_message)
E   yt.common.WaitFailed: Wait failed (timeout = 30.0)
```

If you fail to reproduce – let me know

yt/yt/tests/integration/node/test_user_job.py
```diff
@@ -2085,6 +2085,199 @@ def test_cypress_missing_secure_vault(self):
...
+        )
+
+    @authors("achulkov2")
+    def test_basic(self):
+        create_user("alice")
+        assert not list_user_tokens("alice")
+
+        command = with_breakpoint(f"""
+            export YT_PROXY={self.Env.get_proxy_address()};
+            curl -X POST \"http://$YT_PROXY/api/v3/create?path=//tmp/created_from_operation&type=map_node\" \\
```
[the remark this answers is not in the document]
achulkov2: Actually, I was copying from the [docs](https://ytsaurus.tech/docs/ru/user-guide/proxy/http) about HTTP proxies, and didn't notice it. So we should also update the docs.

achulkov2:
> I observe many failures around clean_operations

Fixed. It was a stupid bug with forgotten `!= NullObjectId` in one place, after which I decided that storing `std::optional<TNodeId>` and not having this attribute in Cypress when no token is issued is clearer. Changed that and addressed the few comments above.

Kontakter:
Shipped!

yt/yt/server/scheduler/scheduler.cpp
```diff
@@ -2659,6 +2660,31 @@ class TScheduler::TImpl
...
+                    THROW_ERROR_EXCEPTION(
+                        "Temporary token environment variable %Qv already exists in secure vault",
+                        operation->Spec()->TemporaryTokenEnvironmentVariableName);
+                }
+
+                auto temporaryToken = WaitFor(MasterConnector_->IssueTemporaryOperationToken(operation))
+                    .ValueOrThrow();
+                // NB: This modifies secure vault contents.
+                operation->SetTemporaryToken(temporaryToken.Token);
+                operation->SetTemporaryTokenNodeId(temporaryToken.NodeId);
```
Kontakter, line 2677: Maybe join this methods into one?

yt/yt/server/scheduler/scheduler.cpp
```diff
@@ -2659,6 +2660,31 @@ class TScheduler::TImpl
             return;
         }
 
+        if (operation->Spec()->IssueTemporaryToken) {
+            try {
+                // We check this first, so that we don't create a token in vain.
+                auto secureVault = operation->GetSecureVault();
+                if (secureVault && secureVault->FindChild(operation->Spec()->TemporaryTokenEnvironmentVariableName)) {
+                    THROW_ERROR_EXCEPTION(
```
Kontakter, line 2668: I think it is better to check this before (in `StartOperation` or even inside spec postprocessing)

yt/yt/server/scheduler/operations_cleaner.h
```diff
@@ -11,9 +11,18 @@ namespace NYT::NScheduler {
...
+struct TRemoveOperationRequest
 {
     TOperationId Id;
+    //! It is guaranteed that all dependent nodes will be successfully
+    //! removed before the operation node is removed itself.
+    std::vector<NCypressClient::TNodeId> DependentNodeIds;
+};
+
+struct TArchiveOperationRequest
+    : public TRemoveOperationRequest
```
omgronny, line 23: Archiving and removing an operation are two different steps of the cleaner, but here we mix them here and I'm not sure this is a good idea

I would suggest to name the base struct something like `TCLeanOperationRequestBase` or just `TCLeanOperationRequest`, and `TRemoveOperationRequest` and `TArchiveOperationRequest` will inherit it

yt/yt/server/scheduler/operations_cleaner.cpp
```diff
@@ -835,6 +839,10 @@ class TOperationsCleaner::TImpl
         result.TaskNames = attributes.FindYson("task_names");
         result.ControllerFeatures = attributes.FindYson("controller_features");
 
+        if (auto temporaryTokenNodeId = attributes.Find<TNodeId>("temporary_token_node_id")) {
+            result.DependentNodeIds = std::vector{*temporaryTokenNodeId};
```
omgronny, line 843: afaik, just `result.DependentNodeIds = {*temporaryTokenNodeId};` will be more conventional

yt/yt/server/scheduler/operations_cleaner.cpp
```diff
@@ -682,15 +684,15 @@ class TOperationsCleaner::TImpl
...
         if (!IsEnabled()) {
             return;
         }
 
-        GetCancelableInvoker()->Invoke(BIND([this, this_ = MakeStrong(this), operationIds = std::move(operationIds)] {
-            for (auto operationId : operationIds) {
-                EnqueueForRemoval(operationId);
+        GetCancelableInvoker()->Invoke(BIND([this, this_ = MakeStrong(this), requests = std::move(requests)] {
+            for (auto request : requests) {
+                EnqueueForRemoval(request);
```
omgronny, line 695: std::move

yt/yt/server/scheduler/operations_cleaner.cpp
```diff
@@ -1184,13 +1196,13 @@ class TOperationsCleaner::TImpl
...
-    void EnqueueForRemoval(TOperationId operationId)
+    void EnqueueForRemoval(TRemoveOperationRequest request)
     {
         VERIFY_INVOKER_AFFINITY(GetUncancelableInvoker());
 
-        YT_LOG_DEBUG("Operation enqueued for removal (OperationId: %v)", operationId);
+        YT_LOG_DEBUG("Operation enqueued for removal (OperationId: %v, DependentNodeIds: %v)", request.Id, request.DependentNodeIds);
         RemovePending_++;
-        RemoveBatcher_->Enqueue(operationId);
+        RemoveBatcher_->Enqueue(request);
```
omgronny, line 1205: std::move

yt/yt/server/scheduler/operations_cleaner.cpp
```diff
@@ -1487,44 +1510,130 @@ class TOperationsCleaner::TImpl
                         }
                     }
 
-                    auto operationId = operationIds[index];
+                    auto removeOperationRequest = requests[index];
```
omgronny, line 1513: auto&?

omgronny (changes requested):

yt/yt/server/scheduler/operations_cleaner.h
achulkov2: If I do as you say, I will still have to switch to `TCleanOperationRequest` very soon (like RemoveBatcher will have to store `TCleanOperationRequest`), because you can submit an operation for archivation just by operation id => there is a fetch and we build `TArchiveOperationRequest` from attributes => we have to remove this operation afterwards => we need to get `TRemove/CleanOperationRequest` from `TArchiveOperaitonRequest` and pass it to the methods that actually do the removal.

So I can't really see the point? In our paradigm archivation implies removal afterwards.

yt/yt/server/scheduler/scheduler.cpp
achulkov2: Ok, moved the set into the previous method with an extra parameter. Do not want to introduce dependencies on `TIssueTokenResult` into `operation.h` though, it would be weird.

yt/yt/server/scheduler/scheduler.cpp
achulkov2: Moved into spec postprocessing, it is much better, thanks. We also have extra environment variable name validation that I missed this way.

achulkov2:
Addressed comments, rebased and attempted to fix flapping test (ran it for a few hours locally afterwards, didn't fail).

Kontakter:
Shipped!

PR #1007, opened 2024-12-30
Parse docker image with digest

koct9i:
Full scheme is [REGISTRY/]IMAGE[:TAG][@DIGEST]

Digest refers to image manifest of image index and have higher priority
than tag. Unlike to tag image referenced by digest never changes because
of its nature.

Parsing in controller-agent will be used later for linking docker image
from external registry with artifacts stored within cluster.

Link: https://github.com/distribution/reference/blob/main/reference.go
Link: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#definitions
Link: https://github.com/opencontainers/image-spec/blob/main/descriptor.md#digests


Kontakter:
Could you please fix the changelog entry description

koct9i:
Ok, I'll just removed it. This isn't that important.
(and I'm too lazy to figure out what the hell is this)

sergey-v-galtsev:
Shipped!

PR #1012, opened 2025-01-02
Fix concat with append for tables with schema

dmi-feo:
# Remove changelog entry if there is no need to mention it in release notes.

<Message NOT for release notes>

* Changelog entry
Type: fix
Component: python-sdk

<Message for release notes>


dmi-feo:
Doesn't work if input tables have incompatible schemas.
https://github.com/ytsaurus/ytsaurus/issues/1015

PR #1016, opened 2025-01-07
Revive operations_snapshots check

dmi-feo:
- It didn't work with operation aliases (trying to access cypress path with `*`, which is forbidden: https://github.com/ytsaurus/ytsaurus/blob/e7eea1ed49fb49d12d64639161e33943a4e06fa4/yt/yt/core/ytree/node_detail.cpp#L264)
- And it couldn't work at all since operations in `//sys/operations` are stored in dirs with prefixes

* Changelog entry
Type: fix
Component: odin

Odin check `operations_snapshots` was brought back to life.

### DIFF
--- a/yt/odin/checks/bin/operations_snapshots/__main__.py
+++ b/yt/odin/checks/bin/operations_snapshots/__main__.py
@@ -1,18 +1,9 @@
-from yt_odin_checks.lib.check_runner import main
+from datetime import datetime
 
 from yt.common import date_string_to_datetime, YT_DATETIME_FORMAT_STRING
 import yt.wrapper as yt
 
-from datetime import datetime
-
-
-def process_response(rsp):
-    if rsp.is_ok():
-        return rsp.get_result()
-
-    error = yt.YtResponseError(rsp.get_error())
-    if not error.is_resolve_error():
-        raise error
+from yt_odin_checks.lib.check_runner import main
 
 
 def run_check(yt_client, logger, options, states):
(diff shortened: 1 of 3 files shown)


yt/odin/checks/bin/operations_snapshots/__main__.py
```diff
@@ -23,22 +14,37 @@ def run_check(yt_client, logger, options, states):
...
-        events = process_response(responses[(operation, "events")])
-        last_successful_snapshot_time = process_response(
-            responses[(operation, "last_successful_snapshot_time")])
-
-        if events is None or last_successful_snapshot_time is None:
+        op_resp = responses[operation]
+
+        if not op_resp.is_ok():
+            error = yt.YtResponseError(op_resp.get_error())
+            if not error.is_resolve_error():
```
Kontakter, line 35: Let inverse this condition

yt/odin/checks/bin/operations_snapshots/__main__.py
```diff
@@ -23,22 +14,37 @@ def run_check(yt_client, logger, options, states):
...
+                raise error
+            else:
+                logger.warning("Got error resolving %s: %s", operation, error)
```
Kontakter, line 38: `Error resolving path for operation (operation_id: %s, error: %s)`

And I am not sure that it is necessary to log since it is normal situation that operation finished between list request and later get request

yt/odin/checks/bin/operations_snapshots/__main__.py
```diff
@@ -23,22 +14,37 @@ def run_check(yt_client, logger, options, states):
...
+                continue
+
+        assert op_resp.is_ok()
```
Kontakter, line 41: We just checked that it is ok, so this assert is unnecessary; we normally use `assert` for non-trivial invariants

yt/odin/checks/bin/operations_snapshots/__main__.py
```diff
@@ -23,22 +14,37 @@ def run_check(yt_client, logger, options, states):
...
+        result = op_resp.get_result()
+        try:
+            events = result["events"]
+            last_successful_snapshot_time = result["progress"]["last_successful_snapshot_time"]
+        except KeyError as e:
```
Kontakter, line 46: `e` -> `err`

yt/odin/checks/bin/operations_snapshots/__main__.py
```diff
@@ -61,11 +67,9 @@ def run_check(yt_client, logger, options, states):
 
     logger.info("Number of long running operations without built snapshots: %d", len(operations_without_snapshots))
     if operations_without_snapshots:
+        print(operations_without_snapshots)
```
Kontakter, line 70: Use logging instead of print

yt/odin/checks/bin/operations_snapshots/__main__.py
```diff
@@ -23,22 +14,37 @@ def run_check(yt_client, logger, options, states):
...
+                logger.warning("Got error resolving %s: %s", operation, error)
+                continue
+
+        assert op_resp.is_ok()
+        result = op_resp.get_result()
+        try:
+            events = result["events"]
+            last_successful_snapshot_time = result["progress"]["last_successful_snapshot_time"]
+        except KeyError as e:
+            logger.warning("Failed to get operation info for %s: %s", operation, e)
```
Kontakter, line 47: The same issue with logging message format as above

yt/odin/checks/bin/operations_snapshots/__main__.py
dmi-feo: Just to highlight that at this stage we are dealing with a successful response. But okay, I'll delete it

yt/odin/checks/bin/operations_snapshots/__main__.py
dmi-feo: this line never meant to be there...

Kontakter (approved):

Kontakter:
@dmi-feo – could I also ask you to add a simple test on this check? (something similar to existing tests here https://github.com/ytsaurus/ytsaurus/tree/main/yt/odin/checks/tests)

Kontakter:
Shipped!

dmi-feo:
btw, it seems that https://github.com/ytsaurus/ytsaurus/blob/50428eef70a3139306a90e0ebc3d84b5b4fd7717/yt/odin/checks/tests/canondata/result.json is not used in opensource, but blocks ya make tests, let's delete it maybe?

Kontakter:
> btw, it seems that https://github.com/ytsaurus/ytsaurus/blob/50428eef70a3139306a90e0ebc3d84b5b4fd7717/yt/odin/checks/tests/canondata/result.json is not used in opensource, but blocks ya make tests, let's delete it maybe?

For sure, looks like useless stuff, I'll delete it.

PR #1019, opened 2025-01-10
Allow sending/reading values larger than 16MB to RPC proxy via wire protocol to/from methods dealing with static tables.

faucct:
Historically, it data written through RPC proxies had pass dynamic table row validation standards, like max 16MB value size. This is inconvenient for static tables.
SPYT needs to write large values – https://github.com/ytsaurus/ytsaurus-spyt/issues/43. 
Disable wire input Max*ValueLength validation on RPC proxies.
The methods affected by this change are `WriteTable` and `ReadShuffleData` in rpc proxy server and TTableReader in rpc proxy client.

* Changelog entry
Type: feature
Component: proxy

Allow sending/reading values larger than 16MB to RPC proxy via wire protocol to/from methods dealing with static tables.

### DIFF
--- a/yt/yt/client/api/rpc_proxy/row_batch_reader.cpp
+++ b/yt/yt/client/api/rpc_proxy/row_batch_reader.cpp
@@ -18,7 +18,11 @@ TRowBatchReader::TRowBatchReader(
     IAsyncZeroCopyInputStreamPtr underlying,
     bool isStreamWithStatistics)
     : Underlying_(std::move(underlying))
-    , Decoder_(CreateWireRowStreamDecoder(NameTable_))
+    , Decoder_(CreateWireRowStreamDecoder(NameTable_, TWireProtocolOptions {
+        .MaxStringValueLength = std::numeric_limits<i64>::max(),
+        .MaxAnyValueLength = std::numeric_limits<i64>::max(),
+        .MaxCompositeValueLength = std::numeric_limits<i64>::max(),
+    }))
     , IsStreamWithStatistics_(isStreamWithStatistics)
 {
     YT_VERIFY(Underlying_);
--- a/yt/yt/client/api/rpc_proxy/row_batch_reader.cpp
+++ b/yt/yt/client/api/rpc_proxy/row_batch_reader.cpp
@@ -18,11 +18,7 @@ TRowBatchReader::TRowBatchReader(
     IAsyncZeroCopyInputStreamPtr underlying,
     bool isStreamWithStatistics)
     : Underlying_(std::move(underlying))
-    , Decoder_(CreateWireRowStreamDecoder(NameTable_, TWireProtocolOptions {
-        .MaxStringValueLength = std::numeric_limits<i64>::max(),
-        .MaxAnyValueLength = std::numeric_limits<i64>::max(),
-        .MaxCompositeValueLength = std::numeric_limits<i64>::max(),
-    }))
+    , Decoder_(CreateWireRowStreamDecoder(NameTable_, CreateUnlimitedWireProtocolOptions()))
     , IsStreamWithStatistics_(isStreamWithStatistics)
 {
     YT_VERIFY(Underlying_);
--- a/yt/yt/server/lib/rpc_proxy/api_service.cpp
+++ b/yt/yt/server/lib/rpc_proxy/api_service.cpp
@@ -329,7 +329,13 @@ IRowStreamDecoderPtr CreateRowStreamDecoder(
 {
     switch (rowsetFormat) {
         case NApi::NRpcProxy::NProto::RF_YT_WIRE:
-            return CreateWireRowStreamDecoder(std::move(nameTable));
+        {
+            TWireProtocolOptions wireProtocolOptions;
+            wireProtocolOptions.MaxStringValueLength = std::numeric_limits<i64>::max();
+            wireProtocolOptions.MaxAnyValueLength = std::numeric_limits<i64>::max();
+            wireProtocolOptions.MaxCompositeValueLength = std::numeric_limits<i64>::max();
+            return CreateWireRowStreamDecoder(std::move(nameTable), wireProtocolOptions);
+        }
 
         case NApi::NRpcProxy::NProto::RF_ARROW:
             return CreateArrowRowStreamDecoder(std::move(schema), std::move(nameTable));
--- a/yt/yt/server/lib/rpc_proxy/api_service.cpp
+++ b/yt/yt/server/lib/rpc_proxy/api_service.cpp
@@ -329,13 +329,7 @@ IRowStreamDecoderPtr CreateRowStreamDecoder(
 {
     switch (rowsetFormat) {
         case NApi::NRpcProxy::NProto::RF_YT_WIRE:
-        {
-            TWireProtocolOptions wireProtocolOptions;
-            wireProtocolOptions.MaxStringValueLength = std::numeric_limits<i64>::max();
-            wireProtocolOptions.MaxAnyValueLength = std::numeric_limits<i64>::max();
-            wireProtocolOptions.MaxCompositeValueLength = std::numeric_limits<i64>::max();
-            return CreateWireRowStreamDecoder(std::move(nameTable), wireProtocolOptions);
-        }
+            return CreateWireRowStreamDecoder(std::move(nameTable), CreateUnlimitedWireProtocolOptions());
 
         case NApi::NRpcProxy::NProto::RF_ARROW:
             return CreateArrowRowStreamDecoder(std::move(schema), std::move(nameTable));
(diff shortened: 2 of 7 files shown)


achulkov2:
@faucct I see you force-pushed a few days ago.
1. Is this PR ready for review now, or are there still some things to do? 
2. Let's write a better description in the PR, with some text for changelog. Something like "Disable wire input Max*ValueLength validation on RPC proxies."  And a more verbose description just for the PR. 
3. Remind me, are you removing limits in all places where WireProtocolReader is used, or did you leave some be? If there are some remaining, let's list them in the description along with reasons.
4. Let's add test for dynamic tables that it is still impossible to write a large value into dynamic tables via RPC proxies (limit should be 16 megabytes from validation on wire output). Please also test different atomicity (!) modes for these writes into dynamic tables, you can find examples by looking for `"atomicity": "none"/"full"` in integration tests.

All in all, this is a somewhat scary change, because removing validation in a big system is always a bit scary, so let's work on making the change as verifiable as possible.

achulkov2 (changes requested):
(extra comment just to change status to Changes Requested, actual requests in the comment above)

faucct:
1. It is ready: I was making sure the tests work.
3. There are a lot of places where `WireProtocolReader` is used. I have only touched the ones necessary for my test to work. Is it necessary to list all of them in the method description?
4. I have added the dynamic table writing tests.

achulkov2:
> There are a lot of places where WireProtocolReader is used. I have only touched the ones necessary for my test to work. Is it necessary to list all of them in the method description?

I mean the ones in rpc proxy, not across the whole codebase. There are just a few in `*rpc_proxy*`.

yt/yt/tests/integration/proxies/test_rpc_proxy.py
```diff
@@ -853,6 +853,29 @@ def test_sort(self):
...
+    @authors("faucct")
+    @pytest.mark.parametrize("atomicity", ["none", "full"])
+    def test_writing_large_dynamic_rows(self, atomicity):
+        data = [{"index": 0, "str": "F" * (17 << 20)}]
+        create("table", "//tmp/dynamic", attributes={"dynamic": True, "schema": self._schema, "atomicity": atomicity})
+        sync_create_cells(1)
+        sync_mount_table("//tmp/dynamic")
+        try:
+            insert_rows("//tmp/dynamic", data, table_writer={"max_row_weight": 20 << 20})
+        except YtResponseError as err:
```
achulkov2, line 872: you can use `raises_yt_error` helper

yt/yt/server/lib/rpc_proxy/api_service.cpp
```diff
@@ -328,7 +328,13 @@ IRowStreamDecoderPtr CreateRowStreamDecoder(
 {
     switch (rowsetFormat) {
         case NApi::NRpcProxy::NProto::RF_YT_WIRE:
-            return CreateWireRowStreamDecoder(std::move(nameTable));
+        {
```
achulkov2, line 331: Let's add some helper for filling options with max near the definition of `TWireProtocolOptions`. Like `CreateUnlimitedWireProtocolOptions` or a class static method.

achulkov2 (changes requested):

savnadya:
Shipped!

yt/yt/client/api/rpc_proxy/row_batch_reader.cpp
```diff
@@ -18,7 +18,11 @@ TRowBatchReader::TRowBatchReader(
     IAsyncZeroCopyInputStreamPtr underlying,
     bool isStreamWithStatistics)
     : Underlying_(std::move(underlying))
-    , Decoder_(CreateWireRowStreamDecoder(NameTable_))
+    , Decoder_(CreateWireRowStreamDecoder(NameTable_, TWireProtocolOptions {
+        .MaxStringValueLength = std::numeric_limits<i64>::max(),
+        .MaxAnyValueLength = std::numeric_limits<i64>::max(),
+        .MaxCompositeValueLength = std::numeric_limits<i64>::max(),
```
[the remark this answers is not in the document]
achulkov2: This is reading, not writing. It should not even be possible to send huge proto messages, since they are simply invalid. The idea is that we do not want to validate any `TUnversionedRow/TVersionedRow` constraints at the stage of parsing input from RPC client, we should only check conformance with the wire format. Validating limits should be the responsibility of the method implementation within the proxy (in the native client, for example).

yt/yt/tests/integration/proxies/test_rpc_proxy.py
```diff
@@ -853,6 +853,29 @@ def test_sort(self):
 
         assert select_rows("* from [//tmp/t_out] LIMIT " + str(2 * size)) == original_table
 
+    @authors("faucct")
+    def test_writing_large_rows(self):
+        data = [{"index": 0, "str": "F" * (17 << 20)}]
+        create("table", "//tmp/table", attributes={"dynamic": False, "schema": self._schema})
+        write_table("//tmp/table", data, table_writer={"max_row_weight": 20 << 20})
```
achulkov2, line 860: Let's also test the static table limits, that they are cut off at 128M. My only concern is how long a test with 128M + eps of data will run, but I have no ideas on how to improve it yet.

achulkov2 (changes requested):

achulkov2:
@faucct Can you format the description according to our preset, with a general description and then a changelog entry? Like  I do [here](https://github.com/ytsaurus/ytsaurus/pull/1054) (but you don't have to write so much text).

yt/yt/tests/integration/proxies/test_rpc_proxy.py
```diff
@@ -853,6 +853,36 @@ def test_sort(self):
...
+
+    @authors("faucct")
+    def test_writing_large_rows_129(self):
+        data = [{"index": 0, "str": "F" * (129 << 20)}]
+        create("table", "//tmp/table", attributes={"dynamic": False, "schema": self._schema})
+        with raises_yt_error(f'Expected <= {128 << 20}, found {130 << 20}'):
+            write_table("//tmp/table", data, table_writer={"max_row_weight": 130 << 20})
+        with raises_yt_error() as e:
+            write_table("//tmp/table", data, table_writer={"max_row_weight": 128 << 20})
+        assert 'Memory limit exceeded while parsing YSON stream: ' in str(e[0]) or 'Row weight is too large' in str(e[0])
```
achulkov2, line 871: were you able to figure out what causes one or the other error to trigger? from the looks of the code we should hit one of them first, and that the memory limit in yson parser is pretty strict

but since both errors are from server code I guess it is fine..

achulkov2 (changes requested):
small stuff, generally ok

yt/yt/tests/integration/proxies/test_rpc_proxy.py
faucct: No, I wasn't. Most of the time I was getting the second error, but got the first error once and did not want to risk the test flapping.

yt/yt/tests/integration/proxies/test_rpc_proxy.py
```diff
@@ -853,6 +853,39 @@ def test_sort(self):
 
         assert select_rows("* from [//tmp/t_out] LIMIT " + str(2 * size)) == original_table
 
+    @authors("faucct")
+    def test_writing_large_rows(self):
+        MB = 1 << 20
```
Kontakter, line 858: We usually write it as `1024 ** 2`

Kontakter:
Shipped!

Kontakter:
(shipped to run tests in internal CI)

Kontakter:
Everything is ok in internal CI

yt/yt/tests/integration/proxies/test_rpc_proxy.py
```diff
@@ -853,6 +853,39 @@ def test_sort(self):
 
         assert select_rows("* from [//tmp/t_out] LIMIT " + str(2 * size)) == original_table
 
+    @authors("faucct")
+    def test_writing_large_rows(self):
```
achulkov2, line 857: Noticed that this test is in `TestOperationsRpcProxy`, I don't think that is appropriate, operations = MR. Let's either move it somewhere else, or make a separate suite inherited from TestRpcProxyBase. Like TestWireProtocolValidation.

P.S. You are writing large values here (though rows are also large), maybe test_writing_large_values is a better name.

achulkov2 (changes requested):

achulkov2 (changes requested):
Found this decoder we are also creating: https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/server/lib/rpc_proxy/api_service.cpp#L6869. 
I think we don't need to check limits here as well. Let's call someone responsible to double check. @apollo1321

apollo1321:
Yes, there is no need to check limits in `WriteShuffleData` method.

faucct:
merge?

PR #1035, opened 2025-01-16
Add docker image option in operaion spec in cpp sdk

jasminetealover:
* Changelog entry
Type: fix
Component: cpp-sdk

Add `docker image` option into cpp sdk.


hdnpth:
Shipped!

hdnpth:
Shipped!

Issue #1044, opened 2025-01-22
Job should store all node addresses


Krock21:
Currently, job stores a single node address. (`//sys/operations_archive/jobs`, in controller agent, etc)

Scheduler and master store address map for exec nodes (network -> host:port), while controller agent stores and uses a single address

This single address may be unreachable from nodes who prefer an address from a different network. We need to store all addresses in jobs, so correct one can be chosen using network preferences

PR #1052, opened 2025-01-28
Add "addresses" to job description. Use it in get-job-stderr and get-job-spec

Krock21:
Issue: #1044

Added "addresses" field to job description. It has consequences on operations archive, `get-job`, etc

Main motivation behind it is to make `get-job-stderr` and `get-job-spec` work for jobs that run on exec nodes with non-default preferred address (remote exec nodes)

Version in this PR is tested locally. Publishing it early to run tests and get early feedback

TODO:
- Add tests for remote exec nodes
- Adjust go code (affects rpc proxy)

---

* Changelog entry
Type: fix
Component: map-reduce

Added "addresses" field to job description. Fixed get-job-stderr and get-job-spec for exec nodes with non-default preferred addresses

### DIFF
--- a/yt/go/proto/client/node_tracker_client/ya.make
+++ b/yt/go/proto/client/node_tracker_client/ya.make
@@ -1,0 +1,20 @@
+PROTO_LIBRARY()
+
+ONLY_TAGS(GO_PROTO)
+
+PEERDIR(
+    yt/go/proto/core/misc
+)
+
+PROTO_ADDINCL(
+    GLOBAL
+    yt
+)
+
+SRCS(
+    # TODO(aleksandr.gaev): Adding this file produces an error `node_directory.pb.go:188:2: undefined: file_yt_proto_yt_client_node_tracker_client_proto_node_proto_init`.
+    # ${ARCADIA_ROOT}/yt/yt_proto/yt/client/node_tracker_client/proto/node_directory.proto
+    ${ARCADIA_ROOT}/yt/yt_proto/yt/client/node_tracker_client/proto/node.proto
+)
+
+END()
--- a/yt/python/yt/environment/init_operations_archive.py
+++ b/yt/python/yt/environment/init_operations_archive.py
@@ -688,7 +688,6 @@ TRANSFORMS[55] = [
             })),
 ]
 
-
 TRANSFORMS[56] = [
     Conversion(
         "jobs",
@@ -744,6 +743,64 @@ TRANSFORMS[56] = [
                 "account": OPERATIONS_ARCHIVE_ACCOUNT_NAME,
             })),
 ]
+
+TRANSFORMS[57] = [
+    Conversion(
+        "jobs",
+        table_info=TableInfo(
+            [
+                ("job_id_partition_hash", "uint64", "farm_hash(job_id_hi, job_id_lo) % {}".format(JOB_TABLE_PARTITION_COUNT)),
+                ("operation_id_hash", "uint64", "farm_hash(operation_id_hi, operation_id_lo)"),
+                ("operation_id_hi", "uint64"),
+                ("operation_id_lo", "uint64"),
+                ("job_id_hi", "uint64"),
+                ("job_id_lo", "uint64")
+            ], [
+                ("type", "string"),
+                ("state", "string"),
+                ("start_time", "int64"),
+                ("finish_time", "int64"),
+                ("address", "string"),
+                ("error", "any"),
+                ("interruption_info", "any"),
+                ("statistics", "any"),
+                ("stderr_size", "uint64"),
+                ("spec", "string"),
+                ("spec_version", "int64"),
+                ("has_spec", "boolean"),
+                ("has_fail_context", "boolean"),
+                ("fail_context_size", "uint64"),
+                ("events", "any"),
+                ("transient_state", "string"),
+                ("update_time", "int64"),
+                ("core_infos", "any"),
+                ("job_competition_id", "string"),
+                ("has_competitors", "boolean"),
+                ("exec_attributes", "any"),
+                ("task_name", "string"),
+                ("statistics_lz4", "string"),
+                ("brief_statistics", "any"),
+                ("pool_tree", "string"),
+                ("monitoring_descriptor", "string"),
+                ("probing_job_competition_id", "string"),
+                ("has_probing_competitors", "boolean"),
+                ("job_cookie", "int64"),
+                ("controller_state", "string"),
+                ("archive_features", "any"),
+                ("$ttl", "uint64"),
+                ("operation_incarnation", "string"),
+                ("allocation_id_hi", "uint64"),
+                ("allocation_id_lo", "uint64"),
+                ("addresses", "any"),
+            ],
+            default_lock="operations_cleaner",
+            attributes={
+                "atomicity": "none",
+                "tablet_cell_bundle": SYS_BUNDLE_NAME,
+                "account": OPERATIONS_ARCHIVE_ACCOUNT_NAME,
+            })),
+]
+
 
 # NB(renadeen): don't forget to update min_required_archive_version at yt/yt/server/lib/scheduler/config.cpp
 
--- a/yt/yt/server/lib/misc/estimate_size_helpers-inl.h
+++ b/yt/yt/server/lib/misc/estimate_size_helpers-inl.h
@@ -28,6 +28,12 @@ size_t EstimateSize(const std::optional<T>& v)
     return v ? EstimateSize(*v) : 0;
 }
 
+template <typename T1, typename T2>
+size_t EstimateSize(const std::pair<T1, T2>& v)
+{
+    return EstimateSize(v.first) + EstimateSize(v.second);
+}
+
 template <typename C>
     requires std::ranges::range<C>
 size_t EstimateSize(const C& value)
--- a/yt/yt/server/lib/misc/job_reporter.cpp
+++ b/yt/yt/server/lib/misc/job_reporter.cpp
@@ -83,6 +83,7 @@ public:
             Report_.FinishTime(),
             /*updateTime*/ TInstant::Now().MicroSeconds(),
             Report_.Address(),
+            Report_.Addresses(),
             Report_.StderrSize(),
             Report_.HasCompetitors(),
             Report_.HasProbingCompetitors(),
@@ -188,6 +189,7 @@ public:
         if (archiveVersion >= 53 && Report_.Ttl()) {
             record.Ttl = Report_.Ttl()->MilliSeconds();
         }
+
         // COMPAT(eshcherbin)
         if (archiveVersion >= 55 && Report_.OperationIncarnation()) {
             record.OperationIncarnation = Report_.OperationIncarnation();
@@ -197,6 +199,11 @@ public:
             auto allocationIdAsGuid = Report_.AllocationId().Underlying();
             record.AllocationIdHi = allocationIdAsGuid.Parts64[0];
             record.AllocationIdLo = allocationIdAsGuid.Parts64[1];
+        }
+
+        // COMPAT(aleksandr.gaev)
+        if (archiveVersion >= 57 && Report_.Addresses()) {
+            record.Addresses = ConvertToYsonString(*Report_.Addresses());
         }
 
         return FromRecord(record);
--- a/yt/yt/server/lib/scheduler/exec_node_descriptor.cpp
+++ b/yt/yt/server/lib/scheduler/exec_node_descriptor.cpp
@@ -1,11 +1,14 @@
 #include "exec_node_descriptor.h"
 
+#include <yt/yt/ytlib/controller_agent/serialize.h>
+
 #include <yt/yt/ytlib/node_tracker_client/helpers.h>
 
 #include <yt/yt/ytlib/scheduler/disk_resources.h>
 #include <yt/yt/ytlib/scheduler/job_resources_helpers.h>
 
 #include <yt/yt/core/misc/protobuf_helpers.h>
+
 
 namespace NYT::NScheduler {
 
@@ -21,6 +24,7 @@ using NYT::ToProto;
 TExecNodeDescriptor::TExecNodeDescriptor(
     NNodeTrackerClient::TNodeId id,
     const std::string& address,
+    const NNodeTrackerClient::TAddressMap& addresses,
     const std::optional<std::string>& dataCenter,
     double ioWeight,
     bool online,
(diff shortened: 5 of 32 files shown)


eshcherbin:
I will try to review this PR during the week.

yt/yt/ytlib/api/native/client_job_info_impl.cpp
```diff
@@ -295,12 +297,31 @@ TErrorOr<IChannelPtr> TClient::TryCreateChannelToJobNode(
...
+        // COMPAT(aleksandr.gaev) Remove "address" after 25.2
+        options.Attributes = {"address", "addresses"};
         // TODO(ignat): support structured return value in GetJob.
         auto jobYsonString = WaitFor(GetJob(operationId, jobId, options))
             .ValueOrThrow();
-        auto address = ConvertToNode(jobYsonString)->AsMap()->GetChildValueOrThrow<TString>("address");
-        return ChannelFactory_->CreateChannel(address);
+        auto jobYsonMap = ConvertToNode(jobYsonString)->AsMap();
+        if (jobYsonMap->FindChild("addresses")) {
+            auto addresses = jobYsonMap->GetChildValueOrThrow<TAddressMap>("addresses");
```
eshcherbin, line 307: `if (auto addresses = jobYsonMap->FindChildValue("addresses"))`

yt/yt/ytlib/api/native/client_job_info_impl.cpp
```diff
@@ -1450,11 +1471,15 @@ static std::vector<TJob> ParseJobsFromArchiveResponse(
...
         }
 
-        if (responseIdMapping.Address) {
+        if (responseIdMapping.Address) { // COMPAT(aleksandr.gaev) Remove after 25.2
             // This field previously was non-optional.
             job.Address = record.Address.value_or("");
         }
 
+        if (responseIdMapping.Addresses) {
+            job.Addresses = ConvertTo<TAddressMap>(record.Addresses.value_or(TYsonString()));
```
eshcherbin, line 1480: I doubt that an empty yson string could convert to a map successfully.

Krock21:
Rebased

Krock21:
> 1. In all user-facing parts we cannot just remove the `address` because it would be an incompatible API change, so it would require all users to migrate, and that would be long, if not impossible. So I wouldn't even dare to promise to "remove after 25.2" and instead just settle with returning both options for now.

Agreed. By user-facing parts you mean `get-job` API and other apis who return `address` from a job?
Is `orchid` considered a user-facing part?

> 2\. Conversely, in all internal representations (structs and protocols) there might not be a need for storing both the old and the new field, as the latter contains the former (if we're certain that all address maps have a default address). So there is probably no benefit in "removing  after 25.2" compared to removing it right away.

I think it is safer and easier to do it in 2 steps:
1. Add `addresses` to internal and user-visible parts
2. Remove `address` from internal parts, replacing it in user-visible parts with `GetDefaultAddress(Addresses)`

In terms of reviewing PRs and avoiding mistakes
I also don't know how to remove it from internal parts in a compatible manner, will need help with that

yt/yt/server/lib/scheduler/exec_node_descriptor.cpp
```diff
@@ -53,7 +55,8 @@ void TExecNodeDescriptor::Persist(const TStreamPersistenceContext& context)
     using NYT::Persist;
 
     Persist(context, Id);
-    Persist(context, Address);
+    Persist(context, Address); // COMPAT(aleksandr.gaev) Remove after 25.2
+    Persist(context, Addresses);
```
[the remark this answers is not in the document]
Krock21: Could you please explain how it works?

I will have to write it in any case by some reason(I don't understand why):
```
    if (context.GetVersion() >= ESnapshotVersion::AddAddressesToJob) {
        Persist(context, Addresses);
    }
```

Why if I delete "Address" I wouldn't have to write something like this:
```
    if (context.GetVersion() < ESnapshotVersion::RemoveAddress) {
        Persist(context, Address);
    }
```
And this will not be considered a deletion of "Address" from internal structure

yt/python/yt/environment/init_operations_archive.py
```diff
@@ -634,6 +634,81 @@ def create_operations_archive_account(client):
...
+        "tablet_cell_bundle": SYS_BUNDLE_NAME,
+        "account": OPERATIONS_ARCHIVE_ACCOUNT_NAME,
+    },
+)
+
+TRANSFORMS[55] = [
+    Conversion(
+        "jobs",
+        table_info=JOBS_TABLE_V55,
+        mapper=get_addresses_mapper(JOBS_TABLE_V55),
```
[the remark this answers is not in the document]
Krock21: > I don't find this mapping benefitial as jobs have a quite short lifespan

Examples of long-runnings jobs are CHYT SPYT Jupyt DQ

> Internally, our jobs table is tens of terabytes, so it would also cause a huge unwanted load.
This means you have many jobs running at once, and they will need to be migrated to a different structure

What would you propose to do here? Just add a nulllable empty column, and leave old jobs with empty "addresses"?

yt/yt/server/lib/misc/estimate_size_helpers-inl.h
```diff
@@ -28,6 +28,12 @@ size_t EstimateSize(const std::optional<T>& v)
     return v ? EstimateSize(*v) : 0;
 }
 
+template <typename T1, typename T2>
+size_t EstimateSize(const std::pair<T1, T2>& v)
+{
+    return EstimateSize(v.first) + EstimateSize(v.second);
+}
```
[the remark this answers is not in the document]
Krock21: I failed to do it, some magic code needs to be written for it

If you know how to do it, please tell me

yt/yt/server/lib/scheduler/exec_node_descriptor.cpp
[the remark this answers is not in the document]
Krock21: I have added snapshot version checks
I wasn't able to find snapshot versions for scheduler (seems like we only have them for controller_agent), so I used versions from controller agent. Please check it and tell me what is a correct way

Krock21:
Fixed comments, added a test(it works for rpc proxies, surprisingly)

Kontakter:
Shipped!

Kontakter:
(to run tests in internal CI)

Kontakter:
Some cosmetics:

```
File contains space characters before EOL or EOF
  * yt/yt/tests/integration/scheduler/test_get_job.py:538:1
```

yt/yt/server/lib/scheduler/exec_node_descriptor.cpp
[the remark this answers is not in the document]
eshcherbin: > Why if I delete "Address" I wouldn't have to write something like this

You would have to, because this would change the binary format of the snapshot. Each snapshot version has its own binary layout defined by the order of fields persisted in the code. If you add a new field, you need to skip it when loading a snapshot of some earlier version. The same holds for deleting fields (however we don't delete fields ofter).

The only exception is when we deliberately break compatibility by forbidding loading from too old snapshot version. The lower bound is defined in https://github.com/ytsaurus/ytsaurus/blob/764835faf16d4d5dd99be8fab79008647a2ab483/yt/yt/ytlib/controller_agent/serialize.cpp#L14. But we do that only in major version releases.

yt/yt/server/lib/misc/estimate_size_helpers-inl.h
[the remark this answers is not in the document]
eshcherbin: That's right, there is some magic needed. The easiest way seems to be using `std::apply` with `EstimateSizes`.

Regardless, I've realized that a `std::tuple` specialization would not work for pairs automatically, so no changes necessary here.

yt/yt/ytlib/controller_agent/serialize.cpp
```diff
@@ -15,7 +15,7 @@ bool ValidateSnapshotVersion(int version)
 {
     // NB: Version can be not valid enum value, so we do not cast version to enum here.
     return
-        version >= ToUnderlying(ESnapshotVersion::GroupedNeededResources) &&
+        version >= ToUnderlying(ESnapshotVersion::AddAddressesToJob) &&
```
eshcherbin, line 18: This is unnecessary.

yt/yt/ytlib/api/native/client_job_info_impl.cpp
```diff
@@ -1458,6 +1477,10 @@ static std::vector<TJob> ParseJobsFromArchiveResponse(
             job.Address = record.Address.value_or("");
         }
 
+        if (responseIdMapping.Addresses) {
+            job.Addresses = ConvertTo<TAddressMap>(record.Addresses.value_or(TYsonString(TString("{}"))));
```
eshcherbin, line 1481: To my taste, an explicit `if (record.Addresses)` would be better.

We check `responseIdMapping` to know if the field has been requested in the first place. This is needed for fields that are mandatory, so we must put some default value. In this case, I am not sure we want to make this new field mandatory.

yt/yt/tests/integration/scheduler/test_get_job.py
```diff
@@ -535,6 +535,28 @@ def test_job_archive_ttl(self):
         time.sleep(5)
 
         get_job(op.id, job_id)
+    
+    @authors("aleksandr.gaev")
+    def test_job_addresses(self):
+        op = run_test_vanilla(
+            with_breakpoint("BREAKPOINT"),
+            job_count=1,
```
eshcherbin, line 543: `job_count=1` is the default value.

eshcherbin:
> Agreed. By user-facing parts you mean `get-job` API and other apis who return `address` from a job? Is `orchid` considered a user-facing part?

Yes. `orchid` is also a user-facing part.

> I think it is safer and easier to do it in 2 steps:
> 
> 1. Add `addresses` to internal and user-visible parts
> 2. Remove `address` from internal parts, replacing it in user-visible parts with `GetDefaultAddress(Addresses)`
> 
> In terms of reviewing PRs and avoiding mistakes. Removing `address` also has other challenges around `ListJobs`, that can filter and sort by address
> 
> What do you think?
> 
> I also don't know how to remove it from internal parts in a compatible manner, will need help with that

When reasoning about compatibility we need to think about three things: user-facing APIs, persistent storage and communication protocols. In this case, in user-facing APIs (`get-job`, `list-jobs`, orchid and other) I am strongly for leaving singular `address` everywhere. Regarding persistent storage, we can either stop writing `address` to the archive table, or duplicate it. The latter would require futher careful compatibility logic (including sorting by address). The same is true for writing snapshots, however, in this case this logic would be local and easier to implement.

By "removing singular `address` internally" I mostly mean removing it from `TExecNodeDescriptor` and similar places to avoid unnecessary duplication. I am not against doing it in a separate PR, but I would like to ask you not to leave it for long in this case.

Krock21:
Fixed comments, rebased, added a test for list_jobs and fixed list_jobs for rpc proxy

Krock21:
Rebased

yt/yt/server/lib/misc/job_reporter.cpp
```diff
@@ -199,6 +201,11 @@ class TJobRowlet
             record.AllocationIdLo = allocationIdAsGuid.Parts64[1];
         }
 
+        // COMPAT(aleksandr.gaev)
+        if (archiveVersion >= 56 && Report_.Addresses()) {
```
eshcherbin, line 205: Version should be 57 now.

--
We have a ticket to replace number versions with enums for convenience, but we haven't found time to do it yet.

yt/yt/tests/integration/scheduler/test_get_job.py
```diff
@@ -519,6 +519,27 @@ def test_job_archive_ttl(self):
...
+        wait(lambda: check_addresses(get_job(op.id, job_id)))
+
+        release_breakpoint()
+
+        op.track()
+
+        job_info = get_job(op.id, job_id)
+        assert check_addresses(job_info)
+        assert len(job_info.get("address")) > 0
+        assert len(job_info.get("addresses")) > 0
```
eshcherbin, line 541: I'd suggest checking that `address` is the default address in `addresses`.

yt/yt/ytlib/api/native/client_job_info_impl.cpp
```diff
@@ -1461,6 +1480,10 @@ static std::vector<TJob> ParseJobsFromArchiveResponse(
             job.Address = record.Address.value_or("");
         }
 
+        if (responseIdMapping.Addresses && record.Addresses) {
```
eshcherbin, line 1483: This condition should be just `record.Addresses`. We plan to drop `This field previously was non-optional` compats later for other fields as well.

yt/yt/ytlib/api/native/client_job_info_impl.cpp
```diff
@@ -1869,6 +1901,7 @@ static void ParseJobsFromControllerAgentResponse(
     auto filter = [&] (const INodePtr& jobNode) -> bool {
         const auto& jobMap = jobNode->AsMap();
         auto address = jobMap->GetChildValueOrThrow<TString>("address");
+        auto addresses = jobMap->GetChildValueOrDefault("addresses", TAddressMap());
```
eshcherbin, line 1904: This variable is not used currently.

eshcherbin (changes requested):
Sorry I've taken some time to return to this PR. Only minor changes left, most important of which is the wrong archive version number in a compat.

Krock21:
Rebased and fixed comments

yt/go/proto/client/node_tracker_client/ya.make
```diff
@@ -0,0 +1,20 @@
...
+)
+
+PROTO_ADDINCL(
+    GLOBAL
+    yt
+)
+
+SRCS(
+    # TODO(aleksandr.gaev): Adding this file produces an error `node_directory.pb.go:188:2: undefined: file_yt_proto_yt_client_node_tracker_client_proto_node_proto_init`.
+    # ${ARCADIA_ROOT}/yt/yt_proto/yt/client/node_tracker_client/proto/node_directory.proto
```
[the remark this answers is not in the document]
Krock21: When trying to uncomment it I face these errors while doing `ya make yt/go`: https://gist.github.com/Krock21/126c8b3735c0aa4341da078af836792f

2nd error is after adding `option go_package = "go.ytsaurus.tech/yt/go/proto/client/node_tracker_client";` to `node_directory.proto`, as first error suggests

I don't really know how to fix it and whether it is possible to do so. I need help with that

cc @verytable

Kontakter:
Shipped!

PR #1054, opened 2025-01-28
Allow impersonation in HTTP protocol for superusers

achulkov2:
This PR improves the general approach to user impersonation in our HTTP protocol.
Previously, only for the `yql_agent` user and working by specifying the `X-YT-User-Name`. The user list was hardcoded in the server binary.
Now:
* Similar to the existing ImpersonationUser option in the C++ HTTP client, `impersonation_user` is supported in configuration of go and python clients (using the same header). **UPD: Moved to #1072 and #1073.**
* Specifying impersonation headers is allowed for all users who are members of the `superusers` group, which is checked via a new cache within native connection, similar to permission cache.
* The same cache is used to check for users' banned flag before allowing impersonation, so, if banned, users (including `yql_agent`) will no longer be allowed to perform cluster operations with impersonation.

Whitelist is kept as part of the binary for compatibility purposes, but in the future can be eliminated completely in favour of adding the relevant users to `superusers`. This is pretty reasonable: impersonation allows impersonating as root, which is always a superuser, so we might as well require being a superuser.

* Changelog entry
Type: feature
Component: proxy

Extend impersonation functionality in HTTP protocol. It is now allowed for all unbanned superusers.


yt/yt/ytlib/cypress_client/batch_attribute_fetcher.cpp
```diff
@@ -43,7 +43,12 @@ TBatchAttributeFetcher::TBatchAttributeFetcher(
     , MasterReadOptions_(options)
     , Logger(logger)
 {
-    YT_VERIFY(std::find(AttributeNames_.begin(), AttributeNames_.end(), "type") != AttributeNames_.end());
+    // We need the "type" attribute to distinguish links from regular nodes. If it is not requested, we fetch it, but
+    // later remove it from the externally visible result.
+    if (std::find(AttributeNames_.begin(), AttributeNames_.end(), "type") == AttributeNames_.end()) {
+        AddedTypeAttribute_ = true;
+        AttributeNames_.push_back("type");
```
savnadya, line 50: Let's add a constant with "type" string?

yt/yt/server/http_proxy/http_authenticator.cpp
```diff
@@ -145,15 +151,35 @@ TErrorOr<TAuthenticationResultAndToken> THttpAuthenticator::Authenticate(
             auto authenticationResult = rsp.Value();
 
             if (auto userHeader = request->GetHeaders()->Find(UserNameHeader)) {
-                if (UserImpersonationWhitelist.contains(authenticationResult.Login)) {
+                auto isWhitelisted = UserImpersonationWhitelist.contains(authenticationResult.Login);
+
+                auto isSuperuserOrError = WaitFor(NApi::NNative::IsSuperuser(Bootstrap_->GetNativeConnection(), TString(authenticationResult.Login)));
```
savnadya, line 156: using namespace?

yt/yt/ytlib/api/native/connection.cpp
```diff
@@ -1414,4 +1429,24 @@ TFuture<TTableMountInfoPtr> GetTableMountInfo(const TRichYPath& objectPath, cons
...
+TFuture<bool> IsSuperuser(const IConnectionPtr& connection, const TString& user)
+{
+    return connection->GetUserAttributeCache()->Get(user)
+        .Apply(BIND([] (const TUserAttributesPtr& attributes) {
+            YT_VERIFY(attributes);
+            return attributes->MemberOfClosure.contains(SuperusersGroupName);
+        }));
+}
+
+TFuture<bool> IsUserBanned(const IConnectionPtr &connection, const TString &user)
```
savnadya, line 1441: `const IConnectionPtr &connection` -> `const IConnectionPtr& connection`

yt/yt/server/http_proxy/http_authenticator.cpp
savnadya, line 156: If we have `std::string` here, maybe it would be better to use `std::string` in user attribute cache too?

yt/python/yt/wrapper/default_config.py
```diff
@@ -227,6 +227,13 @@ def get_dynamic_table_retries():
     # Set to yt.wrapper.tvm.ServiceTicketAuth(tvm_client) or yt.wrapper.tvm.UserTicketFixedAuth()
     "tvm_auth": None,
 
+    # This option allows the client to impersonate another user.
+    # Setting this option is only allowed for superusers that are not banned,
+    # all other attempts at impersonation will result in an authorization error.
+    # For now only the HTTP driver supports option, in the RPC driver it is ignored.
+    # If you are using the native driver, use the `driver_user_name` option below instead.
+    "impersonation_user": None,
```
savnadya, line 235: Let's split this PR into three: server part, python and go?

savnadya (changes requested):

yt/python/yt/wrapper/default_config.py
achulkov2: I don't like to merge stuff without tests though. I can separate go, but I want to keep the python part, because the wrapper test is more sophisticated than the integration test, since setting the user at the client level is our main intent. (plus it is annoying to replicate http proxy interaction in integration tests when we this is more of a wrapper feature)

yt/yt/ytlib/api/native/connection.cpp
achulkov2: damn you copilot!

yt/yt/server/http_proxy/http_authenticator.cpp
achulkov2: now this PR only removes `TString` usages!

yt/python/yt/wrapper/default_config.py
achulkov2: Split (after our additional discussion about how it is simpler to cherry-pick sdk stuff if it is separate).

savnadya:
Shipped!

savnadya:
Shipped!

PR #1062, opened 2025-02-05
Support auth via SSO cookie

dmi-feo:


dmi-feo:
I see that some tests have failed, but these failures seem to have nothing to do with this PR

yt/chyt/controller/internal/auth/ythttputil.go
```diff
@@ -63,6 +63,14 @@ func GetCredentials(r *http.Request) (yt.Credentials, error) {
 		return credentials, nil
 	}
 
+	if cookie, err := r.Cookie("access_token"); err == nil {
```
l0kix2, line 66: Is this a request from ui? In that case cookie called `yt_oauth_access_token` and it is passed to proxies as `access_token `

l0kix2: @Flunt1k said that it is cookie from ui backend, not browser, so should be ok when UI will send cookie instead of oauth token to the proxie

PR #1066, opened 2025-02-07
Use yt/core/crypto SSL context for BUS TLS

koct9i:
- yt/core/crypto: refactor SSL context
  
- yt/core/bus: use yt/core/crypto SSL context implementation
  

---
>
> * Changelog entry

* Changelog entry
Type:
Component:

<Your description.>


koct9i:
Reasons:
- unification
- sanitization (singleton ssl context seems insane)
- add support for ECDSA
- optimize initialization at connect (TODO)

psushin:
Shipped!

yalekseev:
Shipped!

PR #1047, opened 2025-01-24
controller agent should fallback to YT_TOKEN when pulling internal docker images

faucct:

PR #1069, opened 2025-02-10
Handle both docker image and layer artifacts in job spec simultaneously

koct9i:
Exec node should choose method for building job workspace using either
docker image or layer tarball artifacts.

"Simple" job environment must report failure if any image/layer is specified.

---
>
> * Changelog entry

* Changelog entry
Type:
Component:

<Your description.>


koct9i:
Exec node counterpart of #1047

k-pogorelov (approved):

PR #1070, opened 2025-02-11
yt/error: optimize constructing TError from TErrorException

koct9i:
Here we can avoid dynamic cast and exception hazard.

---
>
> * Changelog entry


maxim-babenko:
Shipped!

PR #1071, opened 2025-02-13
Extend OAuth login transformations with to_lower and to_upper

achulkov2:
Added `to_lower` and `to_upper` as part of `TStringReplacementConfig`. It is applied after regexp replacement (but before the next transformation). Now you don't have to specify `match_pattern` if one of `to_lower`/`to_upper` is specified.

* Changelog entry
Type: feature
Component: proxy

Add to_lower and to_upper option to OAuth login_transformations.


achulkov2:
And yes, these transformations should ideally be defined by different configs and dispatched depending on transformation type name. I will attempt it tomorrow.

savnadya:
Shipped!

PR #1080, opened 2025-02-17
Add tablet balancer in dockerfile

jasminetealover:
* Changelog entry
Type: feature
Component: tablet-balancer

Add `tablet-balancer` into YTsaurus `Dockerfile`. This also should be added into stable branches.


savnadya:
Shipped!

PR #1085, opened 2025-02-20
[24.1] Add service orchid endpoint in tablet balancer

jasminetealover:
* Changelog entry
Type: fix
Component: tablet-balancer

Add service orchid endpoint. For 24.1 only.


Kontakter:
Shipped!

PR #1088, opened 2025-02-21
updatable scheduling_tag_filter

faucct:
Needed to support adding reserved execution nodes and manually scheduling specific operations to those afterwards.

### DIFF
--- a/yt/yt/server/scheduler/fair_share_tree.cpp
+++ b/yt/yt/server/scheduler/fair_share_tree.cpp
@@ -564,7 +564,7 @@ public:
         YT_ASSERT_INVOKERS_AFFINITY(FeasibleInvokers_);
 
         if (const auto& element = FindOperationElement(operationId)) {
-            element->SetSchedulingTagFilter(schedulingTagFilter);
+            element->SetSchedulingTagFilter(std::move(schedulingTagFilter));
             element->SetRuntimeParameters(runtimeParameters);
         }
     }
--- a/yt/yt/server/scheduler/fair_share_tree.cpp
+++ b/yt/yt/server/scheduler/fair_share_tree.cpp
@@ -558,11 +558,13 @@ public:
 
     void UpdateOperationRuntimeParameters(
         TOperationId operationId,
+        TSchedulingTagFilter schedulingTagFilter,
         const TOperationFairShareTreeRuntimeParametersPtr& runtimeParameters) override
     {
         YT_ASSERT_INVOKERS_AFFINITY(FeasibleInvokers_);
 
         if (const auto& element = FindOperationElement(operationId)) {
+            element->SetSchedulingTagFilter(schedulingTagFilter);
             element->SetRuntimeParameters(runtimeParameters);
         }
     }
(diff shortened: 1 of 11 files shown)


yt/yt/tests/integration/scheduler/test_scheduler_preemption.py
```diff
@@ -256,6 +256,21 @@ def test_graceful_preemption_timeout(self):
         assert op.get_job_count("total") == 1
         assert op.get_job_count("aborted") == 1
 
+    @authors("faucct")
+    def test_update_scheduling_tag_filter(self):
+        update_pool_tree_config("default", {"node_reconnection_timeout": 0})
+        op = run_test_vanilla(command=with_breakpoint("BREAKPOINT ; sleep 100"), spec={"preemption_mode": "graceful"})
+        wait_breakpoint()
+        release_breakpoint()
```
Kontakter, line 264: Why `release_breakpoint` is needed here?

yt/yt/ytlib/scheduler/config.h
```diff
@@ -2189,6 +2191,7 @@ class TOperationFairShareTreeRuntimeParametersUpdate
     TJobResourcesConfigPtr ResourceLimits;
     // Can only be set by an administrator.
     std::optional<bool> EnableDetailedLogs;
+    std::optional<TBooleanFormula> SchedulingTagFilter;
```
Kontakter, line 2194: Do we actually need separate scheduling tag filter values per tree? I think it is not a good idea because this option affects the state of operation controller

faucct: I will try removing it.

yt/yt/ytlib/scheduler/config.h
```diff
@@ -2116,6 +2116,7 @@ class TOperationFairShareTreeRuntimeParameters
     std::optional<double> Weight;
     TPoolName Pool;
     TJobResourcesConfigPtr ResourceLimits;
+    std::optional<TBooleanFormula> SchedulingTagFilter;
```
Kontakter, line 2119: Here in per tree runtime parameters SchedulingTagFilter is still presented

Kontakter: But you can pass scheduling tag filter separately to `TFairShareTree::UpdateOperationRuntimeParameters` and avoid storing it in `TOperationFairShareTreeRuntimeParameters`

yt/yt/server/scheduler/fair_share_tree.cpp
```diff
@@ -558,11 +558,13 @@ class TFairShareTree
 
     void UpdateOperationRuntimeParameters(
         TOperationId operationId,
+        TSchedulingTagFilter schedulingTagFilter,
         const TOperationFairShareTreeRuntimeParametersPtr& runtimeParameters) override
     {
         YT_ASSERT_INVOKERS_AFFINITY(FeasibleInvokers_);
 
         if (const auto& element = FindOperationElement(operationId)) {
+            element->SetSchedulingTagFilter(schedulingTagFilter);
```
Kontakter, line 567: Let add `std::move`

Kontakter (approved):

PR #1096, opened 2025-02-25
Retry CRI container remove if it is starting

koct9i:
Workaround for containerd design flaw.
Match error using regex because prefix matching isn't specific enough.

---
>
> * Changelog entry

* Changelog entry
Type: fix
Component: map-reduce

Retry CRI container remove, workaround for containerd design flaw.


k-pogorelov:
I will review this PR within a couple of days.

k-pogorelov (approved):

PR #1101, opened 2025-03-03
Add controller agent config flag "translate_internal_images_into_layers"

koct9i:
Allows to disable translation from docker image into layers when
- exec nodes are unable to handle them correctly
- internal registry does not provide correct metadata

---
>
> * Changelog entry

<Your description.>

### DIFF
--- a/yt/yt/server/controller_agent/config.cpp
+++ b/yt/yt/server/controller_agent/config.cpp
@@ -699,6 +699,13 @@ void TDockerRegistryConfig::Register(TRegistrar registrar)
         .Default(false);
     registrar.Parameter("forward_internal_images_to_job_specs", &TThis::ForwardInternalImagesToJobSpecs)
         .Default(false);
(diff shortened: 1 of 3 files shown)


koct9i:
This is followup for #1047 and #1069

sergey-v-galtsev:
LGTM

koct9i:
What? There is no "changelog entry" in pr description, in commit message too.

koct9i:
Yes. This is common practice and most projects.
And I already have commits like that in ytsaurus.
Do you have got policy against that?

sergey-v-galtsev:
No, I am just confused by phrases
"Type: ? # fix/feature (Select one value, example: Type: fix)"
and
"# cpp-sdk/go-sdk (Select one value, example: Component: scheduler)"

Should they really be included in the final commit message?

koct9i:
Thanks. We've discussed this in ci/cd/devtools sig chat.

PR #1117, opened 2025-03-06
Extract implementation of read, write and flush methods to TIOEngineBaseCommon

Krock21:
This is a refactoring that will allow to avoid copypasting when introducing new IO Engines

Current schema:
2 drawio

New schema
3 drawio

This will allow reusing read/write implementations between different IOEngines

This will help reduce copypaste in new IOEngines. For example, [this one](https://github.com/ytsaurus/ytsaurus/pull/1078)

### DIFF
--- a/yt/yt/server/lib/io/io_engine.cpp
+++ b/yt/yt/server/lib/io/io_engine.cpp
@@ -27,20 +27,10 @@
 
 #include <array>
 
-#ifdef _linux_
-    #include <sys/uio.h>
-#endif
-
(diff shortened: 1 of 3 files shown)


Krock21:
We discussed with @don-dron and decided to not introduce a new class, but to add these helper methods to io_engine_base and move related configs there. uring and other future io engines will simply not use them

Krock21:
We decided that it is better to make separate static methods that can be used, instead of a class

PR #1123, opened 2025-03-10
Disable fastops in clickhouse for non-x86_64

koct9i:
Fastops is implemented only for x86_64 AVX and AVX2.
"Plain" implementation call "fmath" which is also x86_64-only.


p4vook (changes requested):
Please split the config as in other projects: create a dispatcher `clickhouse_config.h` and redirect it to `config-linux.h`, `config-linux-aarch64.h`, etc depending on macro values.

p4vook:
Fixed in commit https://github.com/ytsaurus/ytsaurus/commit/6804c4d37b778e0f5c229e5e703f80e8816b2d3e, thank you!

PR #1124, opened 2025-03-10
Add generic divideImpl to clickhouse

koct9i:
Fix compilation without SSE2/AVX2, for example for arm64.

Link: https://github.com/ClickHouse/ClickHouse/blob/master/src/Functions/divide/CMakeLists.txt


p4vook (approved):
Shipped!

Will require some fixes on our side, so import may fail at first.

p4vook (approved):
Let's try again...

LGTM!

PR #1133, opened 2025-03-14
Add "internal_registry_alternative_addresses" for alternative "internal_registry_address"

koct9i:
- Add "internal_registry_alternative_addresses" for alternative "internal_registry_address"
  When internal registry has several alternative endpoints or DNS CNAMEs
  all these addresses should be considered as references to internal registry.

- Add registry address for internal images passed to job spec
  Normalize docker image reference passed into job spec.
  This allows to omit registry address for internal docker images in
  operation spec - controller agent will add it to job spec.

### DIFF
--- a/yt/yt/server/controller_agent/config.cpp
+++ b/yt/yt/server/controller_agent/config.cpp
@@ -693,6 +693,8 @@ void TDockerRegistryConfig::Register(TRegistrar registrar)
 {
     registrar.Parameter("internal_registry_address", &TThis::InternalRegistryAddress)
         .Default();
+    registrar.Parameter("internal_registry_alternative_addresses", &TThis::InternalRegistryAlternativeAddresses)
+        .Default();
     registrar.Parameter("use_yt_token_for_internal_registry", &TThis::UseYtTokenForInternalRegistry)
         .Default(false);
     registrar.Parameter("forward_internal_images_to_job_specs", &TThis::ForwardInternalImagesToJobSpecs)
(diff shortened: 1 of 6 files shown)


yt/yt/server/controller_agent/config.cpp
```diff
@@ -706,6 +708,10 @@ void TDockerRegistryConfig::Register(TRegistrar registrar)
         .Default(true);
 
     registrar.Postprocessor([&] (TDockerRegistryConfig* options) {
+        if (!options->InternalRegistryAddress.empty()) {
```
Kontakter, line 711: Checking string on emptiness is a bad practice, let make it optional

koct9i: Why?
FQDN cannot be empty.

sergey-v-galtsev:
LGTM

sergey-v-galtsev:
Tests fail:
```
/storage/d/arcc/mount/1/a85ecbfa-d6c426bd-78cfe7be-95bf68a9/yt/yt/server/controller_agent/unittests/docker_image_ut.cpp:194:
Expected equality of these values:
  image.Registry
    Which is: "internal.registry"
  ""
    Which is: 0x48abd82


/storage/d/arcc/mount/1/a85ecbfa-d6c426bd-78cfe7be-95bf68a9/yt/yt/server/controller_agent/unittests/docker_image_ut.cpp:198:
Expected equality of these values:
  image.IsInternal()
    Which is: false
  true
```

sergey-v-galtsev:
LGTM

PR #1135, opened 2025-03-17
Fixing backtrace for arm64

koct9i:
For now it crashes for arm64 target.


koct9i:
Unwinding stack in fiber does not stop at tramplin. It seems libunwind drawrf unwinder does not stop at NULL frame-pointer because it does not use frame-pointers. Link-register is incorrect at this stage too.

It seems there are some over issues. Work in progress.

koct9i:
"__stacktrampoline" fixes both stack unwinding internally by libunwind and dwarf-based stack unwinding by any external tool.

It seems there are still some issues within sanitizer intergation.

PR #1136, opened 2025-03-17
map_result/sort_result and system_quotas Odin check improvements

kruftik:
* ability to pass primary_medium option for sort_result and map_result checks
* ability to specify default threshold for system_quotas check

### DIFF
--- a/yt/odin/checks/lib/system_quotas/__init__.py
+++ b/yt/odin/checks/lib/system_quotas/__init__.py
@@ -180,6 +180,9 @@ def get_quota_holder_threshold(
 
 def do_run_check(yt_client, logger, states, options, facade, config):
     cluster = get_cluster_name(yt_client)
+
+    default_threshold = options.get("default_threshold", config.DEFAULT_THRESHOLD)
+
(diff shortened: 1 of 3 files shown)


yt/odin/checks/lib/system_quotas/__init__.py
```diff
@@ -180,6 +180,9 @@ def get_threshold_from_schedule(schedule: ScheduleOptionsDict):
 
 def do_run_check(yt_client, logger, states, options, facade, config):
     cluster = get_cluster_name(yt_client)
+
+    default_threshold = options.get("default_threshold", config.DEFAULT_THRESHOLD)
```
Krock21, line 184: nit: default_threshold -> threshold

Krock21 (approved):

yt/odin/checks/lib/system_quotas/__init__.py
kruftik: since it's a value that can be override on an account basis, I think default_threshold is a better name

dim-an (approved):
Shipped!

PR #1139, opened 2025-03-18
Allow TLS 1.3 for HTTPS proxies and bus RPC

koct9i:
That should be done years ago.

* Allow TLS 1.3 for https and bus rps
Type: feature
Component: proxy

Better late than never.

### DIFF
--- a/yt/yt/core/bus/tcp/ssl_context.cpp
+++ b/yt/yt/core/bus/tcp/ssl_context.cpp
@@ -36,7 +36,7 @@ public:
             THROW_ERROR_EXCEPTION("Failed to set min protocol version: %v", GetLastSslErrorString());
         }
 
-        if (SSL_CTX_set_max_proto_version(SslCtx_.get(), TLS1_2_VERSION) != 1) {
+        if (SSL_CTX_set_max_proto_version(SslCtx_.get(), TLS1_3_VERSION) != 1) {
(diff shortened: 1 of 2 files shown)


Kontakter:
Some tests have failed:
```
/storage/d/arcc/mount/1/8edb9854-ed8c69ca-45a38b2d-e025646a/yt/yt/core/bus/unittests/ssl_ut.cpp:774:
Value of: bus->GetReadyFuture().Get().IsOK()
  Actual: true
Expected: false
```

koct9i:
cipher-list affects only TLS 1.2
https://docs.openssl.org/1.1.1/man3/SSL_CTX_set_cipher_list/#description
TLS 1.3 uses diffect concept: cipher-suites

I don't see reason why BUS TLS code even need such config as list of ciphers to support.
It would be better to provide path to openssl config or something like this.

koct9i:
BTW nginx use "ssl_conf_command" which constructs SSL config from text commands.

https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_conf_command
https://docs.openssl.org/1.1.1/man3/SSL_CONF_cmd/

I'll do same approach.

koct9i:
To feed `SSL_CONF_cmd()` config should proved list of string pairs.
I prefer to keep them as vector of structures:

```
ssl_configuration = [
 { command="CipherString"; argument="..."; };
 { command="CipherSuites"; argument="..."; };
];
```

orloffv:
any news here?

PR #1143, opened 2025-03-19
Get rid of SelectRows

Tony-Romanov:


yt/yql/providers/yt/gateway/lib/yt_helpers.cpp
```diff
@@ -369,7 +369,7 @@ static bool IterateRows(NYT::ITransactionPtr tx,
     } else {
         auto format = specsCache.GetSpecs().MakeInputFormat(tableIndex);
         auto rawReader = tx->CreateRawReader(path, format, readerOptions);
-        TMkqlReaderImpl reader(*rawReader, 0, 4 << 10, tableIndex);
+        TMkqlReaderImpl reader(*rawReader, 0, 4 << 10, tableIndex, true);
```
nepal, line 372: Этот код ведь работает для статических таблиц. Мы тут ничего не сломаем?

Tony-Romanov: Да, должен. При union all tableIndex может быть любой. А из потока, при независимом чтении, всегда приходит только 0 и при парсинге использутся чужая спека.  Обычно, при том же union all спеки одинаковые, и я думаю отому этот баг и жив до сих пор.

nepal (approved):

yt/yql/providers/yt/gateway/native/yql_yt_native.cpp
```diff
@@ -3343,30 +3343,21 @@ class TYtNativeGateway : public IYtGateway {
...
-                        if (srcTable.TransactionId_) {
-                            readTx = entry->GetSnapshotTx(*srcTable.TransactionId_);
-                            srcTable.TransactionId_.Clear();
-                        }
-                        if (execCtx->YamrInput) {
-                            stop = NYql::IterateYamredRows(readTx, srcTable, i, specsCache, pullData, limiter, execCtx->Sampling);
-                        } else {
-                            stop = NYql::IterateYsonRows(readTx, srcTable, i, specsCache, pullData, limiter, execCtx->Sampling);
-                        }
+                        stop = NYql::IterateYsonRows(readTx, srcTable, i, specsCache, pullData, limiter, execCtx->Sampling);
```
rvu1024, line 3360: Код на строках https://github.com/ytsaurus/ytsaurus/pull/1143/files#diff-07f5752fb37d36a427246c248df1d50de2e2c800ebcd5a05fdec3a421f1550fdL330-L340 не будет работать с динамическими таблицами. Для теста напиши селект из статичской + динамической, что бы лимит приходился на середину динамической

yt/yql/providers/yt/gateway/native/yql_yt_native.cpp
```diff
@@ -3343,30 +3343,21 @@ class TYtNativeGateway : public IYtGateway {
...
-                    bool isDynamic = execCtx->InputTables_[i].Dynamic;
-                    ui64 recordsCount = execCtx->InputTables_[i].Records;
-                    if (!isDynamic) {
+                    if (const auto recordsCount = execCtx->InputTables_[i].Records) {
                         if (!limiter.NextTable(recordsCount)) {
                             continue;
                         }
-                    } else {
-                        limiter.NextDynamicTable();
                     }
```
rvu1024, line 3350: Для динамических таблиц число записей приблизительное. Тут нельзя просто убирать эту логику. Нужно аккуратно переписать с использованием актуально прочитанных записей

rvu1024 (changes requested):

yt/yql/tests/agent/test_simple.py
```diff
@@ -964,3 +969,33 @@ def test_different_sources(self, query_tracker, yql_agent):
...
+            "schema": [{"name": "a", "type": "int64"}, {"name": "b", "type": "string"}, {"name": "c", "type": "float"}],
+            "enable_dynamic_store_read": True,
+        })
+        write_table("//tmp/t2", [{"a": 45, "b": "bar", "c": -3.0}, {"a": 46, "b": "abc", "c": -4.0}, {"a": 47, "b": "def", "c": -5.0}])
+
+        self._test_simple_query("""
+            select a, b, c from primary.`//tmp/t1`
+            union all
+            select a, b, c from primary.`//tmp/t2`
+            limit 4
```
rvu1024, line 994: limit приведет к появлению YtMerge. Нужно вместо limit указать две прагмы
pragma ResultRowsLimit="4";
pragma DirectRead;

Kontakter:
@rvu1024 @Tony-Romanov  – could you please communicate in English in public PRs

rvu1024 (changes requested):
I checked these changes over yql services. It seems to be working.
But any way, please add a feature flag (disabled by default) and add new behavior if this flag is on

Tony-Romanov:
> But any way, please add a feature flag (disabled by default) and add new behavior if this flag is on

No, I don't need this feature as a "feature". I did it to solve the task of switching the YQL-agent results from the Skiff format to Yson. The original problem is in the incompatibility of the format that YQL produces with the YQL-library for parsing the result. 
@nepal refused to fix the library or the YT-provider, and said that now you can read dynamic tables the same way as static, and that's how it should be done for fix. And yes, it looks like that's true, except for a couple of small bugs that this PR fixes.
Turning this into a "feature of YQL" cancels the ability to switch the format. So let's stick to the original plan and upload this change.

nepal:
> > But any way, please add a feature flag (disabled by default) and add new behavior if this flag is on
> 
> No, I don't need this feature as a "feature". I did it to solve the task of switching the YQL-agent results from the Skiff format to Yson. 

All significant/non-trivial code changes should be put under feature flag which is off by default.
After internal release we will gradually enable this feature flag from 0% to 100% users. Only after that it can be enabled by default and, after some time, old code (and feature flag itself) can be removed.

This is YQL policy for developing new features. The main purpose is not to block new releases  if there is some problem with new functionality. If such problem is discovered, we simply turn feature flag off and continue deploying release or enabling other feature flags.
So, please add feature flag. And if you want, you can always enable it by default in _your_ gateways.conf.

nepal:
> > So, please add feature flag. And if you want, you can always enable it by default in _your_ gateways.conf.
> 
> So, after discuss in tg...
> 
> Switching the format from Skiff to Yson when reading dyntables via YQL-agent will lead to errors on the same queries that work correctly now and on default settings.
> 
> Do you confirm this behavior?

You mean that there are some working queries that will fail after this PR?

nepal:
deprecated by #1180

PR #1151, opened 2025-03-25
YQL-19747 Complete a simple type name

vityaman:


vityaman:
@vitsn, need your early feedback to continue in the right direction.

yql/essentials/sql/v1/complete/name/name_service.h
```diff
@@ -0,0 +1,47 @@
...
+#include <util/generic/vector.h>
+#include <util/generic/string.h>
+
+namespace NSQLComplete {
+
+    using NThreading::TFuture;
+
+    struct TIndentifier {
+        TString Indentifier;
+    };
```
vityaman, line 14: Decided to add a type hierarchy for a an easier `TGenericName` to `TCandidate` conversion.

yql/essentials/sql/v1/complete/name/name_service.h
```diff
@@ -0,0 +1,47 @@
...
+
+    struct TTypeName: TIndentifier {
+        using TConstraints = std::monostate;
+    };
+
+    using TGenericName = std::variant<
+        TTypeName>;
```
vityaman, line 21: I decided that I want to support function names here also because it should be as easy as simple types on the same level of quality.

vityaman: Also thought about caching implementation. The CachedNameService decorator will accumulate names received in local containers of the NameCache. 

Eviction is done when some COST of memory used will be larger than limit. The COST can be defined as a sum of bytes stored in a local containers.

Also caching decorator and cache should be separate objects.

NameCache should expose methods for certain fields invalidation.

Need to choose an eviction policy.

vityaman:
@vitstn, I would like to get "Built-in YQL functions" for the `TStaticNameService`. I found the `FunctionRegistry` interface and `FillStaticModules` procedure. Is this what I need?

vitstn:
Nope. Please add some hardcoded functions right now.

I will refactor code to get real functions soon

yql/essentials/sql/v1/complete/name/name_service.h
vityaman: @vitstn, I quickly [made a prototype](https://github.com/vityaman/ytsaurus/commit/7c0ff1d538639bf8718d12e009f575354e0d7713#diff-f1ae62673005285bd54c676f2ca42c0f137f26238c4175f1eae6858648c883d8R16-R31) of how function names suggestions would be implemented. The code is not so clear, but it shows `INameService` is changing on a new name kind support added.

I realized a better module design for a function name detection, so it needs to be refactored, so I want to do it in a separate MR, therefore leave this as an only types support.

vityaman:
@vitstn, removed enum parsing for built-in types. Enumerated them manually in strings.

vitstn (approved):

Kontakter:
Shipped!

vityaman:
@vitstn, conflicts are resolved.

vitstn (approved):

Kontakter:
Shipped!

PR #1157, opened 2025-03-26
Batchify operation processing in operations_snapshots check

dmi-feo:
* Changelog entry
Type: fix
Component: odin

- operations are processed in batches in order to reduce memory usage
- operation aliases are skipped since id of their operations are checked anyway


yt/odin/checks/bin/operations_snapshots/__main__.py
```diff
@@ -1,33 +1,49 @@
...
 
-    for operation in operations:
-        args = {"attributes": ["events", "progress"]}
-        if operation.startswith("*"):
-            args["operation_alias"] = operation
-            args["include_runtime"] = True
-        else:
-            args["operation_id"] = operation
+# TODO: replace with itertools.batched when python is updated to 3.12.
+def batched(iterable: Iterable[Any], batch_size: int):
```
Kontakter, line 17: Is it same as https://github.com/ytsaurus/ytsaurus/blob/main/yt/python/yt/wrapper/common.py#L221 ?

Kontakter (approved):

yt/odin/checks/bin/operations_snapshots/__main__.py
dmi-feo: seems so. replaced `batched` with `chunk_iter_list`

Kontakter:
Shipped!

PR #1166, opened 2025-03-30
YQL-19616 Add unified lexer tests and fix lexer/regex

vityaman:
- [x] Add unified lexer tests at `sql/v1/lexer/lexer_ut.cpp`.
- [x] Fix `lexer/regex` to be case-insensitive.
- [x] Fix `lexer/regex` to prefer `DIGITS` over `INTEGER_VALUE` by preserving lexer rules order at `TLexerGrammar`.


vityaman:

PR #1168, opened 2025-03-30
YQL-19747 Add REPL mode for the tools/yql_complete

vityaman:
- [x] Added a REPL mode for the `tools/yql_complete`. This should make experimenting with the completer more convenient, than building the YDB CLI.

After an addition of the `sql/v1/highlight` we could add syntax highlighting here too.


vityaman:

yql/essentials/tools/yql_complete/yql_complete.cpp
```diff
@@ -1,20 +1,62 @@
 #include <yql/essentials/sql/v1/complete/sql_complete.h>
+#include <yql/essentials/sql/v1/complete/text/word.h>
+
+#include <contrib/restricted/patched/replxx/include/replxx.hxx>
```
vitstn, line 4: another dependency on restricted... 
do we have another repl engine for C++?

vityaman: There are not many cool alternatives

- [Isocline, C, 251 stars, last commit 3 years ago](https://github.com/daanx/isocline) has a lot of advanced features such as better completions UX, but is not so popular and seems abandoned. 

- [Linenoise, C, 3.9K stars, last commit last year](https://github.com/antirez/linenoise), not sure that it has highlighting support and seems limited in UX support.

- [GNU Readline, C](https://tiswww.case.edu/php/chet/readline/rltop.html) seems too old and lacks highlighting.

vitstn: we decided to use '#' symbol as caret position
please don't use repl mode at the moment

PR #1178, opened 2025-04-02
Add sanity check for docker registry config

koct9i:
Setting use_yt_token_for_internal_registry = %true should be used
together with setting internal_registry_address = "{registry-fqdn}".
Otherwise docker pull requests with yt token as docker_auth for
internal docker images could be redirected to external registry
whatever is set default in CRI runtime service.

---

* Changelog entry
Type: fix
Component: misc-server


sergey-v-galtsev:
LGTM

PR #1180, opened 2025-04-04
Use native read of dynamic tables.

Tony-Romanov:
Type: improvement
Component: yql


nepal (approved):

nepal:
Can you please add tests (like in previous versions of this PR)

Tony-Romanov:
> Can you please add tests (like in previous versions of this PR)

Tests from the previous versions of this PR are already in the main branch.

And all tests for query-tracker will use NativeDynamicTableRead after switch Skiff->Yson in my next PR.

Tony-Romanov:
> Can you please add tests (like in previous versions of this PR)

Tests from the previous versions of this PR are already in the main branch.

And all tests for query-tracker will use NativeDynamicTableRead after switch Skiff->Yson in my next PR.

yt/yql/providers/yt/gateway/native/yql_yt_native.cpp
```diff
@@ -3376,20 +3376,22 @@ class TYtNativeGateway : public IYtGateway {
...
-                    if (!isDynamic) {
-                        if (!limiter.NextTable(recordsCount)) {
-                            continue;
+                    const bool isDynamic = execCtx->InputTables_[i].Dynamic;
+                    if (!isDynamic || useNativeDyntableRead) {
+                        if (const auto recordsCount = execCtx->InputTables_[i].Records; recordsCount || !isDynamic) {
+                            if (!limiter.NextTable(recordsCount)) {
+                                continue;
+                            }
                         }
```
rvu1024, line 3389: This will not work properly with dynamic tables. `limiter.NextTable()` shifts `Current` by exactly the specified number of records, but for dynamic tables it is approximate. You can actually read more or fewer records. For the next table, `Current` is used to calculate range. As a result, it may be incorrect.

The current code does not change the `Current` inside `NextDynamicTable()`, but updates it to the actual value inside `NYql.::SelectRows`

Tony-Romanov: We already discussed this in the previous PR without a feature flag.

Here is exactly the same but with a switch.

What are the steps to reproduce the incorrect behavior?

rvu1024 (approved):

PR #1181, opened 2025-04-04
YQL-19747 Complete after PRAGMA and multi-token names

vityaman:
- [x] Complete after PRAGMA
- [x] Complete multi-token names correctly, for example, `yt.` returns only `DisableStrict`, not `yt.DisableStrict` and `DateTime::` returns `Split`, not `DateTime::Split`.

I tried to implement it using `CompletedToken` edition, but not all completion environments support candidates with various `contextLen` (`Replxx` does not). So I decided that completions should rewrite only the current token, not sequences. For example, on `DateTime::Spl` rewrite only `Spl`. It makes sense as multi-token names have some namespace separated by a punctuation, so used types only namespace and gets names inside of it.

### DIFF
--- a/yql/essentials/sql/v1/complete/sql_complete_ut.cpp
+++ b/yql/essentials/sql/v1/complete/sql_complete_ut.cpp
@@ -269,14 +269,36 @@ Y_UNIT_TEST_SUITE(SqlCompleteTests) {
     }
 
     Y_UNIT_TEST(Pragma) {
-        TVector<TCandidate> expected = {
-            {Keyword, "ANSI"},
-            {PragmaName, "yson.CastToString"}
-        };
-
         auto engine = MakeSqlCompletionEngineUT();
-        UNIT_ASSERT_VALUES_EQUAL(Complete(engine, {"PRAGMA "}), expected);
-        // UNIT_ASSERT_VALUES_EQUAL(Complete(engine, {"PRAGMA yson."}), expected);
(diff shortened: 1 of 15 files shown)


vityaman:
@Kontakter, could you please add @vitstn as a reviewer here also 🥺.

vityaman:
@vitstn, rebased on the `main` branch.

yql/essentials/sql/v1/complete/sql_complete_ut.cpp
```diff
@@ -402,6 +431,45 @@ Y_UNIT_TEST_SUITE(SqlCompleteTests) {
...
+        }
+        {
+            TVector<TCandidate> expected = {
+                {FunctionName, "Split("},
+            };
+            auto completion = engine->Complete({"SELECT DateTime::s"});
+            UNIT_ASSERT_VALUES_EQUAL(completion.Candidates, expected);
+            UNIT_ASSERT_VALUES_EQUAL(completion.CompletedToken.Content, "s");
+        }
+>>>>>>> 7237d79f4a9 (YQL-19747 Fix namespaced function completion)
```
vitstn, line 470: ?

vityaman: Fixed

vityaman:
@vitstn, rebased on the `main` branch.

PR #1182, opened 2025-04-05
YQL-19747 Complete type name as a function argument

vityaman:
As I understand, type name should not be completed at `SELECT |`, so I added a check that we are at `invoke_expr` context.

Currently composite type keywords are suggested at `SELECT |` and also are uppercased. I will fix it separately when this merged during 
- https://github.com/users/vityaman/projects/5?pane=issue&itemId=105056723&issue=vityaman%7Cydb%7C8

---

- Related to https://github.com/users/vityaman/projects/5/views/1?pane=issue&itemId=105056423&issue=vityaman%7Cydb%7C7


vityaman:
@Kontakter, could you please add @vitstn as a reviewer here also 🥺.

vitstn (approved):

Issue #1140, opened 2025-03-19
Master chunk replicator improvements for clusters with a small number of racks


achulkov2:
The current replicator configuration is not very convenient for clusters with a small number of racks. 

The main problem is with erasure chunks: `max_erasure_replicas_per_rack` is not really a uniform value for all erasure codecs. For example, for a cluster with 7 racks, we can satisfy <= 1 replicas per rack for `reed_solomon_3_3`, <= 2 replicas per rack for `reed_solomon_6_3` and <= 3 replicas per rack for `isa_lrc_12_2_2`. In the current configuration this would mean setting `max_reasure_replicas_per_rack` equal to 3, which is bad for `reed_solomon_3_3` — theoretically it is possible to have 3 replicas in 2 racks, and we might not survive the downtime of a rack and a host for this chunk. This chunk should have been considered unsafely placed and rebalanced. You also have to keep that current write target allocation does not consider rack-balancing, it only cares about satisfying `max_*_replicas_per_rack`, so this situation is not unlikely.

We need to work out the exact details, but the general proposal is that the replicator should derive `max_erasure_replicas_per_rack` from the erasure codec, the number of available racks, and maybe some configurable multiplier.

PR #1198, opened 2025-04-10
Erasure replicas per rack should be adaptive to erasure codec parts and racks number

faucct:
https://github.com/ytsaurus/ytsaurus/issues/1140


raundn (changes requested):
Apart from the minor issues I've commented on, there's a number of larger questions I'd like to discuss.

  1. The general idea. The first concern that comes to mind when thinking about packing erasure replicas to racks is "what if there aren't enough racks?". Each erasure codec allows losing a certain maximum number of replicas before the chunk is no longer guaranteed to be repairable. It follows, then, that using a particular erasure codec while maintaining data durability requires having a certain minimum amount of racks. **Having fewer racks is not a reason to pack tighter, it's a reason to forbid using the codec.**
  1. The use case. What is the precise goal we're trying to achieve here? Should the adaptive packing react to gradual growth of a cluster in the course of weeks, months and years? Or should it react to individual nodes/racks going offline for maintenance and then coming back? Or something in the middle (like the cluster being inflated and deflated in accordance with the load)? These are sufficiently different scenarios that may merit different solutions.
  1. The reactiveness. When the number of racks changes enough so that the max-replicas-per-rack is affected for an erasure codec, this should be treated as a global event that concerns all corresponding chunks. (Strictly speaking, this is only true for max-replicas-per-rack going *down*, but still.) The current implementation, however, only affects those chunks that are already being "looked at" by the replicator by some other reason. **Somewhere, a refresh must be scheduled for all affected chunks.** (In practice, scheduling global refresh is a better idea than maintaining per-codec sets of chunks.) Bu-u-ut such a massive refresh is costly and should be scheduled as infrequently as possible. Which brings me to the next point.
  1. The stabilization algorithm. First of all, it's essentially disabled by default (the stabilization window is "infinite"). This doesn't look like a good idea because it might result in a lot of thrashing in chunk replicator. (Admittedly, there's a defensive mechanism that will probably save us: the replicator is disabled automatically when too many data nodes are offline. But relying on it here seems weird.)
    Second of all, the algorithm itself doesn't seem to stabilize much. If I understand correctly, it simply looks back a certain amount of time. It the purpose of stabilization is to ignore sharp spikes (or dips) in the number of racks, I fail to see how this helps us. What the stabilization should do is throttling the (observable) rate of change so that rapid (and momentary) changes are completely ignored while sustained growth (or, ahem, negative growth) is acted upon. Another requirement for the algorithm is hysteresis: fluctuating around (but never settling at) a certain amount of racks should nevertheless produce a stable max-replicas-per-rack. 
  And another thing. I'm not sure managing hearbeats and tracking online nodes is such a good idea here. Instead, why not take into account the total number of known-but-not-necessarily online data nodes. Yes, this would require cleaning up old, forever-more-offline nodes in a timely manner (there's a mechanism for that, btw), and yes, the stabilization would still be necessary. But at least this would eliminate the most likely reason for thrashing because brand-new nodes coming online for the first time is a much rarer event than routine node restarts.

PR #1199, opened 2025-04-10
Close access to operations_archive

Krock21:
This change sets `inherit_acl = %false` to `//sys/operations_archive`

The main goal is to see how many tests will break and estimate impact


Kontakter:
It looks like everything if fine: all tests passed successfully excluding some flaps.

PR #1200, opened 2025-04-10
Close operations archive with fix

Krock21:
This change sets `inherit_acl = %false` to `//sys/operations_archive`, **and fixes some scheduler commands that use operations archive**

The main goal is to see how many tests will break and estimate impact

Just operations archive change: https://github.com/ytsaurus/ytsaurus/pull/1199


Kontakter:
Shipped!

Kontakter:
(to run tests in internal CI)

PR #1206, opened 2025-04-14
Add password strength validation in set_user_password request

pavel-bash:
# Description

There's no mechanism of password strength validation when setting a new password for the user. It's not possible with the current architecture to validate it server-side, so for now we're introducing a Python client configuration parameter to enforce the length. Other requirement can be added later.

### DIFF
--- a/yt/python/yt/wrapper/tests/test_auth_commands.py
+++ b/yt/python/yt/wrapper/tests/test_auth_commands.py
@@ -1,0 +1,31 @@
+from .conftest import authors
+from .helpers import set_config_option
+
+from yt.wrapper.operation_commands import add_failed_operation_stderrs_to_error_message
+
+import yt.wrapper as yt
+
+import pytest
+import string
+import random
+
+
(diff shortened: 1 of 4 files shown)


gritukan (approved):

yt/python/yt/wrapper/tests/test_auth_commands.py
```diff
@@ -0,0 +1,31 @@
...
+        password = ''.join(random.choice(string.ascii_letters) for i in range(11))
+        yt.set_user_password("admin", password)
+        with set_config_option("enable_password_strength_validation", True):
+            with pytest.raises(ValueError):
+                yt.set_user_password("admin", password)
+
+        password = ''.join(random.choice(string.ascii_letters) for i in range(12))
+        yt.set_user_password("admin", password)
+
+        password = ''.join(random.choice(string.ascii_letters) for i in range(129))
```
Kontakter, line 27: Use double quotes, please

Kontakter (approved):

PR #1208, opened 2025-04-14
Hide details about unsuccessful login attempts from the caller

pavel-bash:
# Description

When the login attempt attempt is unsuccessful due to incorrect username and/or password, the system leaks sensitive information about the nature of the failure. This PR unifies the error messages, returned to the user.

### DIFF
--- a/yt/yt/library/auth_server/cypress_cookie_login.cpp
+++ b/yt/yt/library/auth_server/cypress_cookie_login.cpp
@@ -80,14 +80,12 @@ private:
     {
         auto replyAndLogError = [&] (
             const TError& error,
-            const std::optional<std::string>& user = {},
-            bool replyWithGenericError = false)
+            bool maskError,
+            const std::optional<std::string>& user = {})
         {
-            if (replyWithGenericError) {
+            if (maskError) {
                 // Hide details about unsuccessful login attempts for security reasons.
-                auto genericError = error;
-                genericError.SetMessage("Incorrect login or password");
-                ReplyError(rsp, genericError);
+                ReplyError(rsp, TError("Incorrect login or password"));
             }
             else {
                 ReplyError(rsp, error);
@@ -103,7 +101,7 @@ private:
             rsp->SetStatus(EStatusCode::BadRequest);
 
             auto error = TError("Malformed \"Authorization\" header: failed to parse authorization method");
-            replyAndLogError(error);
+            replyAndLogError(error, /*maskError*/ false);
             return;
         }
 
@@ -111,7 +109,7 @@ private:
             rsp->SetStatus(EStatusCode::BadRequest);
 
             auto error = TError("Unsupported authorization method %Qlv", authorizationMethod);
-            replyAndLogError(error);
+            replyAndLogError(error, /*maskError*/ false);
             return;
         }
 
@@ -122,7 +120,7 @@ private:
             rsp->SetStatus(EStatusCode::BadRequest);
 
             auto error = TError("Failed to parse user credentials");
-            replyAndLogError(error);
+            replyAndLogError(error, /*maskError*/ false);
             return;
         }
 
@@ -135,13 +133,13 @@ private:
                 HandleRegularRequest(rsp);
 
                 error = TError("No such user %Qlv or user has no password set", user) << error;
-                replyAndLogError(error, TString{user}, true);
+                replyAndLogError(error, /*maskError*/ true, TString{user});
                 return;
             }
 
             // Unknown error, reply 500.
             error = TError("Failed to fetch info for user %Qlv during logging", user) << error;
-            replyAndLogError(error, TString{user});
+            replyAndLogError(error, /*maskError*/ true, TString{user});
             throw;
         }
 
@@ -149,7 +147,7 @@ private:
             HandleRegularRequest(rsp);
 
             auto error = TError("Invalid password");
-            replyAndLogError(error, TString{user}, true);
+            replyAndLogError(error, /*maskError*/ true, TString{user});
             return;
         }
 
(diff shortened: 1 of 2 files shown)


yt/yt/library/auth_server/cypress_cookie_login.cpp
```diff
@@ -78,8 +78,20 @@ class TCypressCookieLoginHandler
...
-            ReplyError(rsp, error);
+        auto replyAndLogError = [&] (
+            const TError& error,
+            const std::optional<std::string>& user = {},
+            bool replyWithGenericError = false)
+        {
+            if (replyWithGenericError) {
+                // Hide details about unsuccessful login attempts for security reasons.
+                auto genericError = error;
+                genericError.SetMessage("Incorrect login or password");
```
gritukan, line 89: This can be dangerous. In YTsaurus, errors are hierarchical, i.e. error is tuple of error message, error code, some attributes and also nested errors, so errors form a tree.

Here you change the message of the root error, potentially leaving some sensitive information in the other nodes. Let's make it simple: `TError("Incorrect login or password")` will create a new error that definitely does not have anything inside.

yt/yt/library/auth_server/cypress_cookie_login.cpp
```diff
@@ -78,8 +78,20 @@ class TCypressCookieLoginHandler
         const IRequestPtr& req,
         const IResponseWriterPtr& rsp)
     {
-        auto replyAndLogError = [&] (const TError& error, const std::optional<std::string>& user = {}) {
-            ReplyError(rsp, error);
+        auto replyAndLogError = [&] (
+            const TError& error,
+            const std::optional<std::string>& user = {},
+            bool replyWithGenericError = false)
```
gritukan, line 84: I'd prefer `bool maskError` or something like this, but current option is fine, so up to you.

yt/yt/library/auth_server/cypress_cookie_login.cpp
```diff
@@ -123,7 +135,7 @@ class TCypressCookieLoginHandler
                 HandleRegularRequest(rsp);
 
                 error = TError("No such user %Qlv or user has no password set", user) << error;
-                replyAndLogError(error, TString{user});
+                replyAndLogError(error, TString{user}, true);
```
gritukan, line 138: In C++ code we usually write `/*replyWithGenericError*/ true` in order to highlight what does this `true` mean.

yt/yt/library/auth_server/cypress_cookie_login.cpp
```diff
@@ -78,8 +78,18 @@ class TCypressCookieLoginHandler
...
+            bool maskError,
+            const std::optional<std::string>& user = {})
+        {
+            if (maskError) {
+                // Hide details about unsuccessful login attempts for security reasons.
+                ReplyError(rsp, TError("Incorrect login or password"));
+            }
+            else {
+                ReplyError(rsp, error);
+            }
```
gritukan, line 92: Let's add a blank line here for readability.

gritukan (approved):

yt/yt/library/auth_server/cypress_cookie_login.cpp
```diff
@@ -78,8 +78,19 @@ class TCypressCookieLoginHandler
...
+        auto replyAndLogError = [&] (
+            const TError& error,
+            bool maskError,
+            const std::optional<std::string>& user = {})
+        {
+            if (maskError) {
+                // Hide details about unsuccessful login attempts for security reasons.
+                ReplyError(rsp, TError("Incorrect login or password"));
+            }
+            else {
```
savnadya, line 90: `else` should be on the previous line

yt/yt/library/auth_server/cypress_cookie_login.cpp
```diff
@@ -123,21 +134,21 @@ class TCypressCookieLoginHandler
                 HandleRegularRequest(rsp);
 
                 error = TError("No such user %Qlv or user has no password set", user) << error;
-                replyAndLogError(error, TString{user});
+                replyAndLogError(error, /*maskError*/ true, TString{user});
```
savnadya, line 137: Let's make it configurable?
Now passing `error` here doesn't make sense at all :)

yt/yt/library/auth_server/cypress_cookie_login.cpp
```diff
@@ -123,21 +134,21 @@ class TCypressCookieLoginHandler
...
                 return;
             }
 
             // Unknown error, reply 500.
             error = TError("Failed to fetch info for user %Qlv during logging", user) << error;
```
savnadya, line 142: I think `Failed to fetch info for user %Qlv during logging` shouldn't be masked.

savnadya (changes requested):

yt/yt/library/auth_server/cypress_cookie_login.cpp
savnadya: Or maybe log original error and doesn't return it?

pavel-bash: Well, this error goes to `YT_LOG_DEBUG` on line 94 in the lambda, so I think it still makes sense to pass it to the lambda itself, isn't it?

yt/yt/library/auth_server/cypress_cookie_login.cpp
pavel-bash: Agree, I'll fix it.

pavel-bash:
@savnadya I've fixed all comments except for the one I had question about.

Kontakter:
Shipped!

Kontakter:
(to run tests in internal CI)

Kontakter:
There are small styling issues
```
yt/yt/tests/integration/proxies/test_cypress_cookie_auth.py:51: [E251] unexpected spaces around keyword / parameter equals
yt/yt/tests/integration/proxies/test_cypress_cookie_auth.py:51: [E251] unexpected spaces around keyword / parameter equals
```

pavel-bash:
@Kontakter fixed the styling issue.

Kontakter:
Shipped!

PR #1209, opened 2025-04-15
YQL-19747 Introduce Schema and complete a table name

vityaman:


yql/essentials/sql/v1/complete/syntax/parser_call_stack.cpp
```diff
@@ -101,6 +106,18 @@ namespace NSQLComplete {
                Contains({RULE(External_call_param), RULE(An_id)}, stack);
     }
 
+    bool IsLikelyTableStack(const TParserCallStack& stack) {
+        return !Contains({RULE(Create_table_stmt),
```
vitstn, line 110: create table 'foo/#bar/baz' should suggest bar folder

vityaman: For the first version I propose not to implement this feature, because this requires a token content analysis and should be a separate feature, because it is a completion not on a parser level, but on a lexer level, so should be implemented separately.

If leave it as is, we would be able just to load table names such as `.sys/partition_stats`, `.sys/top_queries_by_duration_one_minute`, `.sys/query_metrics_one_minute` to a `NameSet` and have a such behaviour.

```kotlin
:) SELECT * FROM `.sys/<TAB>`
[.sys/]partition_stats
[.sys/]top_queries_by_duration_one_minute
[.sys/]query_metrics_one_minute
```

which seems good enough for a start.

vityaman:
Notes: https://gist.github.com/vityaman/104188d0cfe117fd6cc3167af31e9ef4

vityaman:
Need to design a Gateway to get database objects.

```c++
using Path = List[String] // [aaa, bbb, ccc] means /aaa/bbb/ccc

enum ObjectType {
  Folder,
  Table,
  ...
}

enum ObjectProperty {
  Columns,
  ...,
}

struct Column {
  name: String
}

struct Table {
  columns: List<Column>
}

...

using Object = Variant[Table, ...]

struct Filter {
  namePrefix: String
  types: Set[ObjectType]
}

struct FolderEntry {
  type: ObjectType
  name: String
}

class Schema {
  List(cluster: String, folder: Path, filter: Filter, limit: UInt): Set[FolderEntry]
  Describe(cluster: String, folder: Path, entryName: String, properties: Set[ObjectProperty]): Object
}
```

vityaman:
@vitstn, what do you think about this gateway C++ interface design?

vitstn:
>using Object = Variant[Folder, Table, ...]
we need to represent an unknown Object (only name & type are available)

>using Path = List[String] // [aaa, bbb, ccc] means /aaa/bbb/ccc
isn't a good - each system may use different path separator

>List(cluster: String, folder: Path
we should use prefix here instead of folder

> Describe(cluster: String, folder: Path
objectPath instead of folder

vityaman:
@vitstn, 

> we need to represent an unknown Object (only name & type are available)

ok, then 

```
struct Unexpected {
  name: String
}

using Object = Variant[Folder, Table, ..., Unexpected]
```

> isn't a good - each system may use different path separator

ok, then

```
using Path = String
```

> we should use prefix here instead of folder

ok, then 

```c++
List(cluster: String, path: Path, filter: Filter, limit: UInt): List[FolderEntry]
```

but `path` does not contain a `suffix`. Want to have `path` meaning a valid path, not like path with prefix for filtration, filtration should be in the `Filter` struct.

> objectPath instead of folder

ok, then

```c++
Describe(cluster: String, path: Path, properties: Set[ObjectProperty]): Object
```

vityaman:
- name outside object variant
- need to have string repr for unknown just to show
- `Filter = (prefix: string, types: Maybe<Set<String>>)`.

vityaman:
Some notes 

```
- NameService takes Gateways for each service
- Gateway: get objects by prefix, returns object types (directory, table, ...)
- Mapping from a object type to a object type at a concrete target system
- Consider checking USE, need deafult system (ydb at YDB CLI)

NSQLComplete::Gateway
- def List(clusterName: String, prefix: Path, logicalTypes: [String], limit: Int): Future[List[Object = (LocalName: String, LogicalType: String)]]
- def Describe(clusterName: String, objectPath: Path, properties: [String]): Object

note: prefix is equal to database path on ID_PLAIN
note: USE specifies a clusterName

LogicalType = Folder | Table | Uknown
note: logical type is a string but can be uknown but shown on a frontend (in candidates list)

TODO: introduce gateway at PR "tables complete"

Names for Logical Types "Keyword", "External Table", ...

Object = Variant<Table, Topic, ...>

NSQLComplete::Table : Object
  columns: Maybe[[Columns]]

Topic: 
 ...

Describe(Table)

Alter table ... remove column <column>.

note: batching may be useful on joined tables schema, but multiple futures is ok

Object = Table columns | 

prefix examples:
1.  aaa/b         -> aaa/
2.  aaa/bbb/      -> aaa/bbb/

SELECT * FROM `aaa/b#bb/ccc`

keywords: navigate, scheme, 

TODO: ranking for tables. F

select a.col1 from 

TODO: suggest table names in backticks everytime

note: Highlighter Specification should contain algorithm, json key names, logical token names, not concrete regexes
note: client maps token kind kind to theme
note: Fix key set of highlighting JSON
note: 


```

vityaman:
TODO: filter `FolderEntry` by privileges.

PR #1212, opened 2025-04-17
Store a user ID alongside the Cypress token instead of a username

pavel-bash:
# Description

Current authentication token attribute contains a username in the form of `@user` attribute, which is used to retrieve the user when authenticating with a token. This leads to a problem when a user is renamed - the system will still try to use the old username which it retrieved from a token, resulting in authentication failures.

This PR fixes the issue by replacing the old attribute by a `@user_id` containing ID of the user which is not changed during the renaming process. Backward compatibility is also introduced, so that the users can still use, list and revoke their tokens which were created with the old attribute. However, all new tokens will use the new attribute.

The only problem left is proxy-side caching which will still try to use the old username. This will be the scope of another PR.


dim-an (approved):
Ship to run the tests

yt/yt/library/auth_server/cypress_token_authenticator.cpp
```diff
@@ -75,6 +82,59 @@ class TCypressTokenAuthenticator
...
             }
         }
+        auto tokenNode = ConvertTo<INodePtr>(rspOrError.Value());
+        const auto& tokenAttributes = tokenNode->Attributes();
+
+        auto userIdAttribute = tokenAttributes.Find<TString>("user_id");
+        if (userIdAttribute) {
+            // New authentication schema: now we need to get the username given the user ID which we received.
+            auto path = Format("#%v/@name", ToYPathLiteral(*userIdAttribute));
+            return Client_->GetNode(path, /*options*/ {})
```
dim-an, line 92: I think we should `.ReadFrom=Cache` as well

yt/yt/library/auth_server/cypress_token_authenticator.cpp
```diff
@@ -75,6 +82,59 @@ class TCypressTokenAuthenticator
...
+            return MakeFuture(TAuthenticationResult{
+                .Login = *userAttribute,
+            });
+        }
+
+        YT_LOG_DEBUG(rspOrError, "Cypress authentication failed (TokenHash: %v)",
+            tokenHash);
+        THROW_ERROR_EXCEPTION("Cypress authentication failed")
+            << TErrorAttribute("token_hash", tokenHash)
+            << rspOrError;
```
dim-an, line 115: As far as I understand rpsOrError is in ok condition here and error message here is not helpful.

yt/yt/ytlib/api/native/client_authentication_impl.cpp
```diff
@@ -142,7 +142,22 @@ TIssueTokenResult TClient::DoIssueTokenImpl(
...
+        /*options*/ {}));
+    if (!userIdRspOrError.IsOK()) {
+        YT_LOG_DEBUG(userIdRspOrError, "Failed to issue new token for user: "
+            "could not get user ID by username "
+            "(User: %v, TokenPrefix: %v, TokenHash: %v)",
+            user,
+            tokenPrefix,
+            tokenHash);
+        auto error = TError("Failed to issue new token for user") << userIdRspOrError;
+        THROW_ERROR error;
```
dim-an, line 157: Why not `THROW_ERROR_EXCEPTION` ?

yt/yt/ytlib/api/native/client_authentication_impl.cpp
```diff
@@ -142,7 +142,22 @@ TIssueTokenResult TClient::DoIssueTokenImpl(
...
+    }
+
+    attributes->Set("user_id", ConvertTo<TString>(userIdRspOrError.Value()));
```
dim-an, line 160: This file uses `std::string` now, let's avoid using TString here.

yt/yt/ytlib/api/native/client_authentication_impl.cpp
```diff
@@ -226,20 +240,48 @@ void TClient::DoRevokeToken(
...
+        // Resolve the username with the retrieved user_id.
+        auto tokenUsernameOrError = WaitFor(rootClient->GetNode(
+            Format("#%v/@name", ToYPathLiteral(*userIdAttribute)),
+            getOptions));
+        if (!tokenUsernameOrError.IsOK()) {
+            YT_LOG_DEBUG(tokenUsernameOrError, "Failed to get user for token (TokenHash: %v)",
+                tokenSha256);
+            auto error = TError("Failed to get user for token")
+                << tokenUsernameOrError;
+            THROW_ERROR error;
```
dim-an, line 269: Why not `THROW_ERROR_EXCEPTION` ?

yt/yt/ytlib/api/native/client_authentication_impl.cpp
```diff
@@ -300,14 +342,26 @@ TListUserTokensResult TClient::DoListUserTokens(
...
     }
 
+    auto userIdRspOrError = WaitFor(rootClient->GetNode(
+        Format("//sys/users/%v/@id", user),
+        /*options*/ {}));
+    if (!userIdRspOrError.IsOK()) {
+        YT_LOG_DEBUG(userIdRspOrError, "Failed to list tokens: could not get user ID by username (User: %v)",
+            user);
+        auto error = TError("Failed to list tokens") << userIdRspOrError;
+        THROW_ERROR error;
```
dim-an, line 352: why not `THROW_ERROR_EXCEPTION` ?

yt/yt/library/auth_server/cypress_token_authenticator.cpp
```diff
@@ -75,6 +82,59 @@ class TCypressTokenAuthenticator
...
+    }
+
+    TAuthenticationResult OnCallUsernameResult(
+        const TString& tokenHash,
+        const TString& userId,
+        const TErrorOr<TYsonString>& rspOrError)
+    {
+        if (!rspOrError.IsOK()) {
+            if (rspOrError.FindMatching(NYTree::EErrorCode::ResolveError)) {
```
dim-an, line 124: Does it really make sense to split error handling here? 

I would maybe add error wrapper "Could not get username by user ID" for the case of resolve error. But would keep YT_LOG_DEBUG messages the same. Otherwise they are harder to grep.

dim-an (commented):
Can you explain, what's your plan to fix caching in next PR (or may be it was discussed in SIG chat) ?

pavel-bash:
@dim-an fixed most of the comments. Some remarks:

> why not THROW_ERROR_EXCEPTION ?

Most of the exceptions in that file are thrown with `THROW_ERROR`, so I've done the same for consistency. If you want, I can still change it.

> .ReadFrom=Cache

Not possible due to the rules of this initializer - `ReadFrom` comes from one of the parent, so it's possible to do it right in this initializer. I've done it with a line after.

pavel-bash:
@Kontakter Fixed.

Kontakter:
Shipped!

yt/yt/library/auth_server/cypress_token_authenticator.cpp
```diff
@@ -78,6 +83,58 @@ class TCypressTokenAuthenticator
...
             }
         }
+        auto tokenNode = ConvertTo<INodePtr>(rspOrError.Value());
+        const auto& tokenAttributes = tokenNode->Attributes();
+
+        auto userIdAttribute = tokenAttributes.Find<TString>("user_id");
+        if (userIdAttribute) {
+            // New authentication schema: now we need to get the username given the user ID which we received.
+            auto path = Format("#%v/@name", ToYPathLiteral(*userIdAttribute));
+            TGetNodeOptions options{};
```
dim-an, line 93: We don't use such initializers

yt/yt/library/auth_server/cypress_token_authenticator.cpp
```diff
@@ -78,6 +83,58 @@ class TCypressTokenAuthenticator
...
+                    userId);
+            if (rspOrError.FindMatching(NYTree::EErrorCode::ResolveError)) {
+                THROW_ERROR_EXCEPTION("Could not get username by user ID")
+                    << TErrorAttribute("token_hash", userId)
+                    << rspOrError;
+            } else {
+                THROW_ERROR_EXCEPTION("Cypress authentication failed")
+                    << TErrorAttribute("user_id", userId)
+                    << rspOrError;
+            }
```
dim-an, line 136: This separation is not useful since you don't introduce separate error codes for the cases (and user can get explanation of error in inner error).

yt/yt/library/auth_server/cypress_token_authenticator.cpp
```diff
@@ -78,6 +83,58 @@ class TCypressTokenAuthenticator
...
+                *userAttribute);
+            return MakeFuture(TAuthenticationResult{
+                .Login = *userAttribute,
+            });
+        }
+
+        YT_LOG_DEBUG(rspOrError, "Cypress authentication failed (TokenHash: %v)",
+            tokenHash);
+        THROW_ERROR_EXCEPTION("Cypress authentication failed")
+            << TErrorAttribute("token_hash", tokenHash);
```
dim-an, line 117: Logging error and adding attribute to it is repetetive and could be unified.

dim-an (changes requested):
I still have some comments, but I think it would be easier to fix it myself than to send you for another round of code review. So I will merge your PR.

Issue #1213, opened 2025-04-17
Access control for operations and jobs


Krock21:
Currently every operation is visible to every user. Access control for them is implemented partially for some APIs like `get-job-stderr`

We need to implement full access control for operations and jobs:
1. `//sys/operations` and `//sys/operations_archive` nodes should be closed with ACLs. Currently they inherit `allow users read` from `//sys`.
2. Operation and Job APIs should validate access to operation
  - `get-operation`, `get-job`, `list-operations`, `list-jobs`, ...

PR #1214, opened 2025-04-17
Add access control tests for get-job, list-operations, list-jobs

Krock21:
Issue: #1213

This PR adds tests that demonstrate current situation, where `no_rights_user` has access to all operations

Further PRs will change these tests if default behaviour changes, or use them to verify that new functionality under a feature flag does not break default behaviour


Kontakter:
Shipped!

PR #1216, opened 2025-04-17
Close access to //sys/operations, use operations archive client in get-job to access it

Krock21:
Issue: https://github.com/ytsaurus/ytsaurus/issues/1213. Part 3 where we close access to `//sys/operations`

This PR closes access to `//sys/operations` by default and adjusts APIs who relied on it

PR #1215, opened 2025-04-17
Close access to operations_archive by default. Use operations archive client for operations archive requests

Krock21:
Issue: https://github.com/ytsaurus/ytsaurus/issues/1213. Part 2 where we close access to `//sys/operations_archive`

This PR closes access to `//sys/operations_archive` by default and adjusts APIs who relied on it

This PR consists of 2 commits: Tests and actual code. Tests have [a separate PR](https://github.com/ytsaurus/ytsaurus/pull/1214)

PR #1221, opened 2025-04-19
Add strict access control validation for scheduler commands

Krock21:
Issue: https://github.com/ytsaurus/ytsaurus/issues/1213. Part 4 where we introduce a feature flag to enforce strict access control validation for scheduler commands

This PR adds a feature flag `strict_scheduler_commands_access_validation` which fixes access control behaviour of scheduler commands (get-operation, get-job, list-operation, ...).

This PR consists of 3 commits: closure of `//sys/operations_archive`, closure of `//sys/operations` and actual feature.
Every commit has a separate PR. [Archive](https://github.com/ytsaurus/ytsaurus/pull/1215). [Cypress](https://github.com/ytsaurus/ytsaurus/pull/1216)

---

* Changelog entry
Type: feature
Component: proxy

Add a feature flag to enforce access control for scheduler commands (get-operation, get-job, ...)

### DIFF
--- a/yt/yt/client/api/operation_client.h
+++ b/yt/yt/client/api/operation_client.h
@@ -158,6 +158,10 @@ struct TListOperationsOptions
 
     TDuration ArchiveFetchingTimeout = TDuration::Seconds(3);
 
+    // These parameters cannot be set from list_operations command, they must be computed.
+    THashSet<std::string> UserTransitiveClosure = {};
+    bool StrictSchedulerCommandsAccessValidation = false;
+
     TListOperationsOptions()
     {
         ReadFrom = EMasterChannelKind::Cache;
--- a/yt/yt/tests/integration/scheduler/test_scheduler_acls.py
+++ b/yt/yt/tests/integration/scheduler/test_scheduler_acls.py
@@ -30,7 +30,7 @@ import pytest
 import random
 import string
 import time
-from datetime import datetime
+from datetime import datetime, timedelta
 from contextlib import contextmanager
 from copy import deepcopy
 
@@ -56,9 +56,8 @@ def _update_op_parameters(**kwargs):
     update_op_parameters(kwargs.pop("operation_id"), **kwargs)
 
 
-class TestSchedulerAcls(YTEnvSetup):
+class TestSchedulerAclsBase(YTEnvSetup):
     ENABLE_MULTIDAEMON = False  # There are component restarts.
-    NUM_TEST_PARTITIONS = 6
     USE_DYNAMIC_TABLES = True
 
     NUM_MASTERS = 1
@@ -118,7 +117,7 @@ class TestSchedulerAcls(YTEnvSetup):
     }
 
     def setup_method(self, method):
-        super(TestSchedulerAcls, self).setup_method(method)
+        super(TestSchedulerAclsBase, self).setup_method(method)
 
         # Init operations archive.
         sync_create_cells(1)
@@ -166,8 +165,9 @@ class TestSchedulerAcls(YTEnvSetup):
     def _validate_access(user, should_have_access, action, **action_args):
         has_access = True
         authorization_error = None
+        result = None
         try:
-            action(authenticated_user=user, **action_args)
+            result = action(authenticated_user=user, **action_args)
         except YtError as e:
             if not e.contains_code(yt_error_codes.AuthorizationErrorCode):
                 raise
@@ -185,6 +185,7 @@ class TestSchedulerAcls(YTEnvSetup):
             )
             if not has_access:
                 message += ". Got error response {}".format(authorization_error)
+            message += ". Action result: {}".format(result)
             raise AssertionError(message)
 
     def _run_and_fail_op(self, should_update_operation_parameters):
@@ -248,6 +249,10 @@ class TestSchedulerAcls(YTEnvSetup):
             except YtError:
                 # TODO: Ensure it is "no such operation" error or operation has failed or aborted.
                 pass
+
+
+class TestSchedulerAcls(TestSchedulerAclsBase):
+    NUM_TEST_PARTITIONS = 17
 
     @authors("omgronny")
     @pytest.mark.parametrize("should_update_operation_parameters", [False, True])
@@ -337,6 +342,8 @@ class TestSchedulerAcls(YTEnvSetup):
         op, job_id = self._run_and_fail_op(should_update_operation_parameters=False)
         clean_operations()
         for action in actions:
+            # Wait for job to become available.
+            retry(lambda: action(operation_id=op.id, job_id=job_id))
             self._validate_access(self.no_rights_user, False, action, operation_id=op.id, job_id=job_id)
             self._validate_access(self.read_only_user, True, action, operation_id=op.id, job_id=job_id)
             self._validate_access(self.manage_only_user, False, action, operation_id=op.id, job_id=job_id)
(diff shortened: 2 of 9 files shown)


Krock21:
Rebased

Krock21:
Rebased

yt/yt/ytlib/api/native/client_operation_info_impl.cpp
```diff
@@ -1269,12 +1324,28 @@ TListOperationsResult TClient::DoListOperations(const TListOperationsOptions& ol
         options.SubstrFilter = to_lower(*options.SubstrFilter);
     }
 
+    auto client = GetOperationsArchiveClient();
+
+    if (Connection_->GetConfig()->StrictSchedulerCommandsAccessValidation) {
+        client = this;
```
Kontakter, line 1330: What does this `if` and assignment mean?

yt/yt/client/api/operation_client.h
```diff
@@ -158,6 +158,10 @@ struct TListOperationsOptions
 
     TDuration ArchiveFetchingTimeout = TDuration::Seconds(3);
 
+    // These parameters cannot be set from list_operations command, they must be computed.
```
Kontakter, line 161: I don't this that it is a good idea. This API can be used directly.

Kontakter: It is better to pass it directly to corresponding methods and classes, or maybe introduce `struct TListOperationContext` and pass it

Kontakter:
I have the suggestion; maybe the new logic could be expressed in the following way:
1. Expand `AccessFilter` to support multiple expressions with `AND` semantics
2. If `StrictSchedulerCommandsAccessValidation` specified then just add an additional access filter with current user and read permission

yt/yt/ytlib/api/native/client_operation_info_impl.cpp
```diff
@@ -733,6 +757,30 @@ TOperation TClient::DoGetOperation(
...
+                        this,
+                        Logger());
+                } else {
+                    THROW_ERROR_EXCEPTION(
+                        "Failed to validate operation access, operation has no access control rule")
+                        << TErrorAttribute("operation_id", operationId);
+                }
+            }
+
+            if (Connection_->GetConfig()->StrictSchedulerCommandsAccessValidation &&
```
eshcherbin, line 777: We already have an `if (Connection_->GetConfig()->StrictSchedulerCommandsAccessValidation)`, why duplicate it?

yt/yt/ytlib/api/native/client_operation_info_impl.cpp
```diff
@@ -699,6 +702,22 @@ TOperation TClient::DoGetOperationImpl(
...
+            if (auto acl = runtimeParametersNode->AsMap()->FindChild("acl")) {
+                return ConvertTo<TSerializableAccessControlList>(acl);
+            }
+            if (auto acoName = runtimeParametersNode->AsMap()->FindChild("aco_name")) {
+                return ConvertTo<TString>(acoName);
+            }
+        }
+    }
+    return {};
+}
```
eshcherbin, line 719: Let's reuse https://github.com/ytsaurus/ytsaurus/blob/85241756df65575c729debd1602b49491539abd7/yt/yt/ytlib/api/native/client_job_info_impl.cpp#L645 instead of this helper?

yt/yt/ytlib/api/native/config.h
```diff
@@ -360,6 +360,9 @@ struct TConnectionDynamicConfig
 
     THashSet<std::string> BannedInSyncReplicaClusters;
 
+    //! Enables strict access validation in scheduler commands.
+    bool StrictSchedulerCommandsAccessValidation;
```
eshcherbin, line 364: "Scheduler commands" is too broad. I would suggest `StrictOperationInfoAccessValidation` or something like that.

yt/yt/ytlib/api/native/list_operations.cpp
```diff
@@ -593,7 +593,8 @@ class TFilteringConsumer
         cursor->ParseMap([&] (TYsonPullParserCursor* cursor) {
             YT_VERIFY((*cursor)->GetType() == EYsonItemType::StringValue);
             auto key = (*cursor)->UncheckedAsString();
-            if (Options_.AccessFilter && key == TStringBuf("acl")) {
+            if (key == TStringBuf("acl") &&
+                (Options_.StrictSchedulerCommandsAccessValidation || Options_.AccessFilter)) {
```
eshcherbin, line 597: `{` on the new line.

eshcherbin (changes requested):

Krock21:
> 1. Expand `AccessFilter` to support multiple expressions with `AND` semantics

I thought about it, but I think users may expect `OR` semantics and I think it is better to either support something complex with `AND, OR, NOT`, or not support it at all
It will also be hard to change in the future when we will decide to support all cases

I can support `AND` semantics internally but hide it from users

yt/yt/ytlib/api/native/client_operation_info_impl.cpp
```diff
@@ -699,6 +702,22 @@ TOperation TClient::DoGetOperationImpl(
     return *archiveResult;
 }
 
+static std::optional<TAccessControlRule> GetOperationAccessControlRule(const TOperation& operation)
+{
+    if (operation.RuntimeParameters) {
+        auto runtimeParametersNode = ConvertToNode(operation.RuntimeParameters);
+        if (runtimeParametersNode->GetType() == ENodeType::Map) {
```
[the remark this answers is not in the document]
Krock21: I am not sure, but I think I had some issues with it during testing

yt/yt/ytlib/api/native/client_operation_info_impl.cpp
Krock21: It is a client used to create `ObjectServiceReadProxy` to fetch `SubjectClosure (//sys/users/[user]/@member_of_closure)`

By default `OperationsArchiveClient` is used, so I keep it. With `StrictSchedulerCommandsAccessValidation`, `this` is used and it checks that user has `read` access to `//sys/users/[user_for_check]`

yt/yt/ytlib/api/native/config.h
Krock21: I chose it because of [scheduler_commands.h](https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/client/driver/scheduler_commands.h)/.cpp files and some mentions of "scheduler" commands

`get_job` doesn't seem to belong to `OperationInfo` commands, but yt config [already has](https://github.com/ytsaurus/ytsaurus/blob/5f3a4f4dedb5583b856f3ced3b2a97f568422601/yt/python/yt/wrapper/default_config.py#L666) `operation_info_commands_timeout` which includes get_job so I will change it to `StrictOperationInfoAccessValidation`

yt/yt/ytlib/api/native/client_operation_info_impl.cpp
Krock21: It is easier to understand and avoid mistakes in the future
```
DoGetOperation {
  if (StrictMode) options.Attributes.emplace("runtime_parameters"))
  
  CommonLogic();
  
  if (StrictMode) {
    DoActualLogic();
  }

  if(StrictMode && !originalOptions.Attributes.count("runtime_parameters")) result.erase("runtime_parameters");
  
  return result;
}
```
Seems better than
```
DoGetOperation {
  if (StrictMode) options.Attributes.emplace("runtime_parameters"))
  
  CommonLogic();
  
  if (StrictMode) {
    DoActualLogic();
    if (!originalOptions.Attributes.count("runtime_parameters")) {
      // This should be the last logic in the function before returning result
      result.erase("runtime_parameters");
    }
  }
  
  // Please don't write anything that requires runtime_parameters here
  
  return result;
}
```

yt/yt/ytlib/api/native/config.h
eshcherbin: Scheduler commands includes all kind of things, e.g. starting an operation, aborting a job, etc.

Regarding `get_job` and `list_jobs`, one can argue that a job is a part of an operation, so they also could fall under the "operation info" label.

yt/yt/ytlib/api/native/client_operation_info_impl.cpp
eshcherbin: We can customize this method to either fail or validate when it sees an empty ACL.

I am mostly against duplicating the ACR extraction logic.

yt/yt/ytlib/api/native/client_operation_info_impl.cpp
eshcherbin: To my taste, if you want to ensure that `runtime_parameters` are removed at the very end, it's better to wrap it with a `Finally`.

I am not strongly against the current version, but it looks odd.

yt/yt/ytlib/api/native/client_operation_info_impl.cpp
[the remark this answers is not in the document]
eshcherbin: I agree, this condition should be obsolete.

yt/yt/ytlib/api/native/client_operation_info_impl.cpp
eshcherbin: I would suggest using a ternary operator and adding an NB-comment to clarify this.

eshcherbin: I would suggest using a ternary operator and adding an NB-comment to clarify this.

yt/yt/client/api/operation_client.h
Krock21: We actually have similar field inside AccessFilter: https://github.com/ytsaurus/ytsaurus/blob/e1b174fe02ef703344fea34504cd42a5534c6a9c/yt/yt/client/api/operation_client.h#L121-L122

And it follows the same path

I have rewritten it to use `TListOperationContext`. Unfortunately, it had to be placed in the same file with `TListOperationOptions`

yt/yt/ytlib/api/native/client_operation_info_impl.cpp
Krock21: I actually used a similar thing to logic about `state` field for GetOperation: https://github.com/ytsaurus/ytsaurus/blob/e1b174fe02ef703344fea34504cd42a5534c6a9c/yt/yt/ytlib/api/native/client_operation_info_impl.cpp#L695-L698

I am not sure I can use `Finally` here
1. It will modify returning value. Seems dangerous
2. It will be executed when exception is caught, and we don't want it. This code is:
    ```
    while (true) {
      try {
        auto operation = get_operation()
        maybe_throw(operation)
        maybe_adjust_operation(operation)
        return operation
      } catch (error) {
        retry
      }
    }
    ```

I failed to find a working place for Finally or any better solution. Please say if you see one

I can put it as an inner "if" if you wish, but I think it will be harder to understand and easier to make a mistake later (put your code after it)
Did it in next iteration

yt/yt/ytlib/api/native/client_operation_info_impl.cpp
Krock21: Hmm. Seems like validating against an empty ACL is fine since it will allow admins only and is equivalent to failing for regular users. Will use that

yt/yt/ytlib/api/native/client_operation_info_impl.cpp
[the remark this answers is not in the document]
Krock21: Removed this function

yt/yt/ytlib/api/native/list_operations.cpp
```diff
@@ -738,6 +739,21 @@ class TFilteringConsumer
             return false;
         }
 
+        // Verify that the user has "read" access to the operation.
+        if (Options_.StrictSchedulerCommandsAccessValidation &&
+            !Options_.UserTransitiveClosure.contains(SuperusersGroupName)) {
```
Krock21: `{` on the new line here too, I suppose
Assuming this is required for `if` statements with multiline expressions

Krock21:
Fixed comments

Krock21:
Rebased

yt/yt/ytlib/api/native/client_operation_info_impl.cpp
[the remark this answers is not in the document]
Krock21: >I agree, this condition should be obsolete.

Do you mean it may be required in older versions? 24.1?

yt/yt/ytlib/api/native/client_impl.h
```diff
@@ -1315,6 +1315,7 @@ public: \
     void DoListOperationsFromCypress(
         TListOperationsCountingFilter& countingFilter,
         const TListOperationsOptions& options,
+        const TListOperationsContext& context,
```
eshcherbin: Usually, context is passed by raw pointer, because it is often mutated. In this case, we don't need to mutate it yet, but it may be a good idea to pass-by-pointer for later as per convention.

yt/yt/ytlib/api/native/client_operation_info_impl.cpp
```diff
@@ -1269,12 +1298,26 @@ TListOperationsResult TClient::DoListOperations(const TListOperationsOptions& ol
         options.SubstrFilter = to_lower(*options.SubstrFilter);
     }
 
+    // NB(aleksandr.gaev): This client is used to get subject closure (@member_of_closure) of a user, which is protected by an ACL.
+    auto client = Connection_->GetConfig()->StrictOperationInfoAccessValidation ? this : GetOperationsArchiveClient();
```
eshcherbin: This line is rather long, I'd split it into
```
auto val = cond
    ? a
    : b;
```

yt/yt/tests/integration/scheduler/test_scheduler_acls.py
```diff
@@ -261,6 +262,10 @@ def _run_op_context_manager(self, should_update_operation_parameters=False, spec
                 # TODO: Ensure it is "no such operation" error or operation has failed or aborted.
                 pass
 
+
+class TestSchedulerAcls(TestSchedulerAclsBase):
+    NUM_TEST_PARTITIONS = 17
```
eshcherbin: Why 17 after 6 previously? Seems to be too big of a gap.

yt/yt/tests/integration/scheduler/test_scheduler_acls.py
```diff
@@ -856,3 +863,242 @@ def validate_operations(list_result, expected):
...
+        validate_operations(list_operations(authenticated_user=self.manage_and_read_user, access=access_filter), [op.id])
+        set(f"//sys/users/{self.read_only_user}/@acl", [])
+        validate_operations(list_operations(authenticated_user=self.no_rights_user, access=access_filter), [op.id])
+        validate_operations(list_operations(authenticated_user=self.manage_and_read_user, access=access_filter), [op.id])
+
+
+class TestSchedulerAclsStrictMode(TestSchedulerAclsBase):
+    NUM_TEST_PARTITIONS = 6
+
+    # NB(aleksandr.gaev): I have no idea why DELTA_DRIVER_CONFIG is needed and not cluster_connection config.
```
eshcherbin: I suggest determining the reason behind this and not leaving a commented out `DELTA_PROXY_CONFIG` below.

eshcherbin:
Shipped!

yt/yt/tests/integration/scheduler/test_scheduler_acls.py
Krock21: I tried, but failed and left this comment
Can you help me?

yt/yt/tests/integration/scheduler/test_scheduler_acls.py
Krock21: Even 17 didn't help, I still see 480s+ running time. Further splitting of a test is required

eshcherbin: This should not be normal. Which build type are you using? Try relwithdebinfo.

yt/yt/tests/integration/scheduler/test_scheduler_acls.py
eshcherbin: I suspect HTTP proxies are just not used in integration tests.

yt/yt/ytlib/api/native/client_impl.h
Krock21: Raw pointers are scary. I wasn't able to find them in `yt/yt/ytlib`, but I found some `*Ptr` usages
Rewrote for intrusive ptr

yt/yt/tests/integration/scheduler/test_scheduler_acls.py
Krock21: I was just running `ya make -A yt/yt/tests/integration/scheduler -F '*test_scheduler_acls.py*'`. Seems like it uses debug

Adding `--build=relwithdebinfo` helps, thanks

Krock21: Returned 6

yt/yt/tests/integration/scheduler/test_scheduler_acls.py
Krock21: Seems true, but adding `ENABLE_HTTP_PROXY = True` doesn't help
And http proxy tests do some magic with http requests to make http proxies work

Krock21: Did
```
    # NB(aleksandr.gaev): This is a config for tests only, since tests don't use http proxies.
    # For http proxy config, `cluster_connection.strict_operation_info_access_validation` should be adjusted.
    DELTA_DRIVER_CONFIG = {
        "strict_operation_info_access_validation": True,
    }
```

Krock21:
Fixed comments

yt/yt/client/api/operation_client.h
```diff
@@ -164,6 +164,16 @@ struct TListOperationsOptions
     }
 };
 
+struct TListOperationsContext
+    : public TRefCounted
```
eshcherbin: nit: You could just make it `final`.

yt/yt/ytlib/api/native/client_operation_info_impl.cpp
```diff
@@ -1269,12 +1298,28 @@ TListOperationsResult TClient::DoListOperations(const TListOperationsOptions& ol
...
+        ? this
+        : GetOperationsArchiveClient();
+
+    auto proxy = NObjectClient::CreateObjectServiceReadProxy(
+        client,
+        options.ReadFrom,
+        PrimaryMasterCellTagSentinel,
+        Connection_->GetStickyGroupSizeCache());
+
+    TListOperationsContextPtr context = New<TListOperationsContext>();
```
eshcherbin: nit: `auto`

Krock21:
Fixed comments

PR #1225, opened 2025-04-21
YQL-19747 Detect a token at the caret position

vityaman:
When I tried to implement a folder and object names completion at `ID_QUOTED` I faced with a problem, that I actually can't detect, that cursor is at `ID_QUOTED` token because `TCompletionInput::Text` it was cut until the `TCompletionInput::CursorPosition`, therefore at input ``` SELECT * FROM `#` ``` prefix was ``` SELECT * FROM `# ``` and then lexer failed.

While we actually want tokenize the whole current statement, `C3` still needs to receive a prefix as input. I tried to tokenize the whole statement and then on input `SELECT Optional<#>` got nothing because `<>` is solid token in the `SQL`. The only way to fix it I found is to cut a query to prefix until the cursor position.

BTW, current implementation is not so efficient as we tokenize the input multiple times. Especially `SplitQueryToStatemnts` seems heavy. In future we anyway will parse the whole input so will need to design APIs to receive ready token streams to do statements splitting, for example, just not to do the work twice.

image

So I introduce you the following changes

- [x] Select the whole current statement, not just prefix.
- [x] Find the token at caret and output no candidates when caret is at `STRING_VALUE`, `DIGIGTS` and so on.
- [x] Change `C3` wrapper interface to take `TCompletionInput` to hide an implementation detail that it runs on cut prefix.
- [x] `#` annotated queries in unit tests.
- [x] Detect `CaretTokenPosition` -- if is it enclosed with a token or between two.
- [x] Ensure that `maxErrors` in `ILexer::Tokenize` is positive. Just a tiny bugfix.

---

- Related to https://github.com/ytsaurus/ytsaurus/pull/1209


yql/essentials/sql/v1/complete/syntax/local.cpp
```diff
@@ -55,14 +55,29 @@ namespace NSQLComplete {
...
                 return {};
             }
 
-            auto candidates = C3.Complete(prefix);
+            auto candidates = C3.Complete(statement);
 
-            NSQLTranslation::TParsedTokenList tokens = Tokenized(prefix);
+            NSQLTranslation::TParsedTokenList tokens = Tokenized(statement.Text);
+            if (!tokens.empty()) {
+                tokens.crop(CaretTokenIndex(tokens, statement.CursorPosition) + 1);
```
vityaman, line 67: Hide it under GetTokenPrefix with tokens.back() at caret position.

vityaman:
@vitstn, what do you think about these changes?

yql/essentials/sql/v1/complete/sql_complete_ut.cpp
```diff
@@ -510,20 +552,43 @@ Y_UNIT_TEST_SUITE(SqlCompleteTests) {
...
+        auto engine = MakeSqlCompletionEngineUT();
+        UNIT_ASSERT_VALUES_EQUAL(Complete(engine, "SELECT \"#\""), empty);
+        UNIT_ASSERT_VALUES_EQUAL(Complete(engine, "SELECT `#`"), empty);
+        UNIT_ASSERT_VALUES_EQUAL(Complete(engine, "SELECT 21#21"), empty);
+
+        UNIT_ASSERT_GT(Complete(engine, "SELECT `name`").size(), 0);
+        UNIT_ASSERT_GT(Complete(engine, "SELECT #`name`").size(), 0);
+
+        UNIT_ASSERT_GT(Complete(engine, "SELECT \"a\"#\"b\"").size(), 0);
+        UNIT_ASSERT_VALUES_EQUAL_C(Complete(engine, "SELECT `a`#`b`"), empty, "Solid ID_QUOTED");
```
vityaman, line 570: Add test for `a#\#`#b`.

yql/essentials/sql/v1/complete/sql_complete_ut.cpp
```diff
@@ -510,20 +552,43 @@ Y_UNIT_TEST_SUITE(SqlCompleteTests) {
...
+    Y_UNIT_TEST(Enclosed) {
+        TVector<TCandidate> empty = {};
+
+        auto engine = MakeSqlCompletionEngineUT();
+        UNIT_ASSERT_VALUES_EQUAL(Complete(engine, "SELECT \"#\""), empty);
+        UNIT_ASSERT_VALUES_EQUAL(Complete(engine, "SELECT `#`"), empty);
+        UNIT_ASSERT_VALUES_EQUAL(Complete(engine, "SELECT 21#21"), empty);
+
+        UNIT_ASSERT_GT(Complete(engine, "SELECT `name`").size(), 0);
+        UNIT_ASSERT_GT(Complete(engine, "SELECT #`name`").size(), 0);
```
vityaman, line 567: Check that 1st -- from, 2nd -- function.

vitstn (approved):

PR #1227, opened 2025-04-22
YQL-19747 Introduce SchemaGateway interface

vityaman:
I decided to be lazy and specify the only method `List` for a `SchemaGateway`, as `Describe` will be needed only when we would like to add column completions and I think this will not happen on the nearest ~2 weeks, because just listing a schema let us to complete a lot of objects: folders, tables, indexes, roles, users... While implementing these objects completions I will think about schema synthesis at background. I think that more time to think of the API design is better.

I decided not to add some constructions like `std::variant` and other things to make it easier to use and more elegant, because actually we want to keep it simple to implement. So the interface is as dump and simple as possible. I named it the `SchemaGateway` because there maybe will be a `Schema` that will be a wrapper around `SchemaGateway` with that fancy and expressive C++ constructions.

Errors are not specified because I did not come up with an idea how to recover from them. I think that all errors should be just propagated and somewhere written to some error listener and on error replaced with a `pure EmptyResponse`.

I added a `StaticSchemaGateway` to show an expected behaviour from an implementation and also to use it in `sql/v1/complete` unit tests.

---

- Related to https://github.com/ytsaurus/ytsaurus/pull/1209

### DIFF
--- a/yql/essentials/sql/v1/complete/name/object/schema_gateway.h
+++ b/yql/essentials/sql/v1/complete/name/object/schema_gateway.h
@@ -21,7 +21,7 @@ namespace NSQLComplete {
     };
 
     struct TListRequest {
-        TString System;
+        TString Cluster;
 
         // `Path` structure is defined by a `System`.
         // Can end with a folder entry name hint.
--- a/yql/essentials/sql/v1/complete/name/object/schema_gateway.h
+++ b/yql/essentials/sql/v1/complete/name/object/schema_gateway.h
@@ -34,7 +34,7 @@ namespace NSQLComplete {
     };
 
     struct TListResponse {
-        size_t NameHintLength;
+        size_t NameHintLength = 0;
         TVector<TFolderEntry> Entries;
     };
 
--- a/yql/essentials/sql/v1/complete/name/object/schema_gateway.h
+++ b/yql/essentials/sql/v1/complete/name/object/schema_gateway.h
@@ -30,7 +30,7 @@ namespace NSQLComplete {
         TString Path;
 
         TListFilter Filter;
-        TMaybe<size_t> Limit;
+        size_t Limit = 128;
     };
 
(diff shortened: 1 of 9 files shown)


vityaman:
@vitstn, let's discuss this proposal.

yql/essentials/sql/v1/complete/name/object/schema_gateway.h
```diff
@@ -0,0 +1,49 @@
...
+
+        friend bool operator==(const TFolderEntry& lhs, const TFolderEntry& rhs) = default;
+    };
+
+    struct TListFilter {
+        TMaybe<THashSet<TString>> Types;
+    };
+
+    struct TListRequest {
+        TString System;
```
vitstn, line 24: we need a cluster, not only system here

yql/essentials/sql/v1/complete/name/object/schema_gateway.h
```diff
@@ -0,0 +1,49 @@
...
+        // For example, `/local/exa` lists a folder `/local`,
+        // but can rank and filter entries by a hint `exa`.
+        TString Path;
+
+        TListFilter Filter;
+        size_t Limit = 128;
+    };
+
+    struct TListResponse {
+        size_t NameHintLength;
```
vitstn, line 37: let's zero primitive fields to avoid UB

yql/essentials/sql/v1/complete/name/object/schema_gateway.h
vityaman: I think that a authentication should not be a first class field, because we can put its properties just to `Context`.

vityaman:
```
struct TListRequest {
  TString Cluster;

  ...
};
```

vitstn (approved):

PR #1229, opened 2025-04-22
YQL-19747 Normalize names for ranking and filtering

vityaman:
I was lazy to search for a most frequent used name among equivalent by the relation `(a ~ b) iff (NormalizeName(a) = NormalizeName(b))`. Because it seems that names we receive from JSONs are canonized and therefore in a preferable style by the opinion of the YQL language designers. But because of duplicates at `statements_opensource.json` we have, for example, both `IGNORETYPEV3` and `IGNORE_TYPE_V3` in candidates list. I think that we should just remove `IGNORETYPEV3` from the JSON.

---

- Related to https://github.com/vityaman/ydb/issues/21

### DIFF
--- a/yql/essentials/sql/v1/complete/name/static/frequency.cpp
+++ b/yql/essentials/sql/v1/complete/name/static/frequency.cpp
@@ -1,6 +1,6 @@
 #include "frequency.h"
 
-#include <yql/essentials/core/sql_types/normalize_name.h>
+#include "name_index.h"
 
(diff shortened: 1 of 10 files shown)


vityaman:

vityaman:
- [ ] Unique names and choice the most frequent.

vityaman:
@vitstn, now most frequent names are chosen in each equivalence class.

vitstn (approved):

PR #1234, opened 2025-04-23
transaction commits and aborts should validate ACLs

faucct:
It was not being validated for commits and also the authentication identity was being changed to root.

### DIFF
--- a/yt/yt/server/master/transaction_server/transaction_manager.cpp
+++ b/yt/yt/server/master/transaction_server/transaction_manager.cpp
@@ -2575,7 +2575,6 @@ private:
 
         const auto& securityManager = Bootstrap_->GetSecurityManager();
         TAuthenticatedUserGuard userGuard(securityManager);
-        securityManager->ValidatePermission(transaction, EPermission::Write);
 
         auto prerequisiteTransactionIds = FromProto<std::vector<TTransactionId>>(request->prerequisite_transaction_ids());
 
--- a/yt/yt/server/master/transaction_server/transaction_manager.cpp
+++ b/yt/yt/server/master/transaction_server/transaction_manager.cpp
@@ -727,6 +727,9 @@ public:
 
         auto transactionId = transaction->GetId();
 
+        const auto& securityManager = Bootstrap_->GetSecurityManager();
+        securityManager->ValidatePermission(transaction, EPermission::Write);
+
         auto state = transaction->GetPersistentState();
         if (state == ETransactionState::Committed) {
             YT_LOG_DEBUG("Transaction is already committed (TransactionId: %v)",
@@ -879,7 +882,6 @@ public:
             options.CommitTimestampClusterTag,
             time);
 
-        const auto& securityManager = Bootstrap_->GetSecurityManager();
         securityManager->ChargeUser(user, {EUserWorkloadType::Write, 1, time});
     }
 
@@ -1921,6 +1923,7 @@ public:
                 transactionId,
                 prerequisiteTransactionIds,
                 std::move(mutationId),
+                NRpc::GetCurrentAuthenticationIdentity(),
                 isRetry)
                     .AsyncVia(EpochAutomatonInvoker_));
     }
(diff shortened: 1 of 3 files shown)


yt/yt/tests/integration/master/test_master_transactions.py
```diff
@@ -480,6 +480,23 @@ def test_owner(self):
...
+        create_user("u")
+        set("//sys/schemas/transaction/@acl", [*get("//sys/schemas/transaction/@acl"), {
+            "action": "deny",
+            "subjects": ["u"],
+            "permissions": ["write"],
+            "inheritance_mode": "object_and_descendants",
+        }])
+        tx = start_transaction()
+        create("map_node", "//tmp/wut", tx=tx)
+        with pytest.raises(YtError):
```
zlobober, line 494: I'd suggest using a context manager raises_yt_error and specify particular AuthorizationError error code, similar to how it is done here
https://github.com/ytsaurus/ytsaurus/blob/180ae01baadd781a680721ef645ed4724ab79a86/yt/yt/tests/integration/queues/test_producer.py#L226

yt/yt/server/master/transaction_server/transaction_manager.cpp
```diff
@@ -2520,6 +2524,10 @@ class TTransactionManager
         auto transactionId = FromProto<TTransactionId>(request->transaction_id());
         auto* transaction = GetTransactionOrThrow(transactionId);
 
+        const auto& securityManager = Bootstrap_->GetSecurityManager();
+        TAuthenticatedUserGuard userGuard(securityManager);
+        securityManager->ValidatePermission(transaction, EPermission::Write);
```
raundn, line 2529: I believe this only handles one of two places where the validation should be done. The second one is `HydraPrepareAndCommitCommitCypressTransactionInSequoia`. Or better yet, the validation should be moved to `PrepareAndCommitCypressTransaction` as that's called from both places.

yt/yt/tests/integration/master/test_master_transactions.py
```diff
@@ -480,6 +480,23 @@ def test_owner(self):
...
         assert get("#{0}/@owner".format(tx)) == "u"
 
+    @authors("faucct")
+    def test_not_owner(self):
+        create_user("u")
+        set("//sys/schemas/transaction/@acl", [*get("//sys/schemas/transaction/@acl"), {
+            "action": "deny",
+            "subjects": ["u"],
+            "permissions": ["write"],
+            "inheritance_mode": "object_and_descendants",
```
raundn, line 490: - There's a handy `make_ace` helper function in `yt_commands`.
- There's a better way to append to yson list: `set("//sys/schemas/transaction/@acl/end", make_ace("deny", "u", ["write"])`  (Here, `end` is a keyword.)

yt/yt/server/master/transaction_server/transaction_manager.cpp
faucct: `PrepareAndCommitCypressTransaction` already had a `ValidatePermission`, but it wasn't working though.

faucct: I think that `TAuthenticatedUserGuard userGuard(securityManager)` is doing all the work, by copying the identity from global variable to the `SecurityManager`. I can move it closer to the existing `ValidatePermission`, if you want.

faucct: I have already answered that it is not required and `TAuthenticatedUserGuard` is doing all the work.

maxim-babenko:
Please capitalize the commit message

maxim-babenko:
Master reign promotion seems necessary

yt/yt/server/master/transaction_server/transaction_manager.cpp
raundn: Then let's remove this call and I think that's it for this pull request.

PR #1239, opened 2025-04-24
Use CRON to set the schedule of queues export

pavel-bash:
# Description

Currently the interval of queues export can only be set as "Every N units". This sometimes does not work well as it always starts the count from the time `0` (January 1st, 1970). It means that, for instance, if the customer wants to set an export to "Every week", it will happen every Thursday.

This PR fixes the problem by providing a CRON way to set the schedule which is proven to be very flexible and meeting (almost) any requirement the user may have.


yt/yt/server/queue_agent/queue_exporter_old.cpp
```diff
@@ -256,7 +256,8 @@ class TQueueExportTaskOld
 
     ui64 GetMinFragmentUnixTs(TTimestamp timestamp)
     {
-        auto period = ExportConfig_->ExportPeriod.Seconds();
+        // TODO: fix it correctly, or leave like that just to be compiled?
```
pavel-bash, line 259: @achulkov2 please pay attention to this one. I'm not sure if we can leave the code like that - if for some reason this old implementation is used with CRON, it will lead to a crash.

yt/yt/client/queue_client/config.cpp
```diff
@@ -67,15 +71,30 @@ void TQueueStaticExportConfig::Register(TRegistrar registrar)
...
-            THROW_ERROR_EXCEPTION("The value of \"export_period\" must be a multiple of 1000 (1 second)");
+        if (config->ExportPeriod) {
+            if (config->ExportPeriod->GetValue() % TDuration::Seconds(1).GetValue() != 0) {
+                THROW_ERROR_EXCEPTION("The value of \"export_period\" must be a multiple of 1000 (1 second)");
+            }
+            config->ExportSchedule = *config->ExportPeriod;
+        } else if (config->ExportCronExpression) {
+            try {
+                config->ExportSchedule.emplace<TCronExpression>(*config->ExportCronExpression);
+            } catch (const yexception& exc) {
```
achulkov2, line 82: We usually just catch std::exception, our own YT exceptions have a separate path, and yexception is yandex-util-specfic, so best to be generic.

yt/yt/client/queue_client/config.cpp
```diff
@@ -67,15 +71,30 @@ void TQueueStaticExportConfig::Register(TRegistrar registrar)
...
+                THROW_ERROR_EXCEPTION(
+                    "The value of \"export_cron\" is not a well-formed CRON expression (Value: '%v', Error: '%v')",
```
achulkov2, line 84: Unlike logs, in exceptions we format everything within the exception text (sometimes with Qv for automatic quotation marks around strings), and extraneous data via TErrorAttribute (you can codesearch usages). 

So here it would be something like `Export cron expression %Qv is not well-formed`. 
And the error itself can be added as a child-error, via the `<<` operator (same as attributes). Should work for yexception just as well as for ours.

So, something like
```THROW_ERROR_EXCEPTION("Export cron expression %Qv is not well-formed", *config->ExportCronExpression)
    << exc;```
    
Also, we tend to use `ex` instead of `exc`.

File: yt/yt/server/queue_agent/queue_exporter.cpp [hunk 4]
```diff
@@ -78,25 +78,54 @@ namespace {
...
 //! Returns [#beginTs; endTs), where endTs is also export unix ts describing the range.
 //! Alternatively, beginTs is export unix ts that comes before endTs.
-std::pair<ui64, ui64> GetExportUnixTsRange(ui64 unixTs, TDuration exportPeriod)
+std::pair<ui64, ui64> GetExportUnixTsRange(
+    TInstant now,
+    std::variant<TDuration, TCronExpression>& exportSetting)
 {
-    auto period = exportPeriod.Seconds();
-    YT_VERIFY(period > 0);
+    if (exportSetting.index() == 0) {
```
achulkov2, line 89: This will be much clearer with a struct + checking for optional values.

yt/yt/server/queue_agent/queue_exporter.cpp
```diff
@@ -315,7 +344,7 @@ class TQueueExportTask
             transactionId);
 
         TaskInstant_ = TInstant::Now();
-        ExportUnixTsUpperBound_ = GetExportUnixTsUpperBound(TaskInstant_, ExportConfig_->ExportPeriod);
+        ExportUnixTsUpperBound_ = GetExportUnixTsUpperBound(TaskInstant_, ExportConfig_->ExportSchedule);
```
achulkov2, line 347: If done via config inheritance, this will just be an upcast. Or you can keep the fields in the config and make a helper struct which is retrieved via ExportConfig_->GetExportSchedule(). 

You could also store this helper-struct right in the config, but I don't see a good reason to, constructing a struct is free.

yt/yt/server/queue_agent/queue_exporter.cpp
```diff
@@ -1240,7 +1283,17 @@ class TQueueExporter
...
         // NB(apachee): Table lag is rounded up, since export period could've changed after the last successful export task.
-        auto exportPeriodSeconds = exportConfig->ExportPeriod.Seconds();
+        ui64 exportPeriodSeconds;
+        switch (exportSchedule.index()) {
+            case 0: {
+                exportPeriodSeconds = std::get<0>(exportSchedule).Seconds();
+                break;
+            }
+            case 1: {
+                exportPeriodSeconds = GetCronExpressionPeriod(std::get<1>(exportSchedule));
```
achulkov2, line 1293: Here as well, I would try to understand the value being exported here and make it correct. Even though this is a bit less important than the naming.

yt/yt/server/queue_agent/queue_exporter_old.cpp
achulkov2: I think it is best to fix it correctly, shouldn't be too different from the new version.

yt/yt/ytlib/ya.make
```diff
@@ -742,6 +742,7 @@ PEERDIR(
     library/cpp/erasure
     library/cpp/iterator
     library/cpp/yt/backtrace/symbolizers/dwarf
+    library/cpp/cron_expression
```
achulkov2, line 745: And yeah, dependency is probably misplaced. Depending on what you do, it will either go in server or directly into client.

yt/yt/client/queue_client/config.h
```diff
@@ -64,8 +66,17 @@ bool operator==(const TQueueAutoTrimConfig& lhs, const TQueueAutoTrimConfig& rhs
...
+    //! with ExportSchedule parameter.
+    std::optional<TDuration> ExportPeriod;
+
+    //! Export will be performed at schedule that is defined by this CRON expression. Mutually exclusive
+    //! with ExportPeriod parameter.
+    std::optional<TString> ExportCronExpression;
+
+    //! Either the export period or the CRON expression; it is a part of the configuration to avoid creating
+    //! structures many times during the execution of the program.
+    std::variant<TDuration, TCronExpression> ExportSchedule;
```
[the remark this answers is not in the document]
achulkov2: Looking at the fact, that TCronExpression is a new dependency, I am not sure it is really worth placing it into the config itself. Constructing TCronExpression is pretty cheap, it is pretty much just an allocation. And if we ever want to go into economy mode here we can just construct and store it inside QueueExporter-related classes.

So I vote for keeping just TString here and instantiating TCronExpression when needed. This way we only add PEERDIR into server, not client.

achulkov2 (changes requested):

pavel-bash:
@achulkov2 fixed most of your comments. I still need to fix the `queue_exporter_old`, but everything else you can give another look.

ItIsApachee:
@pavel-bash, I believe that code inside `TQueueExportTask` should not have different logic depending on how export schedule is defined (either via export period, or via cron schedule). Code will be much easier to read and maintain if a shared interface for both types of schedule is used. I think the right way to implement that would be to:
1. Re-write everything as if we only have cron schedules.
2. Adapt code for export period using the one for cron schedules, since export period is basically the same thing, but much simpler.

@achulkov2, what do you think about the suggestion?

yt/yt/tests/integration/queues/test_queue_agent.py
```diff
@@ -3995,6 +3995,37 @@ def assert_exported_table_count(count):
...
+        self._flush_table("//tmp/q")
+        wait(lambda: len(ls(export_dir)) == 1, timeout=7)
+
+        insert_rows("//tmp/q", [{"data": "nano"}])
+        self._flush_table("//tmp/q")
+        wait(lambda: len(ls(export_dir)) == 2, timeout=7)
+
+        self.remove_export_destination(export_dir)
+
     # COMPAT(apachee): Ensure old implementation is actually used.
```
achulkov2, line 4029: Let's add test for some non-trivial cron behaviour that cannot be expressed as a single period? Esp. when combined with naming. For example, you can make exports run at some weird non-evenly spaced seconds within every minute.

yt/yt/tests/integration/queues/test_queue_agent.py
```diff
@@ -3995,6 +3995,37 @@ def assert_exported_table_count(count):
...
+    @pytest.mark.timeout(150)
+    def test_cron_annotation_schedule(self):
+        # Check that the new CRON annotation works correctly, at least for this use-case.
+
+        _, queue_id = self._create_queue("//tmp/q")
+
+        export_dir = "//tmp/export"
+        self._create_export_destination(export_dir, queue_id)
+
+        export_cron_expression = "0/5 * * * * *"  # Every 5 seconds.
```
achulkov2, line 4008: I just realized, cron expressions with seconds are not explicitly standard. Linux cron only has minutes, I think. We should mention the fact that seconds are supported in the comments. And probably give a link to the README/header of the library we are using where the expressions are described.

achulkov2 (changes requested):

ItIsApachee:
Do we actually need to add changes to `queue_exporter_old`? The whole point of it being added was to be able to quickly fallback to the old implementation if issues occur. 
What if instead we add cron-like schedules only to the new implementation, and have the old one fallback to export period (meaning no changes to `queue_exporter_old`).  This way if issues with new schedules happen, they can be resolved quickly. This would require always using cron expression if available, and only in case of it missing would the export period be used instead.

pavel-bash:
@ItIsApachee so, essentially, you want to remove the possibility to specify periods for the new implementation, and if they're anyway specified somehow, it will fallback to the old implementation?

achulkov2:
> @pavel-bash, I believe that code inside `TQueueExportTask` should not have different logic depending on how export schedule is defined (either via export period, or via cron schedule). Code will be much easier to read and maintain if a shared interface for both types of schedule is used. I think the right way to implement that would be to:
> 
> 1. Re-write everything as if we only have cron schedules.
> 
> 2. Adapt code for export period using the one for cron schedules, since export period is basically the same thing, but much simpler.
> 
> 
> 
> @achulkov2, what do you think about the suggestion?

It is a good point, but I don't think you can express any period as a cron expression :( 
For period > 60 seconds and not a divisor of minutes/hours/etc I think it cannot be represented. 

If it can be represented, then yes, it would be much better to convert period to cron expression.

Or we can consider this a breaking change and only allow certain periods (which are convertible to CRON expressions). I will think about it, @savnadya can think as well. 
I don't see anything very restricting here.

achulkov2:
> Do we actually need to add changes to `queue_exporter_old`? The whole point of it being added was to be able to quickly fallback to the old implementation if issues occur. 
> 
> What if instead we add cron-like schedules only to the new implementation, and have the old one fallback to export period (meaning no changes to `queue_exporter_old`).  This way if issues with new schedules happen, they can be resolved quickly. This would require always using cron expression if available, and only in case of it missing would the export period be used instead.

Well, it depends on how soon you want to deprecate the old version. What is the prognosis for new version to be considered stable enough? :)

yt/yt/server/queue_agent/queue_exporter.cpp
```diff
@@ -684,7 +717,12 @@ class TQueueExportTask
 
     TString GetOutputTableName(ui64 unixTs)
     {
-        auto periodInSeconds = ExportConfig_->ExportPeriod.Seconds();
+        ui64 periodInSeconds;
+        if (ExportConfig_->ExportPeriod) {
```
achulkov2, line 721: Let's wrap this into a function, like `GetLastExportPeriod` or smth.

yt/yt/server/queue_agent/queue_exporter.cpp
```diff
@@ -78,25 +79,55 @@ namespace {
...
+        NDatetime::Convert(cronExpression.CronNext(nowCivilSecond), UTCTimezone).Seconds()};
+}
 
-    std::pair<ui64, ui64> result;
-    result.first = (unixTs / period) * period;
-    result.second = result.first + period;
-    return result;
+std::pair<ui64, ui64> GetExportUnixTsRange(
+    ui64 unixTs,
+    const TQueueStaticExportConfig& exportConfig)
```
achulkov2, line 110: We tend to pass around pointers to configs because there is no good need for the extra dereferencing.

yt/yt/server/queue_agent/queue_exporter.cpp
```diff
@@ -78,25 +79,55 @@ namespace {
...
+    TInstant now,
+    const TQueueStaticExportConfig& exportConfig)
 {
-    return GetExportUnixTsRange(now.Seconds(), exportPeriod).first;
+    return GetExportUnixTsRange(now, exportConfig).first;
+}
+
+//! Get the difference between #now and the previous CRON
+//! occurence, according to the provided #cronExpressionString.
+ui64 GetCronExpressionPeriod(ui64 now, const TString& cronExpressionString) {
```
achulkov2, line 125: Let's either rename this function (like SecondsSinceSmth) to show that it just subtracts now - last occurrence, or rename `now` -> something signifying that it is the unix ts of a current cron expression occurrence.

yt/yt/server/queue_agent/queue_exporter_old.cpp
achulkov2: Ok, let's require the old setting, but throw an exception instead of failing. Its text will show up in an alert about this queue.

achulkov2 (changes requested):

pavel-bash:
@ItIsApachee @achulkov2 to summarize: the approach will be left as-is, in the form of functions over the `exportConfig` in `queue_exporter.cpp` file + I'll fix the comments from Andrei.

achulkov2:
About config structure: I thought about separating export schedule into a separate config-field, but (even though it is not too hard, just move values into it in the config post-processor), I don't think it is worth it (we either have to migrate clients, or just be stuck with this compatibility code).
A better bet would be aiming to deprecate `export_period` eventually and leave cron schedule as the only option.

All in all, too much time wasted thinking about something not particularly important, let's just do something reasonable.

yt/yt/tests/integration/queues/test_queue_agent.py
pavel-bash: @achulkov2 I would love to add such a test, but I'm not sure it's possible to really check that the export happens in those intervals. There's just too many parts to sync.

Let's say we set seconds `4,20,50` as the seconds of CRON expression. If we really want to check the intervals, we'll need to wait for second 0, push something, wait 4 seconds, check, push more, wait 16 seconds... While it's possible to write it, I am not sure it will always pass due to timing issues. What do you think?

pavel-bash:
@achulkov2 @ItIsApachee I have addressed the PR comments, the only thing left is the test requested by Andrei, about which I'm not sure yet. Also, some of the older tests are now running and successfully passing in CRON mode.

Kontakter:
Shipped!

Kontakter:
(to run tests in internal CI)

Kontakter:
There are compilation errors:
```
$(SOURCE_ROOT)/library/cpp/cron_expression/cron_expression.cpp:1082:21: error: absolute value function 'abs' given an argument of type 'year_t' (aka 'long') but has parameter of type 'int' which may cause truncation of value [-Werror,-Wabsolute-value]
 1082 |                 if (abs(calendar.year() - baseYear) > CRON_MAX_YEARS_DIFF) {
      |                     ^
$(SOURCE_ROOT)/library/cpp/cron_expression/cron_expression.cpp:1082:21: note: use function 'std::abs' instead
 1082 |                 if (abs(calendar.year() - baseYear) > CRON_MAX_YEARS_DIFF) {
      |                     ^~~
      |                     std::abs
$(SOURCE_ROOT)/library/cpp/cron_expression/cron_expression.cpp:1082:53: error: comparison of integers of different signs: 'int' and 'const uint32_t' (aka 'const unsigned int') [-Werror,-Wsign-compare]
 1082 |                 if (abs(calendar.year() - baseYear) > CRON_MAX_YEARS_DIFF) {
      |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~
2 errors generated.
```

achulkov2:
> There are compilation errors:
> 
> ```
> $(SOURCE_ROOT)/library/cpp/cron_expression/cron_expression.cpp:1082:21: error: absolute value function 'abs' given an argument of type 'year_t' (aka 'long') but has parameter of type 'int' which may cause truncation of value [-Werror,-Wabsolute-value]
>  1082 |                 if (abs(calendar.year() - baseYear) > CRON_MAX_YEARS_DIFF) {
>       |                     ^
> $(SOURCE_ROOT)/library/cpp/cron_expression/cron_expression.cpp:1082:21: note: use function 'std::abs' instead
>  1082 |                 if (abs(calendar.year() - baseYear) > CRON_MAX_YEARS_DIFF) {
>       |                     ^~~
>       |                     std::abs
> $(SOURCE_ROOT)/library/cpp/cron_expression/cron_expression.cpp:1082:53: error: comparison of integers of different signs: 'int' and 'const uint32_t' (aka 'const unsigned int') [-Werror,-Wsign-compare]
>  1082 |                 if (abs(calendar.year() - baseYear) > CRON_MAX_YEARS_DIFF) {
>       |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~
> 2 errors generated.
> ```

They look like Werror problems within the library itself :) 
Something induced from including from within YT perhaps, like -Wsign-compare?

Should we fix them in this PR/separate PR/you want to fix it? @Kontakter

Kontakter:
> > There are compilation errors:
> > ```
> > $(SOURCE_ROOT)/library/cpp/cron_expression/cron_expression.cpp:1082:21: error: absolute value function 'abs' given an argument of type 'year_t' (aka 'long') but has parameter of type 'int' which may cause truncation of value [-Werror,-Wabsolute-value]
> >  1082 |                 if (abs(calendar.year() - baseYear) > CRON_MAX_YEARS_DIFF) {
> >       |                     ^
> > $(SOURCE_ROOT)/library/cpp/cron_expression/cron_expression.cpp:1082:21: note: use function 'std::abs' instead
> >  1082 |                 if (abs(calendar.year() - baseYear) > CRON_MAX_YEARS_DIFF) {
> >       |                     ^~~
> >       |                     std::abs
> > $(SOURCE_ROOT)/library/cpp/cron_expression/cron_expression.cpp:1082:53: error: comparison of integers of different signs: 'int' and 'const uint32_t' (aka 'const unsigned int') [-Werror,-Wsign-compare]
> >  1082 |                 if (abs(calendar.year() - baseYear) > CRON_MAX_YEARS_DIFF) {
> >       |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~
> > 2 errors generated.
> > ```
> 
> They look like Werror problems within the library itself :) Something induced from including from within YT perhaps, like -Wsign-compare?
> 
> Should we fix them in this PR/separate PR/you want to fix it? @Kontakter

We need to fix it, because the build is not passed :)

Let do it in this PR

pavel-bash:
@Kontakter I have pushed a fix for the build, but I couldn't reproduce the failure locally, so I hope it'll work.

Kontakter:
Shipped!

yt/yt/tests/integration/queues/test_queue_agent.py
```diff
@@ -3612,20 +3624,35 @@ def test_wrong_originating_queue(self):
...
         self._create_export_destination(export_dir, queue_id)
-        set("//tmp/q/@static_export_config", {
-            "default": {
-                "export_directory": export_dir,
-                "export_period": export_period_seconds * 1000,
-                "output_table_name_pattern": "%ISO-period-is-%PERIOD-fmt-%Y.%m.%d.%H.%M.%S",
-                "use_upper_bound_for_table_names": use_upper_bound_for_table_names,
-            }
-        })
+
```
achulkov2, line 3637: Let's wrap this into a helper method somehow, specifying just the period/expression (or auto-converting period into expression), so that we don't have to duplicate the remainder of the parameters.

achulkov2 (changes requested):
Last few comments.

pavel-bash:
@achulkov2 fixed your last comments as well, also added the test.

Kontakter:
Now, there is some problem on win platform :(

```
$(SOURCE_ROOT)/library/cpp/cron_expression/cron_expression.cpp(1082,43): error: absolute value function 'labs' given an argument of type 'year_t' (aka 'long long') but has parameter of type 'long' which may cause truncation of value [-Werror,-Wabsolute-value]
 1082 |                 if (static_cast<uint32_t>(std::labs(calendar.year() - baseYear)) > CRON_MAX_YEARS_DIFF) {
      |                                           ^
$(SOURCE_ROOT)/library/cpp/cron_expression/cron_expression.cpp(1082,43): note: use function 'std::abs' instead
 1082 |                 if (static_cast<uint32_t>(std::labs(calendar.year() - baseYear)) > CRON_MAX_YEARS_DIFF) {
      |                                           ^~~~~~~~~
      |                                           std::abs
1 error generated.
```

yt/yt/tests/integration/queues/test_queue_agent.py
```diff
@@ -3995,6 +4013,111 @@ def assert_exported_table_count(count):
...
+        if getattr(self, "USE_OLD_QUEUE_EXPORTER_IMPL"):
+            pytest.skip()
+
+        _, queue_id = self._create_queue("//tmp/q")
+
+        export_dir = "//tmp/export"
+        self._create_export_destination(export_dir, queue_id)
+
+        export_seconds = [0, 17, 29, 45]
+        export_cron_expression = "0,17,29,45 * * * * *"  # Every ~15 seconds.
```
achulkov2, line 4059: `",".join(export_seconds)` in order not to repeat the values?

yt/yt/tests/integration/queues/test_queue_agent.py
```diff
@@ -3995,6 +4013,111 @@ def assert_exported_table_count(count):
...
+        time1_second = datetime.datetime.now().second
+
+        insert_rows("//tmp/q", [{"data": "nano"}])
+        self._flush_table("//tmp/q")
+        wait(lambda: len(ls(export_dir)) == 2, timeout=20)
+        time2_second = datetime.datetime.now().second
+
+        # The seconds which we acquired must be (almost) equal to the ones
+        # specified in the CRON expression and they must be consecutive.
+        time1_second_found = None
```
achulkov2, line 4082: Let's base this not on physical time, but on the names of the tables. You can see in other tests that one can specify for the export timestamp to be part of the name. Then you can use these timestamps to check whether they are exactly equal to two consecutive values in the schedule.

yt/yt/tests/integration/queues/test_queue_agent.py
```diff
@@ -3995,6 +4013,111 @@ def assert_exported_table_count(count):
...
+                "export_directory": export_dir,
+                "export_cron_schedule": export_cron_expression,
+            }
+        })
+
+        self._wait_for_component_passes()
+
+        insert_rows("//tmp/q", [{"data": "vim"}])
+        self._flush_table("//tmp/q")
+        wait(lambda: len(ls(export_dir)) == 1, timeout=20)
```
achulkov2, line 4072: We can just use the default timeout.

yt/yt/tests/integration/queues/test_queue_agent.py
```diff
@@ -3612,18 +3617,24 @@ def test_wrong_originating_queue(self):
...
         _, queue_id = self._create_queue("//tmp/q")
         self._create_export_destination(export_dir, queue_id)
+
         set("//tmp/q/@static_export_config", {
             "default": {
                 "export_directory": export_dir,
-                "export_period": export_period_seconds * 1000,
                 "output_table_name_pattern": "%ISO-period-is-%PERIOD-fmt-%Y.%m.%d.%H.%M.%S",
                 "use_upper_bound_for_table_names": use_upper_bound_for_table_names,
+                "export_cron_schedule": "0/3 * * * * *" if use_cron else None,
```
achulkov2, line 3636: Well, now `export_period_seconds` is unused :) 
Let's at least format it into the expression, like `f"0/{export_period_seconds} * * * * *`.

Also, you can make a helper function like `get_export_schedule(use_cron, export_period_seconds)` and then do something like this
```
{
    "export_directory": export_dir,
    ...
    **get_export_schedule(use_cron, export_period_seconds),
}
```

Python!

achulkov2 (changes requested):

Kontakter:
No build errors now!

pavel-bash:
@achulkov2 fixed.

yt/yt/server/queue_agent/ya.make
```diff
@@ -23,6 +23,7 @@ SRCS(
 
 PEERDIR(
     library/cpp/yt/phdr_cache
+    library/cpp/cron_expression
```
ItIsApachee, line 26: Don't think we need that since we already have yt/yt/client, which now has a peerdir to library/cpp/cron_expression.

yt/yt/tests/integration/queues/test_queue_agent.py
```diff
@@ -33,6 +33,17 @@
 
 import yt.environment.init_queue_agent_state as init_queue_agent_state
 
+
+##################################################################
+
+
+def get_export_schedule(use_cron, export_period_seconds):
```
ItIsApachee, line 40: Let's make it a QueueStaticExportHelpers method in https://github.com/ytsaurus/ytsaurus/blob/a99f69ebfef4992582b061bc89ff0d3127f80b0d/yt/yt/tests/library/yt_queue_agent_test_base.py#L389

yt/yt/tests/integration/queues/test_queue_agent.py
```diff
@@ -3995,6 +4021,103 @@ def assert_exported_table_count(count):
 
         self.remove_export_destination(export_dir)
 
+    @authors("pavel-bash")
+    def test_cron_annotation_schedule(self):
```
ItIsApachee, line 4025: I think you should just add use_cron parameter to test_several_exports and this test would be obsolete as it does not check anything that test_several_exports does not check.

yt/yt/tests/integration/queues/test_queue_agent.py
```diff
@@ -3995,6 +4021,103 @@ def assert_exported_table_count(count):
...
+        if getattr(self, "USE_OLD_QUEUE_EXPORTER_IMPL"):
+            pytest.skip()
+
+        _, queue_id = self._create_queue("//tmp/q")
+
+        export_dir = "//tmp/export"
+        self._create_export_destination(export_dir, queue_id)
+
+        export_seconds = [0, 17, 29, 45]  # Every ~15 seconds.
+        export_cron_expression = f"{",".join(str(s) for s in export_seconds)} * * * * *"
```
ItIsApachee, line 4067: I think you need to use single-quotes inside f-string or it may cause some issue for older python versions, which is used in some part of CI. I had this problem in the past.

yt/yt/tests/integration/queues/test_queue_agent.py
```diff
@@ -3995,6 +4021,103 @@ def assert_exported_table_count(count):
...
+
+        insert_rows("//tmp/q", [{"data": "vim"}])
+        self._flush_table("//tmp/q")
+        wait(lambda: len(ls(export_dir)) == 1)
+
+        insert_rows("//tmp/q", [{"data": "nano"}])
+        self._flush_table("//tmp/q")
+        wait(lambda: len(ls(export_dir)) == 2)
+
+        # Find out the seconds when the files were exported.
```
ItIsApachee, line 4087: Let's also check that all exports have correct periods in their name. That would check GetLastExportPeriod implementation.

ItIsApachee (changes requested):
Mostly OK, but needs some improvements in my opinion.

yt/yt/server/queue_agent/queue_exporter.cpp
```diff
@@ -78,25 +79,59 @@ namespace {
...
-    return GetExportUnixTsRange(now.Seconds(), exportPeriod).first;
+    if (exportConfig->ExportPeriod) {
+        return exportConfig->ExportPeriod->Seconds();
+    }
+
+    static const auto UTCTimezone = NDatetime::TTimeZone{};
+
+    TCronExpression cronExpression{*exportConfig->ExportCronSchedule};
+    auto nowCivilSecond = NDatetime::Convert(TInstant::Seconds(now), UTCTimezone);
+    return now - NDatetime::Convert(cronExpression.CronPrev(nowCivilSecond), UTCTimezone).Seconds();
```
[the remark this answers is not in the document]
pavel-bash: I agree with the naming and adding the comment, as it's indeed expected to always be the export timestamp, but isn't the `YT_VERIFY` too strong for it, causing the program to crash? Maybe an exception is enough?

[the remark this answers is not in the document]
ItIsApachee: Let's do an exception then.

ItIsApachee:
@pavel-bash let's add a test with some weird cron expression like `0 0 0 1 1 * 2026`. It's best we check that these kind of expression don't break queue agent.

pavel-bash:
> @pavel-bash let's add a test with some weird cron expression like `0 0 0 1 1 * 2026`. It's best we check that these kind of expressions don't break queue agent.

Do you want to just initialize it in one of the `test_queue_agent.py` tests? Such an expression is not going to export anything.

pavel-bash:
@ItIsApachee I've fixed all your comments except for the "weird" CRON expression, there I had a question.

yt/yt/server/queue_agent/queue_exporter.cpp
```diff
@@ -26,6 +26,7 @@
 #include <yt/yt/client/queue_client/config.h>
 
 #include <util/generic/map.h>
+#include <library/cpp/cron_expression/cron_expression.h>
```
ItIsApachee, line 29: Put library before util and separate them with an empty line

ItIsApachee (approved):
LGTM, let's fix include order in queue_exporter.cpp and then I'll work on merging it

ItIsApachee:
fyi: I wasn't aware of how you start PR importing, and meant to import it after include order fix

pavel-bash:
@ItIsApachee Done.

ItIsApachee (approved):

Kontakter:
Shipped!

Kontakter:
(my little help to @ItIsApachee)

Kontakter:
Shipped!

Kontakter:
Shipped!

achulkov2:
Hi, @ItIsApachee @Kontakter what is the status of this PR, the last import/discard messages are a bit confusing?

ItIsApachee:
> Hi, @ItIsApachee @Kontakter what is the status of this PR, the last import/discard messages are a bit confusing?

I'm working on merging it. Hopefully this PR would already be merged by the end of the day.

Kontakter:
Currently we have some non-trivial errors related to adding dependency `contrib/libs/cctz` to `yt/yt/client`, we are investigating the ways to resolve it

Kontakter:
We actually fix internal problems, caused by new dependency. Let rebase this PR to resolve conflict and we will merge it finally (I hope)

PR #1241, opened 2025-04-25
YQL-19747: Refactor sql/v1/complete usage of Future and Ptr

vityaman:
- [x] Chained futures, add `CompleteAsync` method (then will migrate the YDB CLI on it).
- [x] Removed deadlined and fallback NameService as unused
- [x] Annotate thread-safe methods with `const` and use `AtomicSharedPtr` for them.
- [x] Move `name` to `name/service` with backward compatibility with the YDB CLI.

---

`CompletionEngine` is left thread-unsafe because of the dependency chain `CompletionEngine -> LocalSyntaxAnalysis -> C3Engine` which is thread-unsafe, but readonly indexed data structures such as `Ranking` and `NameService` are annotated with const and distributed via shared pointers.

I removed deadlined and fallback name services because the first is stupid the second is ahead of its time and is better to be added later to keep interfaces as minimal as possible.

---

The migration on async complete plan:
1. Introduce CompleteAsync
2. Migrate clients on CompleteAsync
3. Make Complete to return Future
4. Migrate clients on Complete
5. Remove CompleteAsync

---

- Related to https://github.com/vityaman/ydb/issues/33
- Related to https://github.com/vityaman/ydb/issues/31


yql/essentials/sql/v1/complete/name/object/schema_gateway.h
```diff
@@ -40,10 +40,10 @@ namespace NSQLComplete {
 
     class ISchemaGateway {
     public:
-        using TPtr = THolder<ISchemaGateway>;
+        using TPtr = TAtomicSharedPtr<ISchemaGateway>;
```
vitstn, line 43: I would prefer either std::shared_ptr or, better, IntrusivePtr + TThrRefBase.

Unfortunately atomic shared ptr doesn't support weak pointers and always does extra allocation

yql/essentials/sql/v1/complete/name/service/name_service.h
```diff
@@ -0,0 +1,83 @@
...
+        }
+    };
+
+    struct TNameResponse {
+        TVector<TGenericName> RankedNames;
+    };
+
+    class INameService {
+    public:
+        using TPtr = TAtomicSharedPtr<INameService>;
```
vitstn, line 77: the same, std::shared_ptr or IntrusivePtr

yql/essentials/sql/v1/complete/name/service/static/ranking.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include <yql/essentials/sql/v1/complete/name/service/name_service.h>
+
+#include <util/generic/hash.h>
+
+namespace NSQLComplete {
+
+    class IRanking {
+    public:
+        using TPtr = TAtomicSharedPtr<IRanking>;
```
vitstn, line 13: ditto

vitstn (approved):

PR #1245, opened 2025-04-27
YQL-19747 Introduce SimpleSchemaGateway

vityaman:
Introduce the `SimpleSchemaGateway` to make it easier to implement `SchemaGateway`s. The idea is that actually existing schema providers really do not support filtration such as by name and type, so in practice they return us the whole list and we need to filter it by hand. The `SimpleSchemaGateway` to `SchemaGateway` adapter does this for us -- we only need to implement a path splitting and folder listing.

The other and important feature of the `SimpleSchemaGateway` is that it is simple to implement a caching decorator for it -- just store a mapping `Path -> [FolderEntry]`, while caching a `SchemaGateway` with filters is soooo not trivial.

I also added string constants for known folder entry types, because they should be documented somewhere.

---

- Related to `YQL-19747`
- Related to https://github.com/vityaman/ydb/issues/34

### DIFF
--- a/yql/essentials/sql/v1/complete/name/object/simple/schema_gateway.cpp
+++ b/yql/essentials/sql/v1/complete/name/object/simple/schema_gateway.cpp
@@ -1,0 +1,74 @@
+#include "schema_gateway.h"
+
+#include <util/charset/utf8.h>
+
+namespace NSQLComplete {
+
+    namespace {
+
+        class TSimpleSchemaGateway: public ISchemaGateway {
+        private:
+            static auto FilterByName(TString name) {
+                return [name = std::move(name)](auto f) {
+                    TVector<TFolderEntry> entries = f.ExtractValue();
+                    EraseIf(entries, [prefix = ToLowerUTF8(name)](const TFolderEntry& entry) {
+                        return !entry.Name.StartsWith(prefix);
+                    });
+                    return entries;
+                };
+            }
+
+            static auto FilterByTypes(TMaybe<THashSet<TString>> types) {
+                return [types = std::move(types)](auto f) {
+                    TVector<TFolderEntry> entries = f.ExtractValue();
+                    EraseIf(entries, [types = std::move(types)](const TFolderEntry& entry) {
+                        return types && !types->contains(entry.Type);
+                    });
+                    return entries;
+                };
+            }
+
+            static auto Crop(size_t limit) {
+                return [limit](auto f) {
+                    TVector<TFolderEntry> entries = f.ExtractValue();
+                    entries.crop(limit);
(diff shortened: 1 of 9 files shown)


vityaman:

yql/essentials/sql/v1/complete/name/object/simple/schema_gateway.cpp
```diff
@@ -0,0 +1,74 @@
...
+namespace NSQLComplete {
+
+    namespace {
+
+        class TSimpleSchemaGateway: public ISchemaGateway {
+        private:
+            static auto FilterByName(TString name) {
+                return [name = std::move(name)](auto f) {
+                    TVector<TFolderEntry> entries = f.ExtractValue();
+                    EraseIf(entries, [prefix = ToLowerUTF8(name)](const TFolderEntry& entry) {
```
vitstn, line 14: many underlying systems have case sensitive names

we could add bool TSimpleSchemaGateway::IsCaseSensitive() const = 0 to provide uniform filtering

yql/essentials/sql/v1/complete/name/object/simple/schema_gateway.h
```diff
@@ -0,0 +1,23 @@
...
+        TStringBuf NameHint;
+    };
+
+    class ISimpleSchemaGateway: public TThrRefBase {
+    public:
+        using TPtr = TIntrusivePtr<ISimpleSchemaGateway>;
+
+        virtual ~ISimpleSchemaGateway() = default;
+        virtual TSplittedPath Split(TStringBuf path) const = 0;
+        virtual NThreading::TFuture<TVector<TFolderEntry>> List(TString path) const = 0;
```
vitstn, line 18: List(TString folder)

yql/essentials/sql/v1/complete/name/object/simple/schema_gateway.cpp
vityaman: I thought that `aba` in the context of `aba/c/aba` is a name hint and not a strict prefix of a name. It seems that users usually are lazy to press on `SHIFT` and therefore should tend to write a name with lowercase letters, especially when they do not really know the table name. I think it is better to suggest `EnterpriseQualityTable` even when user types `enter`.

The improved algorithm for name hint filtration could be, for example `Contains(ToLower(hint))`.

yql/essentials/sql/v1/complete/name/object/simple/schema_gateway.h
vityaman: Renamed path to folder.

vitstn (approved):

PR #1246, opened 2025-04-27
YQL-19747 Enable custom NameSet and FrequencyData

vityaman:
Clients might want to use custom `NameSet` and `FrequencyData` in their environment, for example, to get their private UDFs and to have more relevant ranking that includes their private UDF and respect their usage pattern.

To achieve this goal I decided to load pure `NameSet` and `FrequencyData` and provide functions for pruning. Also I checked defaults and decided that it is more common for a client to create a `StaticNameService` from pure `NameSet` and `FrequencyData` to keep their pruning consistent.

--- 

- Related to `YQL-19747`


vityaman:

vitstn (approved):

vitstn (approved):

PR #1248, opened 2025-04-28
YQL-19616: Introduce sql/v1/highlight

vityaman:
## Intro

This Pr is an introduction of `sql/v1/highlight` module. It contains Highlighting Specification object and Highlighter reference implementation. 

## About GenericLexer

As highlighting is actually a tokenization, but with other token types, then the same abstract lexer can be used for both highlighting and lexical analysis. This is why I created the GenericLexer. It takes a lexer grammar as a sequence of pairs (TokenName, Matcher), where Matcher is a function that takes an input prefix and maybe returns a matched token content. 

The GenericLexer algorithm is so simple. It just goes thought the token definitions and tries to match each one. The result is the first found longest one.

There are 2 mappings `TLexerGrammar -> TGenericLexerGrammar` and `TLexerGrammar -> THighlighting -> TGenericLexerGrammar`, so both `sql/v1/lexer/regex` and `sql/v1/highlight` was implemented via the GenericLexer.

## Testing

Test Suite expressed in JSON was added to be usable outside the C++ codebase, for example, to test TypeScript implementation.

## TRegexPattern

The JSON format of `THighlighting` will be created in future PRs.

```cpp
    struct TRegexPattern {
        TString Body;
        TString After = "";
        bool IsCaseInsensitive = false;
        bool IsLongestMatch = true;
    };
```

TRegexPattern let us describe tokens with lookahead. 

I found that it can be translated to VIM using

```
:help pattern
--
/\zs|	\zs	\zs	anything, sets start of match
/\ze|	\ze	\ze	anything, sets end of match
```

VSCode also supports `(?=..)`.

Case insensitivity is available in VSCode via `(?i)` and in Vim via `\c`.

`LongestMatch` is always true in VSCode and Vim, but it needs to be false for correct matching of COMMENT and STRING_VALUE.

---

- Related to `YQL-19616`
- Related to https://github.com/vityaman/ydb/issues/10
- Related to https://github.com/ydb-platform/ydb/issues/15129


vityaman:
@vitstn

vityaman:
@vitstn, fixed review comments. What is especially important is that I removed the usage of `set_longest_match` (the problem was the difference in semantics of `|` in Regexes and ANTLR). Now `TPattern = (Body, After = "", IsCaseSensitive = false)`.

vityaman:
@Raubzeug, to begin with suggested to...

- [ ] Add a pseudocode of the highligher algorithm
- [ ] Add a pseudocode of the test suite driver and to describe the test suite in more details
- [ ] Try to generate a well textual description from pseudocodes using an AI

vityaman:
@vitstn , fixed the TODO comments. Decided to provide `maxErrors` instead of callback-driven execution to be more consistent with `sql/v1/lexer` API. Dark braces color fixed too.

vitstn (approved):

yql/essentials/sql/v1/highlight/README.md
```diff
@@ -0,0 +1,217 @@
...
+
+- `Highlighting Token` is a text fragment matched with a `Highlighting Unit`.
+
+- `Highlighter` is an function parametrized by `Highlighting` transforming a text into a stream of `Highlighting Token`s.
+
+- `Theme` is a mapping from `Highlighting Unit` to a `Color`.
+
+## Highlighting Unit
+
+Here are listed all highlighting units.
```
vitstn, line 21: a note for the next PR:
I would clarify that this is an example of units. They will evolve and should be taken from the json specification.

Only comments are always present, as they may require special processing

yql/essentials/sql/v1/highlight/README.md
```diff
@@ -0,0 +1,217 @@
...
+
+    if text is empty:
+      return Nothing
+
+    text = text[1:]
+```
+
+## Highlighting JSON Example
+
+The highlighting can be generated using the `yql_highlight` tool in JSON format.
```
vitstn, line 133: highlighting specification

yql/essentials/sql/v1/highlight/README.md
```diff
@@ -0,0 +1,217 @@
...
+}
+```
+
+Where the first element is the SQL text to highlight and the second one is a string where each character represents the highlighting unit kind for each character in the input.
+
+Here's the table representation of the unit kind to character mapping:
+
+| Unit Kind                 | Character |
+| ------------------------- | --------- |
+| keyword                   | K         |
```
vitstn, line 178: I would add this mapping to the specification. 
      "kind":"type-identifier",
       >>> "symbol":"T"
       "patterns": [
       
to write generic test driver

vitstn:
broken checks:
yql/essentials/tools/yql_highlight/ya.make  (not found in "yql/essentials/tools/ya.make")

vitstn:
broken tests
image

typical errors:
E   Errors:
E   b'<main>:0:274: Error: no candidates, skipping ERROR\n<main>:0:328: Error: no candidates, skipping ERROR\n<main>:0:385: Error: no candidates, skipping ERROR\n'

yql/essentials/sql/v1/highlight/README.md
vityaman: I wanted to name it just `Highlighting` because the name `HighlightingSpecification` is too long.

yql/essentials/sql/v1/highlight/README.md
vityaman: Hmmm. I think it is better to rewrite testsuite for a more verbose format like

```
{
  "input: "...",
  "expected": [
    { "kind": "...", "content": "..." },
  ]
}
```

I think this one can be generalized for a just a lexer testsuite. I propose to do it in a separate PR.

vityaman:
@vitstn, you can try to run tests again.

vitstn (approved):

PR #1252, opened 2025-04-30
YQL-19747 Introduce Union and split Static NameServices

vityaman:
Essentially we need the `UnionNameService` to integrate dynamic name services into the `sql/v1/complete`. This decorator queries multiple name services and merges their responses. We will use it to union static `NameServices` and the `SchemaNameService` (and the `ClusterNameService`, and so on).

It does not deal with exceptions and fails the whole request on some subtask failed. Client should prepare robust children (later `SwallowingNameService` will be added to return an empty `NameResponse` on errors to provide best-effort dynamic object loading).

`StaticNameService` was split into micro-`NameService`s :)

`NameConstraints` are extracted to provide name qualification via `Qualified` and `Unqualified` methods. This is needed because depending on a context `NameService` can return unqualified names (for example, on `PRAGMA yt.#`). As internal indexes for scanning and ranking are built on a sorted list of, then `Ranking` actually needs an ability to get fully-qualified names, so now it via the `NameConstraints`. 

Also this design potentially let us to improve internal indexes by using partitioning by a namespace. Other option was to make `PragmaName` and `FunctionName` more structured via adding a separate field for a namespace, but it seems to me that it will force to do more parsing from indexes results to a `Name`. Anyway this is an internal component so it can be changed if needed.  I still doubt this decision because structured `PragmaName { Namespace, Identifier }` seems to be cleaner and there should be no noticeable overhead because of COW strings.

### DIFF
--- a/yql/essentials/sql/v1/complete/name/service/name_service.cpp
+++ b/yql/essentials/sql/v1/complete/name/service/name_service.cpp
@@ -23,32 +23,13 @@ namespace NSQLComplete {
 
     } // namespace
 
-    TPragmaName TNameConstraints::Qualified(TPragmaName unqualified) const {
-        SetPrefix(unqualified.Indentifier, ".", *Pragma);
-        return unqualified;
-    }
-
-    TPragmaName TNameConstraints::Unqualified(TPragmaName qualified) const {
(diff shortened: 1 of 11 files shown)


vityaman:
@vitstn

vitstn (approved):

PR #1253, opened 2025-04-30
YQL-19747 Fix cursor position detection

vityaman:
Fixed a bug within cursor position detection algorithm. 

Also improved an infrastructure at cursor and token list related code. This is required for a `ID_QUOTED` completion algorithm.

---

- Related to `YQL-19747`
- Related to https://github.com/ydb-platform/ydb/issues/9056
- Related to https://github.com/vityaman/ydb/issues/35
- Related to https://github.com/vityaman/ydb/issues/14

PR #1257, opened 2025-05-02
YQL-19747 Complete folder, table and cluster names

vityaman:
---

- Related to `YQL-19747`
- On top of https://github.com/ytsaurus/ytsaurus/pull/1253
- Related to https://github.com/vityaman/ydb/issues/35
- Related to https://github.com/vityaman/ydb/issues/40


yql/essentials/sql/v1/complete/name/cluster/discovery.h
```diff
@@ -0,0 +1,21 @@
...
+    struct TClusterCatalog {
+        TVector<TString> Instances;
+    };
+
+    class IClusterDiscovery: public TThrRefBase {
+    public:
+        using TPtr = TIntrusivePtr<IClusterDiscovery>;
+
+        virtual ~IClusterDiscovery() = default;
+        virtual NThreading::TFuture<TClusterCatalog> Query() const = 0;
```
vitstn, line 18: а где тут имя кластера в аргументах?

yql/essentials/sql/v1/complete/name/cluster/discovery.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include <library/cpp/threading/future/core/future.h>
+
+#include <util/generic/ptr.h>
+
+namespace NSQLComplete {
+
+    struct TClusterCatalog {
+        TVector<TString> Instances;
```
vitstn, line 10: I guess it's a cluster list? Instance is a bit ambigous

yql/essentials/sql/v1/complete/name/object/dispatch/schema.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+        class TSchema: public ISchema {
+        public:
+            explicit TSchema(THashMap<TString, ISchema::TPtr> mapping)
+                : Mapping_(std::move(mapping))
+            {
+            }
+
+            NThreading::TFuture<TListResponse> List(const TListRequest& request) const override {
+                auto iter = Mapping_.find(request.Cluster);
```
vitstn, line 15: an incoming cluster can be empty (for YDB queries it's a norm)
we should use something like 'default' schema if it's specified

vityaman: For an empty cluster name it makes sense for me to have an entry `{"", defaultSchema}` at the `THashMap<TString, ISchema::TPtr> Mapping_`, because an empty strings is sensible for an "in case no cluster name provided" intention representation, so the client should add this entry to `mapping` on the `DispatchSchema` construction (what is actually done in unit tests). If the cluster name is not specified, but required (no such entry in mapping), it is okay to raise an error as well as in case when non-existing cluster name was provided, what the current implementation does.

yql/essentials/sql/v1/complete/name/cluster/discovery.h
vityaman, line 1: Caching will be added in separate PRs.

vityaman:
Caching (alternative to `TAsyncExpiringCache`)
- Batch update for cached entries
- Use Tick
- Think about adaptive caching duration (but think about upper bound)
- Eviction of unused for a long time (like Clock eviction policy)
- Entry should live for at least 2 ticks

vitstn (approved):

PR #1258, opened 2025-05-04
Fix backticks at doc page

vityaman:
Now formatting is broken at https://ytsaurus.tech/docs/en/yql/syntax/use.

image


ogorbacheva:
Shipped!

PR #1266, opened 2025-05-08
Do not umount slot location directory in TSimpleJobDirectoryManager

koct9i:
Slot location directory itself should not be managed and umounted.


Kontakter (approved):

Kontakter:
For the history, the bug was introduced here https://github.com/ytsaurus/ytsaurus/commit/c5f8e5af502

PR #1267, opened 2025-05-08
YQL-19747: Fix TSimpleSchema path mutation

vityaman:
`TSimpleSchema` implementation was incorrect at least as it does not know anything about the path structure and therefore have no right to modify path with `/` symbols.

---

- Related to `YQL-19747`
- Related to https://github.com/vityaman/ydb/issues/34
- Related to https://github.com/ydb-platform/ydb/pull/18146

### DIFF
--- a/yql/essentials/sql/v1/complete/name/object/simple/static/schema.cpp
+++ b/yql/essentials/sql/v1/complete/name/object/simple/static/schema.cpp
@@ -27,6 +27,10 @@ namespace NSQLComplete {
             }
 
             NThreading::TFuture<TVector<TFolderEntry>> List(TString folder) const override {
+                if (!folder.StartsWith('/')) {
+                    folder.prepend('/');
+                }
+
                 TVector<TFolderEntry> entries;
                 if (const auto* data = Data_.FindPtr(folder)) {
                     entries = *data;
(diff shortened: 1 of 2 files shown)


yql/essentials/sql/v1/complete/name/object/simple/static/schema.cpp
```diff
@@ -27,6 +27,10 @@ namespace NSQLComplete {
             }
 
             NThreading::TFuture<TVector<TFolderEntry>> List(TString folder) const override {
+                if (!folder.StartsWith('/')) {
```
vitstn, line 30: it blocks usage for YT cluster
home/foo/table is a valid folder, //home/foo/table too, but /home/foo/table isn't

vityaman: I thought that `StaticSimpleSchema` is a dummy implementation for a testing purposes. For the YT the other implementation should be created.

vitstn (approved):

yql/essentials/sql/v1/complete/name/object/simple/static/schema.cpp
vitstn: oh, you have deleted the special case for '/' in the other method

PR #1272, opened 2025-05-13
Fix multilevel optionals.

Tony-Romanov:
Type: fix
Component: query-tracker


yt/yql/tests/agent/test_simple.py
```diff
@@ -228,6 +228,13 @@ def test_exotic_types(self, query_tracker, yql_agent):
                "Yson": [7, 'str', -3.14]
                }])
 
+    @authors("a-romanov")
+    def test_double_optional(self, query_tracker, yql_agent):
```
Krock21, line 232: Could you please add a more complex test with combination of inner maps and lists, like `[[[null],[null]],[[null],[null]]]`

To test logic with OptionalLevels and Depth

nepal (approved):

yt/yql/agent/data_builder.cpp
```diff
@@ -342,6 +359,18 @@ void TDataBuilder::OnPg(TMaybe<TStringBuf> value, bool /*isUtf8*/)
     }
 }
 
+void TDataBuilder::AddBeginOptional() {
```
Krisha11, line 362: { 
should be on the next line

yt/yql/agent/data_builder.cpp
```diff
@@ -342,6 +359,18 @@ void TDataBuilder::OnPg(TMaybe<TStringBuf> value, bool /*isUtf8*/)
...
+    if (OptionalLevels_.top() > 1) {
+        BeginList();
+    }
+}
+
+void TDataBuilder::AddEndOptional() {
```
Krisha11, line 368: {

Krisha11 (approved):

PR #1282, opened 2025-05-19
yt/cri-api: tune ya.make to generate headers

koct9i:
Some tools are stumble on missing api.deps.pb.h in generated headers.


savnadya:
Shipped!

PR #1284, opened 2025-05-19
fix go-sdk: allow empty json object in yson2json module

lesf0:
* Changelog entry
Type: fix
Component: go-sdk

Fixed a bug in go yson2json module, which led to inability to parse an empty json object


tinarsky (approved):
LGTM

PR #1269, opened 2025-05-10
YQL-19747: Introduce the ManagedCache

vityaman:
- Related to `YQL-19747`
- Related to https://github.com/ytsaurus/ytsaurus/issues/1263 (motivation is there)
- Related to https://github.com/vityaman/ydb/issues/41 (requirements are there)

PR #1302, opened 2025-05-28
YQL-19747: Introduce locally cached SimpleSchema

vityaman:
- Add the cache async interface to enable remote implementation via some KV storage
- `Bounded Memory` (vvvv) requirement is satisfied by using `TLFUCache` from the `library/cpp/cache`. I tried to use a thread safe version, but it uses shared pointers at an interface, which are inconvenient to use in this case. Also should not really provide performance benefits as both `Get` and `Update` are mutating operations, due to eviction list promotion.
- `Multitenancy` (vvvv) requirement of cached decorator is satisfied by adding a `Zone` field to caching key.
- `Limited Staleness` (vvvv) requirement is satisfied by a storage of depth instant of an entry. When an entry is expired, flagged as `IsExpired` entry is returned and then background update is initiated in `TCachedQuery`.
- `Negativity` (vvvv) requirement can not be satisfied because of an absence of clear semantics of a failure at `SimpleSchema`, e.g. should it return just an empty list (then the requirement is satisfied) or throw an exception (the bad way I think, it is better to return some `Result` container or return `TIssues`).
- `Inactivity` (vvvv) requirement is satisfied by the implementation of `Limited Staleness`.
- `Simplicity` (vvvv) requirement is satisfied.
- `Generality` (mine) requirement is satisfied.
- `Eviction` (mine) requirement is the same as `Bounded Memory` (vvvv).
- `Refresh` (mine) requirement was rejected due to the `Simplicity` (vvvv) requirement.
- `No Queueing` (mine) and `No Batching` (mine) requirements are satisfied.
- `Inconsistency` (mine) property exists.

---

- Related to `YQL-19747`
- Related to https://github.com/vityaman/ydb/issues/41
- Related to https://github.com/ytsaurus/ytsaurus/pull/1269


yql/essentials/sql/v1/complete/name/cache/cached_ut.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+#include <library/cpp/time_provider/monotonic_provider.h>
+
+using namespace NSQLComplete;
+
+Y_UNIT_TEST_SUITE(CachedQueryTests) {
+
+    Y_UNIT_TEST(OnExpired_WhenApplied_ThenDefferedUpdateAndReturnOld) {
+        size_t queried = 0;
+        auto cache = MakeLocalCache<int, int>(
```
vitstn, line 15: please use the string type as a value

yql/essentials/sql/v1/complete/name/cache/cached_ut.cpp
```diff
@@ -0,0 +1,30 @@
...
+            NMonotonic::CreateDefaultMonotonicTimeProvider(), {.TTL = TDuration::Zero()});
+        auto cached = TCachedQuery<int, int>(cache, [&](const int& key) {
+            queried += 1;
+            return NThreading::MakeFuture<int>(key);
+        });
+        cache->Update(1, 2);
+
+        int value = cached(1).GetValueSync();
+
+        Y_ENSURE(value, 2);
```
vitstn, line 25: UNIT_ASSERT_VALUES_EQUAL here and below

yql/essentials/sql/v1/complete/name/cache/local/cache.h
```diff
@@ -0,0 +1,90 @@
...
+                        entry.IsExpired = (it->Death < Clock_->Now());
+                    }
+                }
+                return NThreading::MakeFuture(std::move(entry));
+            }
+
+            NThreading::TFuture<void> Update(const TKey& key, TValue value) const override {
+                TCell entry = {
+                    .Value = std::move(value),
+                    .Death = Clock_->Now() + Config_.TTL,
```
vitstn, line 62: 'Deadline' term is more common here

yql/essentials/sql/v1/complete/name/cache/local/cache.h
```diff
@@ -0,0 +1,105 @@
...
+
+            struct TCell {
+                TValue Value;
+                NMonotonic::TMonotonic Deadline;
+                size_t KeySize = 0;
+            };
+
+            struct TCellSizeProvider {
+                size_t operator()(const TCell& entry) const {
+                    return TValueSizeProvider()(entry.Value) + entry.KeySize;
```
vitstn, line 36: Let's cache size from TValueSizeProvider() + KeySize as CellSize

vityaman:
TODO: `LFU` vs `LRU`.

vityaman:
TODO: size interpreted as bytes.

vityaman:

vityaman:
@vitstn, fixed the build.

vitstn (approved):

PR #1306, opened 2025-05-31
Re-enable Clang Tidy bugprone-use-after-move check

vityaman:
Seems that the issue https://github.com/llvm/llvm-project/issues/57758 is resolved.

### DIFF
--- a/build/config/tests/clang_tidy/config.yaml
+++ b/build/config/tests/clang_tidy/config.yaml
@@ -2,14 +2,7 @@ Checks: >
   -*,
   arcadia-typeid-name-restriction,
   arcadia-ascii-compare-ignorecase,
-  # bugprone-use-after-move is disabled as it does not treat . as a synchornization point, as guaranteed by C++17
-  # See the discussion in
-  # https://github.com/llvm/llvm-project/issues/57758
-  # and
-  # https://stackoverflow.com/questions/73729536/is-clang-tidy-version-14-right-with-bugprone-use-after-move-error-on-this-co
-  #
-  # This check should be re-enabled, once the upstream issue is solved.
-  # bugprone-use-after-move,
+  bugprone-use-after-move,
   performance-implicit-conversion-in-loop,
   readability-identifier-naming,
 CheckOptions:


vityaman:
@georgthegreat, maybe you should know better, what actually should be done with it.

georgthegreat:
I think that we should import this and check how many actual bugs were introduced since these checks were disabled.

I do not quite remember the actual semantics of the above config though.
I think the check should be also enabled in project-specific config...

vityaman:
@georgthegreat, 

> I do not quite remember the actual semantics of the above config though.

It seems to me that this `config.yaml` is used by `ya ide --cpp` to emit `.clang-tidy` config at the root of the project. The `.clang-tidy` is detected by `clangd`, so warnings (like identifier naming) are shown in VSCode window.

> I think the check should be also enabled in project-specific config...

Is it possible to setup a project-specific config using the Yatool? I did not found any subcommands and flags in `ya` to run the `clang-tidy`. Or you mean that I need just to put my custom `.clang-tidy` config file in a module and commit it to VCS?

---

I just dream about using some code analysis tools (at least linter) to keep a code quality. Looking at a poorly formatted committed code with improperly named identifiers makes me sad. This is why I care about this `build/config/tests/clang_tidy/config.yaml`. I understand that it is impossible to fix styling in an existing code, but want to keep the style at least at new modules (at least that are maintained mostly by me), so enforcing `ya style` and `clang-tidy` there seems to be a good starting point to make the codebase better.

Kontakter:
Shipped!

build/config/tests/clang_tidy/config.yaml
```diff
@@ -2,14 +2,7 @@ Checks: >
...
   arcadia-ascii-compare-ignorecase,
-  # bugprone-use-after-move is disabled as it does not treat . as a synchornization point, as guaranteed by C++17
-  # See the discussion in
-  # https://github.com/llvm/llvm-project/issues/57758
-  # and
-  # https://stackoverflow.com/questions/73729536/is-clang-tidy-version-14-right-with-bugprone-use-after-move-error-on-this-co
-  #
-  # This check should be re-enabled, once the upstream issue is solved.
-  # bugprone-use-after-move,
+  bugprone-use-after-move,
```
vitstn, line 5: I looks like a company-wide setting, but I will try it

vitstn:
Unfortunately, it breaks a lot of tests. We can't merge it right now

vityaman:
- Try to build a clang format config not breaking the build on more than 0.1% of the code.
- This config should cover the whole `yql/essentials`.
- Need to add local for `yql/essentials` `clang-format` config.
- No more than 500-1000 LOC on each check added.
- Need to figure out how to use custom `clang-format` config.
- For example, namespace comment checks.

Arcadia Rules:
- `PEERDIR` should be reachable
- No ending spaces
- EOL at the end of file
- No TABs

vityaman:
Will try to use style checks locally at `yql/essentials`.

PR #1311, opened 2025-06-02
yt/core/bus: fail connection if client TLS certificate is required but not provided

koct9i:
SSL flag "SSL_VERIFY_PEER" indeed requests and verifies client certificate.
But it does _nothing_ if client have not provided any TLS certificate.
Safe behaviour needs also flag "SSL_VERIFY_FAIL_IF_NO_PEER_CERT".

* fail client certificate verification without client certificate
* add additional check that certificate is available after TLS handshake
* log peer certificate fingerprint to track connectivity
* add early failure on client side to avoid wasting time on TLS handshake
* add test cases for mutual TLS certificate verification

---

* Changelog entry
Type: fix
Component: misc-server

Fix mTLS in bus RPC


koct9i:
This is not a security issue because client certificates actually never worked.
Before #1066 client connection is never used client certificates.

yalekseev:
Shipped!

PR #1318, opened 2025-06-03
Return back of binary representation of decimals in results.

Tony-Romanov:
Type: fix
Component: query-tracker


vitstn:
the format is already defined here
https://github.com/ydb-platform/ydb/blob/main/yql/essentials/docs/en/types/json.md#decimal-decimal
https://github.com/ydb-platform/ydb/blob/main/yql/essentials/docs/ru/types/json.md#int-uint-float-double-decimal-numbers

so we won't accept this PR

Issue #1233, opened 2025-04-23
Support sidecars in job containers


gritukan:
https://docs.google.com/document/d/1H-Jm8S0evQbKUiOKCcu0V1r0NTRfm7FDrwHrSli1VJQ/edit?usp=sharing

PR #1324, opened 2025-06-12
#1233 Support sidecars in job containers

pavel-bash:
# Description

This PR aims to provide a `sidecar` functionality, so that a set of containers may run as part of the user job, sharing the same space/memory and/or resources. They are launched together with the user job and stopped when the user job finishes. They can be used, for instance, to gather logs from the job, monitor it and many other things.

The logic starts at the proto definition of the job, where the `sidecar`-s definition was added, following the docs in the description of the issue. Those definitions are passed around until they arrive into a `JobProxy` which ultimately controls the lifetime of the sidecars. There's also a tracking mechanism which allows to restart the sidecars in case they've exited or failed, or even to fail the whole job if the corresponding `RestartPolicy` is set.

This PR also required some changes (questionable?) in `cri_executor` - some of the structures were turned into `YSON` structures, so that they can be forwarded to job proxy.

Lastly, this PR allows the future extension with other container technologies like Porto. For that, the relevant methods `job_proxy/environment` are to be implemented.


pavel-bash:
@achulkov2 @gritukan PR is ready for review.

pavel-bash:
I am closing this PR to open another one from another branch as I needed to squash-rebase the commits.

PR #1335, opened 2025-06-23
operation_id should be optional in get_job

faucct:
In our installations of YTsaurus users don't have direct network access to the exec nodes. This makes it harder to expose Web UI of services running in Vanilla operations, like Spark.

For that, we've developed a proxy, which accepts HTTP traffic, checks YT authentication/authorization and forwards the traffic to exec nodes.

This proxy needs to somehow know where to route the traffic (job and its port from `port_count`).

Right now this is done by prefixing the URL path with exec node address and physical port, though it could be `operation_id` + `job_id` instead of the exec node address.

But prefixing paths is not totally safe because of browser isolation (CORS, cookies, probably something else).
Also initially I've thought that it would not be possible to adapt the Spark application inside job to the URLs with prefixed paths, but this has managed to be not a problem after all.

A better solution would be to use wildcard-hosts, something like `*.proxy.yt...`. This way those hosts would be more isolated from each other.

The problem is that the wildcard max length is 63 characters, so it can't fit both `operation_id` and `job_id`. If the traffic would be http, we could've used multiple wildcards, but with https we can't sign an SSL certificate for such a host.
So, if it would be possible to skip `operation_id` from those requests, then the wildcard host could fit just `job_id` and port ID.
Still, it is possible to only put `operation_id` in the wildcard for the isolation and put the rest needed for resolving (`job_id`, port) inside the path prefix, though a single `job_id` would be more convenient to use.

So, this pull request is about giving a way to resolve running jobs by their id without the `operation_id`, so the proxy could proxy the traffic to those.
That's why I only care about running jobs.

We can live without this feature, so if it is too hard to resolve completed jobs, then I won't pursue it.

### DIFF
--- a/yt/yt/client/driver/scheduler_commands.h
+++ b/yt/yt/client/driver/scheduler_commands.h
@@ -30,7 +30,7 @@ public:
         registrar.Postprocessor([] (TThis* command) {
             auto operationId = command->OperationId;
             if (operationId && command->OperationAlias.has_value() ||
-                !operationId && !command->OperationAlias.has_value())
+                !operationId && !command->OperationAlias.has_value() && !command->GetJobId())
             {
                 THROW_ERROR_EXCEPTION("Exactly one of \"operation_id\" and \"operation_alias\" should be set")
                     << TErrorAttribute("operation_id", command->OperationId)
@@ -39,7 +39,7 @@ public:
 
             if (command->OperationId) {
                 command->OperationIdOrAlias = command->OperationId;
-            } else {
+            } else if (command->OperationAlias) {
                 command->OperationIdOrAlias = *command->OperationAlias;
             }
         });
@@ -48,6 +48,10 @@ public:
 protected:
     // Is calculated by OperationId and OperationAlias.
     NScheduler::TOperationIdOrAlias OperationIdOrAlias;
+
+    virtual NScheduler::TJobId GetJobId() const {
+        return {};
+    }
 
 private:
     NScheduler::TOperationId OperationId;
@@ -134,6 +138,7 @@ private:
 
     void DoExecute(ICommandContextPtr context) override;
     bool HasResponseParameters() const override;
+    NScheduler::TJobId GetJobId() const override;
 };
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -164,6 +169,7 @@ private:
     NJobTrackerClient::TJobId JobId;
 
     void DoExecute(ICommandContextPtr context) override;
+    NScheduler::TJobId GetJobId() const override;
 };
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -226,6 +232,7 @@ private:
     NJobTrackerClient::TJobId JobId;
 
     void DoExecute(ICommandContextPtr context) override;
+    NScheduler::TJobId GetJobId() const override;
 };
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/yt/yt/client/scheduler/operation_id_or_alias-inl.h
+++ b/yt/yt/client/scheduler/operation_id_or_alias-inl.h
@@ -33,7 +33,8 @@ void FromProto(TOperationIdOrAlias* operationIdOrAlias, const TProtoClass& enclo
             break;
         }
         case TProtoClass::OperationIdOrAliasCase::OPERATION_ID_OR_ALIAS_NOT_SET: {
-            THROW_ERROR_EXCEPTION("None of operation id and operation alias is set in oneof OperationIdOrAlias proto");
+            operationIdOrAlias->Payload = {};
+            break;
         }
     }
 }
@@ -45,7 +46,9 @@ void ToProto(TProtoClassPtr enclosingProtoMessage, const TOperationIdOrAlias& op
 
     Visit(operationIdOrAlias.Payload,
         [&] (const TOperationId& operationId) {
-            ToProto(enclosingProtoMessage->mutable_operation_id(), operationId);
+            if (operationId) {
+                ToProto(enclosingProtoMessage->mutable_operation_id(), operationId);
+            }
         },
         [&] (const TString& alias) {
             enclosingProtoMessage->set_operation_alias(alias);
--- a/yt/yt/tests/integration/scheduler/test_get_job.py
+++ b/yt/yt/tests/integration/scheduler/test_get_job.py
@@ -18,7 +18,7 @@ from yt_operations_archive_helpers import (
 
 import yt.environment.init_operations_archive as init_operations_archive
 
-from yt.common import date_string_to_datetime, uuid_to_parts, parts_to_uuid, update, date_string_to_timestamp_mcs
+from yt.common import date_string_to_datetime, uuid_to_parts, parts_to_uuid, update, date_string_to_timestamp_mcs, YtError
 
 from flaky import flaky
 
@@ -164,6 +164,7 @@ class _TestGetJobCommon(_TestGetJobBase):
 
         self._check_get_job(op.id, job_id, before_start_time, state="running", has_spec=None,
                             pool="my_pool", pool_tree="default")
+        assert get_job(None, job_id)["operation_id"] == op.id
 
         @wait_no_assert
         def correct_stderr_size():
@@ -175,6 +176,8 @@ class _TestGetJobCommon(_TestGetJobBase):
 
         self._check_get_job(op.id, job_id, before_start_time, state="failed", has_spec=True,
                             pool="my_pool", pool_tree="default")
+        with pytest.raises(YtError, match="Allocation .* not found"):
+            assert get_job(None, job_id)
 
         job_info = retry(lambda: get_job(op.id, job_id))
         assert job_info["fail_context_size"] > 0
--- a/yt/yt/tests/integration/scheduler/test_get_job.py
+++ b/yt/yt/tests/integration/scheduler/test_get_job.py
@@ -176,14 +176,13 @@ class _TestGetJobCommon(_TestGetJobBase):
 
         self._check_get_job(op.id, job_id, before_start_time, state="failed", has_spec=True,
                             pool="my_pool", pool_tree="default")
-        with pytest.raises(YtError, match="Allocation .* not found"):
-            assert get_job(None, job_id)
 
         job_info = retry(lambda: get_job(op.id, job_id))
         assert job_info["fail_context_size"] > 0
         events = job_info["events"]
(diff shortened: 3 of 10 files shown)


renadeen:
In progress

yt/yt/ytlib/api/native/client_impl.cpp
```diff
@@ -365,6 +366,23 @@ void TClient::InitChannelsOrThrow(EMasterChannelKind kind, TCellTag cellTag)
...
 
+TOperationId TClient::GetJobOperation(TJobId jobId) const
+{
+    auto allocationId = NScheduler::AllocationIdFromJobId(jobId);
+
+    YT_LOG_DEBUG(
+        "Requesting allocation brief info (AllocationId: %v)",
+        allocationId);
+
+    return WaitFor(NApi::NNative::GetAllocationBriefInfo(
```
bystrov-serg: This should only work for running operations. For completed operations you can, for example, make lookup rows/select rows  to "jobs" in operations archive with key - job id and extract operation id from the result

yt/yt/tests/integration/scheduler/test_get_job.py
```diff
@@ -175,6 +176,8 @@ def correct_stderr_size():
 
         self._check_get_job(op.id, job_id, before_start_time, state="failed", has_spec=True,
                             pool="my_pool", pool_tree="default")
+        with pytest.raises(YtError, match="Allocation .* not found"):
```
bystrov-serg: I don't think this should be a desired behavior :) As a said in the other comment, if we want to allow optional operation id, we should correctly process completed operations

bystrov-serg (commented):
Can you give a more detailed description of the problem? I didn't get how optional op id will help

faucct:
> Can you give a more detailed description of the problem? I didn't get how optional op id will help

I have detailed the motivation in the description.

yt/yt/ytlib/api/native/client_impl.cpp
faucct: Would not this be way too slow?

bystrov-serg: It would not be slow, as `job_id` is a key column in the `jobs` table.
Even if you don't need this feature for completed jobs, I would still insist on supporting it for completed jobs for API consistency

bystrov-serg: I was wrong, we have the table `operation_ids` in `operations_archive` to get `operationd_id` from `job_id`

faucct: Okay, but it will still be fast, right?

bystrov-serg: Using the table `jobs` for this purpose would not be fast, using `operation_ids` to get operation_id is okay

achulkov2:
After some discussions internally, we decided that we won't be pursuing this for the time being. I will close this PR and we can reopen it when/if needed.

yt/yt/ytlib/api/native/client_impl.cpp
faucct: This ended up to be easy, so I have brought the changes anyway.

faucct:
Once again, we can live without this feature, so this is not urgent, but I have decided to finish it, so, please review it, if you have some time and are open to such a change.

bystrov-serg:
I'll review this pr soon

yt/yt/tests/integration/scheduler/test_get_job.py
```diff
@@ -18,7 +18,7 @@
 
 import yt.environment.init_operations_archive as init_operations_archive
 
-from yt.common import date_string_to_datetime, uuid_to_parts, parts_to_uuid, update, date_string_to_timestamp_mcs
+from yt.common import date_string_to_datetime, uuid_to_parts, parts_to_uuid, update, date_string_to_timestamp_mcs, YtError
```
bystrov-serg: No need for YtError here anymore

yt/yt/tests/integration/scheduler/test_get_job_artifacts.py
```diff
@@ -6,7 +6,7 @@
     set, remove, exists, create_tmpdir, create_user, make_ace, insert_rows, select_rows, lookup_rows,
     read_table, write_table, map, reduce, map_reduce,
     sort, list_jobs, get_job_input,
-    get_job_stderr, get_job_stderr_paged, get_job_spec, get_job_input_paths,
+    get_job, get_job_stderr, get_job_stderr_paged, get_job_spec, get_job_input_paths,
```
bystrov-serg: No need for get_job here anymore

yt/yt/client/driver/scheduler_commands.h
```diff
@@ -30,7 +30,7 @@ class TSimpleOperationCommandBase
         registrar.Postprocessor([] (TThis* command) {
             auto operationId = command->OperationId;
             if (operationId && command->OperationAlias.has_value() ||
-                !operationId && !command->OperationAlias.has_value())
+                !operationId && !command->OperationAlias.has_value() && !command->GetJobId())
             {
                 THROW_ERROR_EXCEPTION("Exactly one of \"operation_id\" and \"operation_alias\" should be set")
                     << TErrorAttribute("operation_id", command->OperationId)
```
bystrov-serg, line 36: Probably we should update error message

yt/yt/client/scheduler/operation_id_or_alias-inl.h
```diff
@@ -33,7 +33,8 @@ void FromProto(TOperationIdOrAlias* operationIdOrAlias, const TProtoClass& enclo
             break;
         }
         case TProtoClass::OperationIdOrAliasCase::OPERATION_ID_OR_ALIAS_NOT_SET: {
-            THROW_ERROR_EXCEPTION("None of operation id and operation alias is set in oneof OperationIdOrAlias proto");
+            operationIdOrAlias->Payload = {};
```
bystrov-serg, line 36: I don't like that `Payload` is `std::variant`, so we actually set here an empty OperationId. Maybe we should make Payload `std::optional`?

yt/yt/tests/integration/scheduler/test_scheduler_common.py
```diff
@@ -76,6 +80,14 @@ class TestSchedulerCommon(YTEnvSetup):
     }
     USE_PORTO = False
 
+    def setup_method(self, method):
```
bystrov-serg: We need the archive for only 2 tests in this suite, so i would put initialization in the tests

yt/yt/tests/integration/scheduler/test_scheduler_common.py
```diff
@@ -195,9 +207,12 @@ def test_fail_context(self):
             track=False,
             in_="//tmp/t1",
             out="//tmp/t2",
-            command='python3 -c "import os; os.read(0, 1);"',
+            command=with_breakpoint('BREAKPOINT; python3 -c "import os; os.read(0, 1);"'),
             spec={"mapper": {"input_format": "dsv", "check_input_fully_consumed": True}, "max_failed_job_count": 2},
         )
+        for job_id in wait_breakpoint():
+            get_job_fail_context(None, job_id)
```
bystrov-serg, line 212: Not sure if it works here, because no jobs have failed yet

bystrov-serg (changes requested):

yt/yt/client/scheduler/operation_id_or_alias-inl.h
faucct: An empty `OperationId` is already used as a missing value in the code (e.g. `scheduler_commands.h`). Changing type would require more refactoring. I would rather keep it like this, unless you insist.

yt/yt/tests/integration/scheduler/test_scheduler_common.py
faucct: At least it does not fail with job not found. Do you want me to delete it?

yt/yt/client/driver/scheduler_commands.h
faucct: Can you suggest a new one, please?

bystrov-serg: I actually understood that this class is used for commands where operation_id must be set (for example, list_jobs, get_operation). I suggest to make another class for cases when operation_id can be optional and use it there.
As for error message for a new class I would suggest a simple "Exactly one of \"operation_id\" and \"operation_alias\" or \"job_id\" should be set" and add new error attribute with job_id.

yt/yt/client/scheduler/operation_id_or_alias-inl.h
```diff
@@ -33,7 +33,8 @@ void FromProto(TOperationIdOrAlias* operationIdOrAlias, const TProtoClass& enclo
             break;
         }
         case TProtoClass::OperationIdOrAliasCase::OPERATION_ID_OR_ALIAS_NOT_SET: {
-            THROW_ERROR_EXCEPTION("None of operation id and operation alias is set in oneof OperationIdOrAlias proto");
```
bystrov-serg: We actually should not remove this throw here, because for commands, where operation id must be set, we would allow empty operation id and fail with strange errors later. I think it is better to check for operation id in commands where it _can be empty_ rather then here.

yt/yt/client/scheduler/operation_id_or_alias-inl.h
bystrov-serg: I am okay with using an empty OperationId as missing value, but I don't like using an empty `Payload` as missing value. Maybe it's just me, but for me it isn't obviously that empty `Payload` is an empty `OperationId`, it can be confusing for reader. But as I said in other comment, I don't like allowing empty operation id everywhere either.

yt/yt/tests/integration/scheduler/test_scheduler_common.py
bystrov-serg: Okay, but you do not need `with_breakpoint` and `release_breakpoint` for this. You can just use op.list_jobs() to get jobs. And add some comment, that you check get_job_fail_context with an empty operation_id.

bystrov-serg (changes requested):

yt/yt/tests/integration/scheduler/test_scheduler_common.py
faucct: But I want this to happen while the job is still running: the breakpoints are to stagger it. Should I still drop breakpoints here?

PR #1347, opened 2025-06-30
Allow to run jupyt operations on behalf of user

dmi-feo:
* Changelog entry
Type: feature
Component: strawberry

This allows to run jupyt operations in behalf of its user (speclet creator)


yt/chyt/controller/internal/strawberry/oplet.go
```diff
@@ -866,6 +868,21 @@ func (oplet *Oplet) getOpACL() (acl []yt.ACE) {
...
+			ImpersonationUser: oplet.persistentState.Creator,
+		},
+	)
+	if err != nil {
+		oplet.setError(err)
+		return nil, err
+	}
+	return client, nil
+}
+
```
buyval01, line 886: Let's move the creation of user client to the place where the oplet itself is created, i.e. into the Agent code

Because now we have an unused field :(
https://github.com/ytsaurus/ytsaurus/pull/1347/files#diff-91b1191d3c885f0366edff8026acb1f24bc859e48dfee0ff916be396a632baadR490

In addition, it seems that we used to "mistakenly" refer to the system client as a user client. Now we will have a more explicit division of usage, so we can properly initialize it there.

yt/chyt/controller/internal/strawberry/oplet.go
```diff
@@ -964,7 +981,18 @@ func (oplet *Oplet) restartOp(ctx context.Context, reason string) error {
 		}
 	}
 
-	opID, err := oplet.userClient.StartOperation(ctx, yt.OperationVanilla, spec, nil)
+	var ytClientForOpStart yt.Client
```
buyval01, line 991: Then you need to use this client on this line as well
https://github.com/ytsaurus/ytsaurus/pull/1347/files#diff-679aa48ee11a87a69d1a99845f917365595f596b636ed26083a53fadea3f4bdaR1008

yt/chyt/controller/internal/strawberry/controller.go
```diff
@@ -51,6 +51,9 @@ type Controller interface {
 	// Returns `nil` if scaling is not required.
 	// May be called concurrently since it is accessed from `runScaler`, not from `background` goroutine.
 	GetScalerTarget(ctx context.Context, opletInfo OpletInfoForScaler) (*ScalerTarget, error)
+
+	// RunAsUser returns true if the operation should run as a user.
+	RunAsUser() bool
```
buyval01: Let's return this flag from Prepare method

yt/chyt/controller/internal/strawberry/oplet.go
dmi-feo: Creating user client in agent code would mean that we need to parse speclet there (in order to obtain creator's login), which doesn't seem really subtle. Probably it can be moved to `registerOplet`, but still a new oplet will be created with empty user client in `registerNewOplet`

PR #1350, opened 2025-07-01
yt/core/crypto: handle several certificate sources

koct9i:
Allow specifying both environment variable and file as certificate source.

This allows to use single "//sys/@cluster_connection" and "//sys/clusters" bus
configs for server instances and jobs like chyt: servers will use certificates
from secret files, operations - from secure vault environment variables.


yalekseev:
Shipped!

Issue #1242, opened 2025-04-25
Handle YQL Queries in separate processes


staketd:
### Current State
YQL Agent handles queries in single process and it is not uncommon for queries to cause some segfault due to error in code. If a segfault happens in yql agent process, all queries running on that agent instance are failed. 

### Proposal
We are proposing to separate user queries from each other using different processes. This way if some query is causing segfault or other error, it will not affect other queries running on same yql agent instance.

We are proposing following:
1. Add new implementation of IYqlPlugin interface - TYqlProcessPlugin. This Implementation will create and manage processes of yql-plugin-binary. Key points:
   - On startup, it will create several standby processes (number can be configured) for future queries.
   - Each process is handling only one query
   - On query start, implementation finds process, that doesn't currently runs a query and proxies Run request to that process via RPC request
   - GetProgress, Abort are proxied to process, where query is running
   - GetUsedClusters could be proxied to any process, because it doesn't refer to any query. To maintain isolation, a separate process can be used.
   - On query finish, process is restarted to avoid situation where new query is affected by old query
   - If there is no free processes, an error is thrown
   - If OnDynamicConfigChanged is called, all free processes are restarted. Incoming requests are waiting for restart to finish
   - Metrics of total and free processes
   - Handling of logs should be similar to one in exec node jobs
2. Add new binary - yql-agent-binary, which will contain a unix socket RPC server, that is similar to IYqlPlugin interface. It Proxies calls to it to underlying TYqlPlugin (bridge implementation). 
3. Depending on configuration, TYqlAgent can use either bridge implementation or new process implementation

A brief schema:
![Image](https://github.com/user-attachments/assets/15819ba0-dbf9-490c-8f6b-11087a5107f2)

psushin:
Many kudos for a detailed issue description!

PR #1353, opened 2025-07-03
Handle YQL queries in separate processes

staketd:
* Changelog entry
Type: feature
Component: query-tracker 

Resolves issue #1242 

1) Added new implementation of IYqlPlugin
    * Creates and manages a process pool of fixed size
    * Each process runs a unix-socket rpc server, with interface similar to IYqlPlugin
    * This rpc server proxies calls to bridge implementation of IYqlPlugin
    * Incoming queries acquire one of processes from pool and uses it afterwards
    * When query is finished or aborted, process is restarted
    * Each process falls into one of configured slots
    * Each slot has its own folder with logs, unix-socket and tmp folder for user-attached files
    * On start pool is initialized
2) YqlAgent can use both bridge and process implementations depending on config
3) Added alternative program for yql-agent binary, that launches rpc-server with plugin


yt/yql/agent/bin/main.cpp
```diff
@@ -1,8 +1,12 @@
 #include <yt/yql/agent/program.h>
 
 #include <yt/cpp/mapreduce/interface/client.h>
+#include <yt/yql/plugin/process/program.h>
```
Kontakter, line 3: Move this include upper to `yt/yql` and separate by new line

yt/yql/agent/yql_agent.cpp
```diff
@@ -254,9 +258,9 @@ class TYqlAgent
     void Stop() override
     { }
 
-    NYTree::IMapNodePtr GetOrchidNode() const override
-    {
-        return GetEphemeralNodeFactory()->CreateMap();
+    virtual NYTree::IYPathServicePtr CreateOrchidService() const override {
```
Kontakter: We normally put `{` on new line for classes and functions.

There are a lot of such places in this PR, please, fix all of them.

yt/yql/plugin/process/config.h
```diff
@@ -0,0 +1,72 @@
...
+#include <yt/yt/core/logging/config.h>
+#include <yt/yt/library/server_program/config.h>
+#include <yt/yt/server/lib/misc/config.h>
+
+#include <yt/yql/plugin/plugin.h>
+
+namespace NYT::NYqlPlugin {
+namespace NProcess {
+
+struct TYqlProcessPluginConfig : public NYTree::TYsonStruct {
```
Kontakter: Inheritance specification we usually write on separate line:
```cpp
struct TYqlProcessPluginConfig
    : public NYTree::TYsonStruct
{
...
```

yt/yql/plugin/process/config.h
```diff
@@ -0,0 +1,72 @@
...
+
+    REGISTER_YSON_STRUCT(TYqlPluginProcessInternalConfig);
+
+    static void Register(TRegistrar registrar);
+};
+
+DEFINE_REFCOUNTED_TYPE(TYqlPluginProcessInternalConfig)
+
+} // namespace NProcess
+
```
Kontakter, line 31: We usually don't add empty lines in such cases

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+#include <yt/yt/core/misc/fs.h>
+#include <yt/yt/core/misc/ring_queue.h>
+#include <yt/yt/core/rpc/bus/channel.h>
+#include <yt/yt/core/rpc/client.h>
+#include <yt/yt/core/ytree/convert.h>
+
+#include <yt/yt/library/process/process.h>
+
+#include <yt/yt/ytlib/api/native/config.h>
+#include <yt/yt/ytlib/yql_plugin/yql_plugin_proxy.h>
```
Kontakter: We normally order includes from more locals to globals:
```
#include "plugin.h"
#include "private.h"

#include <yt/yt/ytlib/...>
...

#include <yt/yt/library/...>

#include <yt/yt/core/...>
...

#include <yt/cpp/mapreduce/common/helpers.h>

...
```

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+        TProcessBasePtr process,
+        TFuture<void> result,
+        TYqlPluginProxy proxy)
+        : SlotIndex(slotIndex)
+        , ConfigSnapshot(std::move(configSnapshot))
+        , Process(std::move(process))
+        , Result(std::move(result))
+        , PluginProxy(std::move(proxy))
+        , IsActive(false)
+    {
```
Kontakter: `{ }`

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+    TYqlPluginProcessInternalConfigPtr ConfigSnapshot;
+    TProcessBasePtr Process;
+    TFuture<void> Result;
+    TYqlPluginProxy PluginProxy;
+    std::atomic<bool> IsActive;
+};
+
+DEFINE_REFCOUNTED_TYPE(TYqlPluginRunningProcess)
+constexpr int MODE0711 = S_IRWXU | S_IXGRP | S_IXOTH;
+} // namespace
```
Kontakter: Add newline before closing this anonymous namespace

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+            queryFile->set_type(static_cast<TYqlQueryFile_EContentType>(file.Type));
+        }
+        runQueryReq->set_mode(executeMode);
+
+        auto response = WaitFor(runQueryReq->Invoke());
+        if (!response.IsOK()) {
+            YT_LOG_ERROR(response, "Failed to get query result from subprocess (QueryId: %v, SlotIndex %v)", queryId, pluginProcess->SlotIndex);
+            return TQueryResult{
+                .YsonError = ConvertToYsonString(
+                                 TError("Failed to get query result from subprocess")
```
Kontakter: Too large indentation, we normally use just one indent.

And fix other similar places, please.

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+    NProfiling::TGauge ActiveProcessesGauge_;
+    NProfiling::TGauge ProcessesLimitGauge_;
+
+    std::shared_ptr<IRetryPolicy<const std::exception&>> StartPluginRetryPolicy_ = IRetryPolicy<const std::exception&>::GetExponentialBackoffPolicy(
+        /*retryClassFunction*/ [](const std::exception&) {
+            return ERetryErrorClass::LongRetry;
+        },
+        /*minDelay*/ TDuration::Seconds(5),
+        /*minLongRetryDelay*/ TDuration::Seconds(5),
+        /* maxDelay */ TDuration::Seconds(30),
```
Kontakter: `/*maxDelay*/`

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+        RunningYqlQueries_[queryId] = acquiredSlot;
+        YqlQueriesBySlotIndex_[acquiredSlot] = queryId;
+
+        StandbyProcessesGauge_.Update(StandbyProcessesQueue_.size());
+        ActiveProcessesGauge_.Update(RunningYqlQueries_.size());
+
+        auto acquiredProcess = YqlProcesses_[acquiredSlot];
+
+        // We should check whether query was started in process. If it did not, then
+        // we need to restart process, because otherwise it will be stuck forever
```
Kontakter: Add `.` at the end of the sentense

staketd:
fixed styling

yt/yql/agent/bin/main.cpp
```diff
@@ -1,8 +1,14 @@
 #include <yt/yql/agent/program.h>
 
+#include <yt/yql/plugin/process/program.h>
+
 #include <yt/cpp/mapreduce/interface/client.h>
 
 int main(int argc, const char** argv)
 {
+    if (TStringBuf(argv[1]).StartsWith(NYT::NYqlPlugin::NProcess::YqlPluginSubcommandName)) {
```
psushin: Alternative will be to rely on an environment variable, to distinguish between YqlAgent and YqlPlugin.

Like here.
https://github.com/ytsaurus/ytsaurus/blob/main/yt/cpp/mapreduce/client/init.cpp#L325

What bothers me most here, is StartsWith. Why don't we check for equality?

yt/yql/agent/bootstrap.cpp
```diff
@@ -315,6 +316,11 @@ void TBootstrap::OnDynamicConfigChanged(
         ConvertToYsonString(newConfig, EYsonFormat::Text));
 }
 
+const NServer::TNativeServerBootstrapConfigPtr TBootstrap::GetNativeServerBootstrapConfig() const
```
psushin: using namespace NYT::NServer

yt/yql/agent/yql_agent.cpp
```diff
@@ -243,7 +245,9 @@ class TYqlAgent
             .LogBackend = NYT::NLogging::CreateArcadiaLogBackend(TLogger("YqlPlugin")),
             .YqlPluginSharedLibrary = Config_->YqlPluginSharedLibrary,
         };
-        YqlPlugin_ = NYqlPlugin::CreateYqlPlugin(std::move(options));
+        YqlPlugin_ = Config_->ProcessPluginConfig->Enabled 
+            ? NYqlPlugin::NProcess::CreateProcessYqlPlugin(bootstrap, std::move(options), Config_->ProcessPluginConfig, YqlAgentProfiler().WithPrefix("/yql_plugin"))
```
psushin: using namespace

Maybe use more specific profiler path, like ```/process_yql_plugin``` ?

yt/yql/agent/yql_agent.cpp
```diff
@@ -495,12 +500,20 @@ class TYqlAgent
         }
         // TODO(max42): original YSON tends to unnecessary pretty.
         *((&yqlResponse)->*mutableProtoFieldAccessor)() = *rawField;
-    };
+    }
+
+    void BuildOrchid(NYson::IYsonConsumer* consumer) const
+    {
+        BuildYsonFluently(consumer).BeginMap()
```
psushin, line 723: .BeginMap is usually placed on separate line with the corresponding indent

yt/yql/plugin/plugin.h
```diff
@@ -88,9 +89,10 @@ struct IYqlPlugin
     virtual void Start() = 0;
 
     virtual TClustersResult GetUsedClusters(
+        TQueryId queryId,
         TString queryText,
         NYson::TYsonString settings,
-        std::vector<TQueryFile> files) noexcept = 0;
+        std::vector<TQueryFile> files) = 0;
```
psushin, line 106: Is it ok that we dropped noexcept here?

yt/yql/plugin/process/config.cpp
```diff
@@ -0,0 +1,54 @@
+#include "config.h"
+
+namespace NYT::NYqlPlugin {
+namespace NProcess {
+
```
psushin, line 4: forgotten delimiter string

yt/yql/plugin/process/config.cpp
```diff
@@ -0,0 +1,54 @@
...
+void TYqlProcessPluginConfig::Register(TRegistrar registrar)
+{
+    registrar.Parameter("enabled", &TThis::Enabled)
+        .Default(false);
+
+    registrar.Parameter("slots_count", &TThis::SlotsCount)
```
psushin: slot_count

yt/yql/plugin/process/config.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+    registrar.Parameter("yql_plugin_shared_library", &TThis::YqlPluginSharedLibrary);
+}
+
+void TYqlPluginProcessInternalConfig::Register(TRegistrar registrar)
+{
+    registrar.Parameter("slot_index", &TThis::SlotIndex);
+    registrar.Parameter("plugin_options", &TThis::PluginOptions)
+        .DefaultNew();
+}
```
psushin, line 19: delimiter

yt/yql/plugin/process/config.h
```diff
@@ -0,0 +1,77 @@
+#pragma once
+
+#include "private.h"
+#include "public.h"
+
+#include <yt/yt/core/logging/config.h>
```
psushin: includes should be sorted from the most specific ones (yt/yt/server) to the most generic ones (yt/yt/core)

yt/yql/plugin/process/config.h
```diff
@@ -0,0 +1,77 @@
...
+#include <yt/yt/library/server_program/config.h>
+#include <yt/yt/server/lib/misc/config.h>
+
+#include <yt/yql/plugin/plugin.h>
+
+namespace NYT::NYqlPlugin {
+namespace NProcess {
```
psushin: delimiter

yt/yql/plugin/process/config.h
```diff
@@ -0,0 +1,77 @@
...
+
+struct TYqlProcessPluginConfig
+    : public NYTree::TYsonStruct
+{
+
```
psushin: redundant empty string

yt/yql/plugin/process/config.h
```diff
@@ -0,0 +1,77 @@
...
+    NLogging::TLogManagerConfigPtr LogManagerTemplate;
+
+    REGISTER_YSON_STRUCT(TYqlProcessPluginConfig);
+
+    static void Register(TRegistrar registrar);
+};
+
+DEFINE_REFCOUNTED_TYPE(TYqlProcessPluginConfig)
+
+////////////////////////////////////////////////////////////////////////////////
```
psushin: extra empty string required

yt/yql/plugin/process/config.h
```diff
@@ -0,0 +1,77 @@
...
+
+    TYqlProcessPluginOptionsPtr PluginOptions;
+
+    REGISTER_YSON_STRUCT(TYqlPluginProcessInternalConfig);
+
+    static void Register(TRegistrar registrar);
+};
+
+DEFINE_REFCOUNTED_TYPE(TYqlPluginProcessInternalConfig)
+
```
psushin: delimiter

yt/yql/plugin/process/config.h
```diff
@@ -0,0 +1,77 @@
...
+DEFINE_REFCOUNTED_TYPE(TYqlProcessPluginOptions)
+
+////////////////////////////////////////////////////////////////////////////////
+
+struct TYqlPluginProcessInternalConfig
+    : public NServer::TNativeServerBootstrapConfig
+    , public TServerProgramConfig
+{
+
```
psushin: redundant empty string

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
+#include "plugin.h"
+
+#include "private.h"
+
+#include <library/cpp/retry/retry.h>
```
psushin: library/cpp goes to the end of includes

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+#include <yt/yt/core/misc/ring_queue.h>
+#include <yt/yt/core/rpc/bus/channel.h>
+#include <yt/yt/core/rpc/client.h>
+#include <yt/yt/core/ytree/convert.h>
+
+#include <yt/cpp/mapreduce/common/helpers.h>
+
+namespace NYT::NYqlPlugin {
+namespace NProcess {
+
```
psushin, line 36: delimiter

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+// Implementation of IYqlPlugin that managees a pool of subprocesses for query
+// execution. Each subprocess handles exactly one query and it is killed after
+// query finishes its execution.
+class TYqlProcessPlugin
+    : public IYqlPlugin
+{
+public:
+    explicit TYqlProcessPlugin(
```
psushin: redundant explicit

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+};
+
+DEFINE_REFCOUNTED_TYPE(TYqlPluginRunningProcess)
+constexpr int MODE0711 = S_IRWXU | S_IXGRP | S_IXOTH;
+
+} // namespace
+
+////////////////////////////////////////////////////////////////////////////////
+
+// Implementation of IYqlPlugin that managees a pool of subprocesses for query
```
psushin: please use doxygen style comments like in other files

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+        auto guard = NThreading::ReaderGuard(ProcessesLock_);
+        for (int slotIndex = 0; slotIndex < Config_->SlotsCount; ++slotIndex) {
+            futures.push_back(
+                BIND(&TYqlProcessPlugin::StartPluginInProcess, this, YqlProcesses_[slotIndex])
+                    .AsyncVia(Invoker_)
+                    .Run());
+        }
+        guard.Release();
+
+        WaitFor(AllSucceeded(futures)).ThrowOnError();
```
psushin: ThrowOnError on the next line.

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+using namespace NConcurrency;
+using namespace NYson;
+
+using NYqlClient::NProto::TYqlQueryFile;
+using NYqlClient::NProto::TYqlQueryFile_EContentType;
+using NYqlClient::NProto::TYqlResponse;
+
+namespace {
+
+DECLARE_REFCOUNTED_STRUCT(TYqlPluginRunningProcess);
```
psushin: For non-public classes DECLARE is usually paired with DEFINE after the struct declaration

yt/yql/plugin/process/public.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include <library/cpp/yt/memory/ref_counted.h>
+
+namespace NYT::NYqlPlugin {
+namespace NProcess {
+
+extern const TString YqlAgentProgrammName;
```
Krisha11: Progra**mm**?

yt/yql/plugin/process/program.h
```diff
@@ -0,0 +1,97 @@
...
+
+        auto ControlQueue_ = New<NConcurrency::TActionQueue>("YqlPluginServiceControl");
+        auto ControlInvoker_ = ControlQueue_->GetInvoker();
+
+        YT_VERIFY(config->BusServer->UnixDomainSocketPath);
+
+        TYqlPluginOptions options{
+            .SingletonsConfig = config->PluginOptions->SingletonsConfig,
+            .GatewayConfig = config->PluginOptions->GatewayConfig,
+            .DqGatewayConfig = config->PluginOptions->DqGatewayConfig.value_or(NYT::NYson::TYsonString()),
```
Krisha11: NYT:: seems redundant, doesn't it?

yt/yql/plugin/process/program.h
```diff
@@ -0,0 +1,97 @@
+#pragma once
+
+#include "config.h"
+#include "plugin_service.h"
+
+#include <library/cpp/yt/logging/backends/arcadia/backend.h>
```
Krisha11: `<library/...` -> `<util/...>` should be at the end

`<yt/yql...>` on the contrary upper

yt/yql/plugin/dynamic/impl.cpp
```diff
@@ -123,6 +124,7 @@ TBridgeClustersResult* BridgeGetUsedClusters(
     }
 
     auto result = nativePlugin->GetUsedClusters(
+        NYT::TGuid::FromString(queryId),
```
Krisha11, line 127: extra NYT::

yt/yql/plugin/process/config.cpp
```diff
@@ -0,0 +1,54 @@
+#include "config.h"
+
+namespace NYT::NYqlPlugin {
+namespace NProcess {
+
+void TYqlProcessPluginConfig::Register(TRegistrar registrar)
+{
+    registrar.Parameter("enabled", &TThis::Enabled)
+        .Default(false);
```
Krisha11: Could you please add some tests to `yt/yql/tests/agent` with yqlProcess functionality enabled?

yt/yql/agent/config.cpp
```diff
@@ -10,6 +10,8 @@
 
 #include <yt/yt/core/bus/tcp/config.h>
 
+#include <yt/yql/plugin/process/config.h>
```
psushin, line 11: This shoud go above yt/yt/core

yt/yql/agent/config.h
```diff
@@ -18,6 +18,8 @@
 
 #include <yt/yt/library/server_program/config.h>
 
+#include <yt/yql/plugin/process/public.h>
```
psushin: This should go above yt/yt/library

yt/yql/agent/yql_agent.cpp
```diff
@@ -254,9 +258,10 @@ class TYqlAgent
     void Stop() override
     { }
 
-    NYTree::IMapNodePtr GetOrchidNode() const override
+    virtual NYTree::IYPathServicePtr CreateOrchidService() const override
```
psushin: Typically, we only write `virtual` in the interfaces.
Here `override` is enough.
And trailing space is forbidden by our linters.

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+        auto result = WaitFor(AllSet(futures));
+        if (!result.IsOK()) {
+            YT_LOG_ERROR(result, "Failed to update dynamic config");
+            return;
+        }
+
+        YT_LOG_INFO("Dynamic config updated");
+    }
+
+    virtual NYTree::IMapNodePtr GetOrchidNode() const override
```
psushin: redundat virtual

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+    {
+        auto guard = NThreading::ReaderGuard(ProcessesLock_);
+        auto yqlPluginNode = NYTree::GetEphemeralNodeFactory()->CreateMap();
```
psushin: Please, rewrite with BuildYsonNodeFluently().BeginMap()..

yt/yql/plugin/plugin.cpp
```diff
@@ -4,6 +4,11 @@
 
 namespace NYT::NYqlPlugin {
```
psushin, line 7: delimiter after namespace begin

yt/yql/plugin/plugin.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <yt/yql/plugin/bridge/interface.h>
```
psushin, line 5: empty line

yt/yql/plugin/plugin.h
```diff
@@ -1,6 +1,7 @@
...
+#include <yt/yt/core/ytree/ephemeral_node_factory.h>
```
psushin, line 6: ytree/public.h

yt/yql/plugin/plugin.h
```diff
@@ -99,13 +101,15 @@ struct IYqlPlugin
         TString queryText,
         NYson::TYsonString settings,
         std::vector<TQueryFile> files,
-        int executeMode) noexcept = 0;
+        int executeMode) = 0;
```
psushin, line 115: Shouldn't it be some kind of enum?

yt/yql/plugin/process/config.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+    registrar.Parameter("slots_count", &TThis::SlotsCount)
+        .Default(32);
+
+    registrar.Parameter("slots_root_path", &TThis::SlotsRootPath)
```
psushin: In other parts of the system we use term "Location" for some subsystems located on file system.
Like SlotLocation in exec nodes or ChunkLocation in data nodes.

Here I would suggest name like SlotLocationPath.

yt/yql/plugin/process/config.h
```diff
@@ -0,0 +1,77 @@
...
+#include <yt/yt/core/logging/config.h>
+#include <yt/yt/library/server_program/config.h>
+#include <yt/yt/server/lib/misc/config.h>
+
+#include <yt/yql/plugin/plugin.h>
+
+namespace NYT::NYqlPlugin {
+namespace NProcess {
+
+struct TYqlProcessPluginConfig
```
psushin: TProcessYqlPluginConfig

yt/yql/plugin/process/config.h
```diff
@@ -0,0 +1,77 @@
...
+
+    REGISTER_YSON_STRUCT(TYqlProcessPluginConfig);
+
+    static void Register(TRegistrar registrar);
+};
+
+DEFINE_REFCOUNTED_TYPE(TYqlProcessPluginConfig)
+
+////////////////////////////////////////////////////////////////////////////////
+struct TYqlProcessPluginOptions
```
psushin: TProcessYqlPluginOptions

yt/yql/plugin/process/config.h
```diff
@@ -0,0 +1,77 @@
...
+    REGISTER_YSON_STRUCT(TYqlProcessPluginOptions);
+
+    static void Register(TRegistrar registrar);
+};
+
+DEFINE_REFCOUNTED_TYPE(TYqlProcessPluginOptions)
+
+////////////////////////////////////////////////////////////////////////////////
+
+struct TYqlPluginProcessInternalConfig
```
psushin: TProcessYqlPluginInternalConfig

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+#include <yt/yt/core/rpc/bus/channel.h>
+#include <yt/yt/core/rpc/client.h>
+#include <yt/yt/core/ytree/convert.h>
+
+#include <yt/cpp/mapreduce/common/helpers.h>
+
+namespace NYT::NYqlPlugin {
+namespace NProcess {
+
+static constexpr auto& Logger = YqlProcessPluginLogger;
```
psushin: ProcessYqlPluginLogger

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+using namespace NYson;
+
+using NYqlClient::NProto::TYqlQueryFile;
+using NYqlClient::NProto::TYqlQueryFile_EContentType;
+using NYqlClient::NProto::TYqlResponse;
+
+namespace {
+
+DECLARE_REFCOUNTED_STRUCT(TYqlPluginRunningProcess);
+struct TYqlPluginRunningProcess
```
psushin: May be something like TYqlExecutorProcess?

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+constexpr int MODE0711 = S_IRWXU | S_IXGRP | S_IXOTH;
+
+} // namespace
+
+////////////////////////////////////////////////////////////////////////////////
+
+// Implementation of IYqlPlugin that managees a pool of subprocesses for query
+// execution. Each subprocess handles exactly one query and it is killed after
+// query finishes its execution.
+class TYqlProcessPlugin
```
psushin: Now we have
IYqlPlugin -> TBridgeYqlPluging -> CreateBridgeYqlPlugin, etc

Lets use the same consistent convention for other naming

IYqlPlugin -> TProcessYqlPlugin -> TProcessYqlPluginConfig, etc

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+    : public IYqlPlugin
+{
+public:
+    explicit TYqlProcessPlugin(
+        NYqlAgent::TBootstrap* bootstrap,
+        TYqlPluginOptions options,
+        TYqlProcessPluginConfigPtr config,
+        const NProfiling::TProfiler& profiler)
+        : Config_(std::move(config))
+        , Queue_(New<TActionQueue>("YqlProcessPlugin"))
```
psushin, line 74: ProcessYqlPlugin

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+
+    int SlotIndex;
+    TYqlPluginProcessInternalConfigPtr ConfigSnapshot;
+    TProcessBasePtr Process;
+    TFuture<void> Result;
+    TYqlPluginProxy PluginProxy;
+    std::atomic<bool> IsActive;
+};
+
+DEFINE_REFCOUNTED_TYPE(TYqlPluginRunningProcess)
```
psushin: empty line

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+    // in subsequent Run call. If Run call did not happen in one minute, the process will be reinitialized.
+    TClustersResult GetUsedClusters(
+        TQueryId queryId,
+        TString queryText,
+        NYson::TYsonString settings,
+        std::vector<TQueryFile> files) override
+    {
+        // Yql agent calls this method first when staring query, so we acquire process
+        // in this call and then use it in Run() as well.
+        TYqlPluginRunningProcessPtr acquiredProcess = AcquireSlotForQuery(queryId);
```
psushin: auto

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+
+        if (!acquiredProcess) {
```
psushin, line 108: That's kind of confusing that we acquire process in this quite innocent-looking method.
It seems to me that it would be much better to have an explicit StartQuery method, possibly returning some opaque handler. And in case of ProcessYqlPlugin this handler could keep a reference to an acquired process. 

And when the handler is destroyed - the process would be destroyed.

This would unfortunately require significant reworking of the interfaces, I would discuss it with @Krisha11

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+            return TClustersResult{
+                .YsonError = ConvertToYsonString(TError("No available slots to acquire")).ToString()
+            };
+        }
+
+        YT_LOG_INFO("Acquired slot for query (SlotIndex: %v, QueryId: %v)", acquiredProcess->SlotIndex, queryId);
+
+        auto getUsedClustersReq = acquiredProcess->PluginProxy.GetUsedClusters();
```
psushin: At least we could use more concise RPC interface between plugin and subprocess.

Approximate interface would be.
```
InitQuery(text, settings, files) -> void
GetUsedClusters() -> TClusters
RunQuery() -> ...
```

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+            auto queryFile = getUsedClustersReq->add_files();
+            queryFile->set_name(file.Name);
+            queryFile->set_content(file.Content);
+            queryFile->set_type(static_cast<TYqlQueryFile_EContentType>(file.Type));
+        }
+
+        auto response = WaitFor(getUsedClustersReq->Invoke());
+        if (!response.IsOK()) {
+            YT_LOG_ERROR(response, "Failed to get cluster result from subprocess (QueryId: %v, SlotIndex %v)", queryId, acquiredProcess->SlotIndex);
+            return TClustersResult{
```
psushin: This class is probably inherited from the times of BridgePlugin, but now seems weird.
More straiforward would be to use something like ```TErrorOr<TClusters>```
But this is probably the question of a separate refactoring.

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+                .YsonError = ConvertToYsonString<TError>(
+                    TError("Failed to get used clusters result from subprocess")
+                    << response
+                    << TErrorAttribute("slot_index", acquiredProcess->SlotIndex))
+                    .ToString()
+            };
+        }
+
+        auto responseValue = response.Value();
```
psushin: Please make proper ToProto / FromProto functions.

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+    TQueryResult Run(
+        TQueryId queryId,
+        TString user,
+        NYson::TYsonString credentials,
+        TString queryText,
+        NYson::TYsonString settings,
+        std::vector<TQueryFile> files,
+        int executeMode) override
+    {
+        auto pluginProcessOrError = GetYqlPluginByQueryId(queryId);
```
psushin, line 130: GetYql(Plugin|Executor)ProcessByQueryId

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+        if (!pluginProcessOrError.IsOK()) {
+            return TQueryResult{
+                .YsonError = ConvertToYsonString<TError>(pluginProcessOrError).ToString()
+            };
+        }
+        auto pluginProcess = pluginProcessOrError.Value();
+        YT_LOG_INFO("Running query in slot (SlotIndex: %v, QueryId: %v)", pluginProcess->SlotIndex, queryId);
+
+        // Mark query as active so it will not be restarted in CheckProcessIsActive.
+        pluginProcess->IsActive.store(true);
```
psushin: The implementation with handler would make all this manual bookkeeping redundant.

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+
+        // Handle query finish in the end.
+        auto finishQueryGuard = Finally(BIND(&TYqlProcessPlugin::OnQueryFinish, this, queryId, pluginProcess)
+                                            .Via(Invoker_));
```
psushin: smth is very wrong with indentation here)

yt/yt/ytlib/yql_plugin/yql_plugin_proxy.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
```
psushin, line 1: IMO this file should live somewhere in yt/yql

yt/yt/ytlib/yql_plugin/proto/yql_plugin.proto
```diff
@@ -0,0 +1,82 @@
+package NYT.NYqlPlugin.NProto;
```
psushin, line 1: IMO this file should live somewhere in yt/yql

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+        }
+        if (responseValue->has_error()) {
+            result.YsonError = responseValue->error();
+        }
+
+        return result;
+    }
+
+    // Gets an acquired in GetUsedClusters subprocess and routes Run call to it. Marks subprocess
+    // as active so it would not reinitialize. Long blocking call
```
psushin: dot at the end of comment

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+                                            .Via(Invoker_));
+
+        auto runQueryReq = pluginProcess->PluginProxy.RunQuery();
+        ToProto(runQueryReq->mutable_query_id(), queryId);
+        runQueryReq->set_user(user);
+        runQueryReq->set_credentials(credentials.ToString());
+        runQueryReq->set_query_text(queryText);
+        runQueryReq->set_settings(settings.ToString());
+        for (const auto& file : files) {
+            auto queryFile = runQueryReq->add_files();
```
psushin: Also looks like a good candidate for ToProto / FromProto

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+                    << response
+                    << TErrorAttribute("slot_index", pluginProcess->SlotIndex))
+                    .ToString()
+            };
+        }
+
+        return ToQueryResult(response.Value()->response());
+    }
+
+    virtual TQueryResult GetProgress(TQueryId queryId) override
```
psushin: redundant virtual

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+                    << response
+                    << TErrorAttribute("slot_index", pluginProcess->SlotIndex))
+                    .ToString()
+            };
+        }
+
+        return ToQueryResult(response.Value()->response());
+    }
+
+    virtual TAbortResult Abort(TQueryId queryId) override
```
psushin: redundant virtual

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+        auto pluginProcessOrError = GetYqlPluginByQueryId(queryId);
+        if (!pluginProcessOrError.IsOK()) {
+            return TAbortResult{
+                .YsonError = ConvertToYsonString<TError>(pluginProcessOrError).ToString()
+            };
+        }
+
+        auto pluginProcess = pluginProcessOrError.Value();
+        YT_LOG_INFO("Aborting query (SlotIndex: %v, QueryId: %v)", pluginProcess->SlotIndex, queryId);
+        auto abortQueryReq = pluginProcess->PluginProxy.AbortQuery();
```
psushin: Why do we need a separate rpc call for abort? Why don't we just kill a process?

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+        }
+
+        TAbortResult abortResult;
+        if (response.Value()->has_error()) {
+            abortResult.YsonError = response.Value()->error();
+        }
+        return abortResult;
+    }
+
+    virtual void OnDynamicConfigChanged(TYqlPluginDynamicConfig config) override
```
psushin: redundant virtual

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+            };
+        }
+
+        auto pluginProcess = pluginProcessOrError.Value();
+        YT_LOG_INFO("Aborting query (SlotIndex: %v, QueryId: %v)", pluginProcess->SlotIndex, queryId);
+        auto abortQueryReq = pluginProcess->PluginProxy.AbortQuery();
+        ToProto(abortQueryReq->mutable_query_id(), queryId);
+
+        auto response = WaitFor(abortQueryReq->Invoke());
+        if (!response.IsOK()) {
```
psushin: we could deduplicate some code extracting this error wrapping in some template function like

```return WrapPluginError<TAbortResult>(response, "Failed to abort a query")```

File: yt/yql/plugin/process/plugin.cpp [hunk 71]
```diff
@@ -0,0 +1,713 @@
...
+    virtual void OnDynamicConfigChanged(TYqlPluginDynamicConfig config) override
+    {
+        YT_LOG_INFO("Updating dynamic config");
+
+        auto guard = NThreading::WriterGuard(ProcessesLock_);
+        UpdateConfigTemplateOnConfigChanged(config);
+        std::vector<TFuture<void>> futures;
+        TString gatewaysConfig = config.GatewaysConfig.ToString();
+
+        for (int i = 0; i < Config_->SlotsCount; ++i) {
```
psushin: Looks a bit overengineered to me - why don't we just restart all the processes with a new config?

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+        yqlPluginNode->AddChild("total_processes", NYTree::BuildYsonNodeFluently().Value(Config_->SlotsCount));
+        return yqlPluginNode;
+    }
+
+private:
+    static TString SocketName;
+
+    TYqlProcessPluginConfigPtr Config_;
+
+    TActionQueuePtr Queue_;
```
psushin, line 250: Queue_ -> ActionQueue_

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+        // we need to restart process, because otherwise it will be stuck forever.
+        NYT::NConcurrency::TDelayedExecutor::Submit(
+            BIND(&TYqlProcessPlugin::CheckProcessIsActive, this, acquiredProcess, queryId),
+            Config_->CheckProcessActiveDelay,
+            Invoker_);
+
+        return acquiredProcess;
+    }
+
+    TErrorOr<TYqlPluginRunningProcessPtr> GetYqlPluginByQueryId(TQueryId queryId)
```
psushin: Synchronous methods usually return error via exception.

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+
+    TActionQueuePtr Queue_;
+    IInvokerPtr Invoker_;
+
+    YT_DECLARE_SPIN_LOCK(NThreading::TReaderWriterSpinLock, ProcessesLock_);
+    THashMap<TQueryId, int> RunningYqlQueries_;
+    THashMap<int, TQueryId> YqlQueriesBySlotIndex_;
+    std::vector<TYqlPluginRunningProcessPtr> YqlProcesses_;
+    std::deque<int> StandbyProcessesQueue_;
+    THashSet<int> StandByProcesses_;
```
psushin: StandbyProcesses

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+        const NProfiling::TProfiler& profiler)
+        : Config_(std::move(config))
+        , Queue_(New<TActionQueue>("YqlProcessPlugin"))
+        , Invoker_(Queue_->GetInvoker())
+        , YqlProcesses_(Config_->SlotsCount)
+        , ConfigTemplate_(BuildPluginConfigTemplate(Config_, std::move(options), bootstrap))
+        , StandbyProcessesGauge_(profiler.Gauge("/standby_processes"))
+        , ActiveProcessesGauge_(profiler.Gauge("/active_processes"))
+        , ProcessesLimitGauge_(profiler.Gauge("/processes_limit"))
+    {
```
psushin, line 80: Generally it can be better to decompose this class into three different pieces:
 - extract logic sending internal rpc request into TYqlPluginProcess class
 - make separate ProcessManager class that manages start / stop / acquisition of standby processes
 - current TProcessYqlAgent could be just a glue on top of these two

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+        ActiveProcessesGauge_.Update(RunningYqlQueries_.size());
+        StandbyProcessesGauge_.Update(StandbyProcessesQueue_.size());
+    }
+
+    void CheckProcessIsActive(TYqlPluginRunningProcessPtr process, TGuid queryId)
+    {
+        if (process->IsActive.load()) {
+            return;
+        }
+        YT_LOG_WARNING("Query did not start after acquiring process; Restarting process (SlotIndex: %v, QueryId: %v)", process->SlotIndex, queryId);
```
psushin: ```...; restarting ...```

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+
+        YT_LOG_INFO("Successfully started plugin in subprocess (SlotIndex: %v)", slotIndex);
+    }
+
+    void DoStartPluginInProcess(TYqlPluginRunningProcessPtr process)
+    {
+        YT_ASSERT_INVOKER_AFFINITY(Invoker_);
+
+        YT_LOG_DEBUG("Starting plugin in process (SlotIndex: %v)", process->SlotIndex);
+        Y_ENSURE(
```
psushin: YT_VERIFY

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+            NFS::Exists(GetUnixDomainSocketPath(process->SlotIndex)),
+            "Unix socket in slot must exist before calling Start method");
+        WaitFor(process->PluginProxy.Start()->Invoke()).ThrowOnError();
```
psushin: Why should we start it with explicit RPC call?

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,713 @@
...
+        result->PluginOptions->YTTokenPath = options.YTTokenPath;
+
+        return result;
+    }
+
+    TQueryResult ToQueryResult(const TYqlResponse& yqlResponse) const
+    {
+        TQueryResult result;
+
+        SetQueryResultField(result.YsonResult, yqlResponse, &TYqlResponse::has_result, &TYqlResponse::result);
```
psushin: IMO separate method here doesn't have much value

yt/yql/plugin/process/plugin_service.cpp
```diff
@@ -0,0 +1,186 @@
...
+        : TServiceBase(
+            std::move(controlInvoker),
+            TYqlPluginProxy::GetDescriptor(),
+            YqlPluginServiceLogger())
+        , YqlPlugin_(std::move(yqlPlugin))
+        , QueryActionQueue_(New<TActionQueue>("QueryRunner"))
+    {
+
+        RegisterMethod(RPC_SERVICE_METHOD_DESC(RunQuery)
+                         .SetCancelable(true)
```
psushin, line 34: too big indent

yt/yql/plugin/process/plugin_service.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+
+NRpc::IServicePtr CreateYqlPluginService(
+    IInvokerPtr controlInvoker,
+    std::unique_ptr<IYqlPlugin> yqlAgent)
+{
+    return New<TYqlPluginService>(std::move(controlInvoker), std::move(yqlAgent));
+}
+
```
psushin, line 189: delimiter

yt/yql/plugin/process/plugin_service.cpp
```diff
@@ -0,0 +1,186 @@
...
+          .GatewaysConfig = TYsonString(request->gateways_config())});
+
+        context->Reply();
+    }
+
+private:
+    const std::unique_ptr<IYqlPlugin> YqlPlugin_;
+    const TActionQueuePtr QueryActionQueue_;
+
+    std::vector<TQueryFile> ExtractFiles(size_t size, const google::protobuf::RepeatedPtrField<TYqlQueryFile>& protoFiles)
```
psushin, line 143: RepeatedPtrField has a size() method

yt/yql/plugin/process/plugin_service.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include <yt/yt/core/rpc/public.h>
+
+#include <yt/yql/plugin/plugin.h>
+
+namespace NYT::NYqlPlugin {
+namespace NProcess {
+
```
psushin, line 8: delimiters

yt/yql/plugin/process/private.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include <yt/yt/core/logging/log.h>
+
+namespace NYT::NYqlPlugin {
+namespace NProcess {
+
```
psushin, line 14: delimiters

yt/yql/plugin/process/program.h
```diff
@@ -0,0 +1,97 @@
...
+#include <yt/yt/library/program/program.h>
+#include <yt/yt/library/program/program_config_mixin.h>
+#include <yt/yt/library/program/program_pdeathsig_mixin.h>
+#include <yt/yt/library/program/program_setsid_mixin.h>
+
+#include <yt/yql/plugin/bridge/plugin.h>
+
+namespace NYT::NYqlPlugin {
+namespace NProcess {
+
```
psushin, line 30: delimiters

yt/yql/plugin/process/program.h
```diff
@@ -0,0 +1,97 @@
...
+    , public TProgramSetsidMixin
+    , public TProgramConfigMixin<TYqlPluginProcessInternalConfig>
+{
+public:
+    TYqlPluginProgram()
+        : TProgram()
+        , TProgramPdeathsigMixin(Opts_)
+        , TProgramSetsidMixin(Opts_)
+        , TProgramConfigMixin(Opts_)
+    {
```
psushin: braces on the same line

yt/yql/plugin/process/program.h
```diff
@@ -0,0 +1,97 @@
...
+
+        RpcServer->RegisterService(YqlPluginService);
+
+        RpcServer->Configure(config->RpcServer);
+        RpcServer->Start();
+
+        Sleep(TDuration::Max());
+    }
+};
+
```
psushin, line 98: delimiter

yt/yql/plugin/process/program.h
```diff
@@ -0,0 +1,97 @@
...
+        MlockFileMappings();
+        RunMixinCallbacks();
+
+        auto config = GetConfig();
+
+        ConfigureSingletons(config);
+
+        NProfiling::EnablePerfEventCounterProfiling();
+
+        auto ControlQueue_ = New<NConcurrency::TActionQueue>("YqlPluginServiceControl");
```
psushin: all local variables start with lowercase, no underscore at the end

yt/yql/plugin/process/public.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include <library/cpp/yt/memory/ref_counted.h>
+
+namespace NYT::NYqlPlugin {
+namespace NProcess {
+
```
psushin, line 6: delimiters

psushin (changes requested):
There are quite a few change requests, both cosmetics and substantial.

Probably it would make sense to arrange a zoom call.

yt/yql/plugin/process/plugin.cpp
staketd: I didn't look up into native plugin implementation, so I assumed graceful query abortion is probably needed

yt/yql/plugin/process/plugin.cpp
staketd: You mean call Start in subprocess on startup? I tried to somewhat follow existing logic of yql agent, that is not to call Start before agent did.
If it is not required, then that way will be easer

yt/yql/plugin/plugin.h
staketd: noexcept was dropped because methods in new implementation are no longer noexcept. Exception may arise on Abort, because on Abort agent cancels fiber, causing WaitFor(...->Invoke()) to throw a TFiberCancelationException

yt/yql/plugin/plugin.h
psushin: Probably not)

yt/yql/plugin/process/plugin.cpp
psushin: Yeah, IMO we don't have to transfer all the existing logic method-by-method via RPC into subprocesses. 

Current interfaces are very much influenced by somewhat legacy BridgePlugin implementation, and its pure C dynamic library interface. Although reworking IYqlPlugin is probably out of scope of this PR, we can at least make internal RPC interface more lean and straitforward.

Like:
 - start Yql in subprocess right from the beginning, without special call; extra call can be necessary to ensure that Yql has finished initialization, loaded all the UDF, etc;
 - explicit InitQuery method - the only method that requires query text / files / options etc
 - get rid of DynamicConfigUpdate - just restart all the standby processes
 - get rid of Abort method - just kill the process (this is a controvercial one, let's see what @Krisha11 thinks about it)

staketd: If I understood correctly, we have following RPC interface:
1) InitQuery(queryId, text, files, ...) - initializes query inside subprocess itself, saved arguments then used in other calls. The only thing is that in GetUsedClusters there are not all required params for Run, though it can be fixed by just extending interface. It also returns used clusters
2) Run()
3) GetProgress()
4) Abort() (?)

About Start() - calling it outside of process was a good marker that process is ready and can accept requests. Now different method will be needed - maybe subprocess will make some rpc call back to main process?

Also,
> just restart all the standby processes

Currently if there are no standby process for incoming query, an error is thrown. If we restart all such processes we will temporarly not be able to serve requests. And because a dynamic config is changed, almost all agents can be affected.

Should we wait for standby processes instead of throwing an error instead?

Krisha11: StartQuery currently runs DQ. This should only be done once for yqla, not per process.

yt/yql/plugin/process/plugin.cpp
Krisha11: We need graceful shutdown to stop yt transactions and dq execution immediately

psushin: Ok, seems legit, we need graceful abortion.

yt/yql/plugin/process/plugin.cpp
psushin: > Should we wait for standby processes instead of throwing an error instead?

I think that's a good idea.

yt/yql/plugin/process/plugin.cpp
psushin: We will do the refactoring after this PR is merged.

psushin:
I wanted to run internal tests while I'm reading, but unfortunately we need to resolve conflicts first.

yt/python/yt/environment/components/yql_agent/__init__.py
```diff
@@ -103,7 +110,7 @@ def is_finished():
     def _get_artifact_path(self, subpath=""):
         return os.path.join(self.artifacts_path, subpath)
 
-    def get_default_config(self):
+    def get_default_config(self, instance_index: int):
```
psushin, line 113: looks like we don't use python typing in this file

yt/python/yt/environment/components/yql_agent/__init__.py
```diff
@@ -147,12 +174,15 @@ def get_default_config(self):
...
+        if process_plugin_config:
+            config["yql_agent"]["process_plugin_config"] = process_plugin_config
+
         return config
 
     def wait_for_readiness(self, address):
         wait(lambda: self.client.get(f"//sys/yql_agent/instances/{address}/orchid/service/version"),
              ignore_exceptions=True,
-             timeout=60)
+             timeout=600)
```
psushin, line 185: may be two minutes will be enough?

yt/yql/agent/bootstrap.cpp
```diff
@@ -315,6 +317,11 @@ void TBootstrap::OnDynamicConfigChanged(
         ConvertToYsonString(newConfig, EYsonFormat::Text));
 }
 
+const NApi::NNative::TConnectionCompoundConfigPtr TBootstrap::GetClusterConnectionConfig() const
```
psushin, line 320: NApi::NNative should probably go to "using namespace" section

yt/yql/agent/ya.make
```diff
@@ -46,6 +46,8 @@ PEERDIR(
 
     yt/yql/plugin
 
+    yt/yt/library/query/engine
```
psushin, line 49: what are these new dependencies used for?

yt/yql/plugin/config.cpp
```diff
@@ -0,0 +1,435 @@
...
+        YT_VERIFY(ytflowGatewayConfig->AddChild("default_settings", std::move(ytflowGatewaySettings)));
+
+        auto icSettingsConfig = config->DQManagerConfig->ICSettings->AsMap();
+        auto closeOnIdleMs = icSettingsConfig->FindChild("close_on_idle_ms");
+        if (!closeOnIdleMs) {
+            icSettingsConfig->AddChild("close_on_idle_ms", BuildYsonNodeFluently().Value(0));
+        }
+    });
+}
+
```
psushin, line 434: last delimiter is missing

yt/yql/plugin/config.h
```diff
@@ -0,0 +1,187 @@
...
+    REGISTER_YSON_STRUCT(TYqlPluginConfig);
+
+    static void Register(TRegistrar registrar);
+};
+
+DEFINE_REFCOUNTED_TYPE(TYqlPluginConfig)
+
+////////////////////////////////////////////////////////////////////////////////
+
+}
```
psushin, line 187: // namespace NYT::...

yt/yql/plugin/native/plugin.cpp
```diff
@@ -856,6 +858,8 @@ class TYqlPlugin
     THashMap<TQueryId, TActiveQuery> ActiveQueriesProgress_;
     TUserDataTable UserDataTable_;
 
+    bool StartDqManager_;
```
psushin, line 873: const, and should probably go to the top of parameter list

yt/yql/plugin/process/config.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include <yt/yql/plugin/plugin.h>
+#include <yt/yql/plugin/public.h>
+
+#include <yt/yt/server/lib/misc/config.h>
+
+#include <yt/yt/library/server_program/config.h>
+
+namespace NYT::NYqlPlugin::NProcess {
+
```
psushin, line 13: delimiter

yt/yql/plugin/process/config.h
```diff
@@ -0,0 +1,32 @@
...
+    std::optional<NYson::TYsonString> DynamicGatewaysConfig;
+
+    REGISTER_YSON_STRUCT(TProcessYqlPluginInternalConfig);
+
+    static void Register(TRegistrar registrar);
+};
+
+DEFINE_REFCOUNTED_TYPE(TProcessYqlPluginInternalConfig)
+
```
psushin, line 31: delimiter

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,525 @@
...
+        return result;
+    }
+
+    TQueryResult GetProgress(TQueryId queryId) override
+    {
+        auto pluginProcessOrError = GetYqlPluginByQueryId(queryId);
+        if (!pluginProcessOrError.IsOK()) {
+            return TQueryResult{
+                .YsonError = ConvertToYsonString<TError>(pluginProcessOrError).ToString()
+                };
```
psushin, line 164: wrong indent

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,525 @@
...
+        ++DynamicConfigVersion_;
+
+        if (DqControllerYqlPlugin_) {
+            DqControllerYqlPlugin_->OnDynamicConfigChanged(config);
+        }
+
+        while (!StandbyProcessesQueue_.Empty()) {
+            auto process = *StandbyProcessesQueue_.Pop();
+            YT_LOG_DEBUG("Stopping process for dynamic config update (SlotIndex: %v)", process->SlotIndex());
+            process->Stop();
```
psushin, line 224: I would just empty the queue while holding the guard, and the processes could be stopped after the lock is released.

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,525 @@
...
+
+    YT_DECLARE_SPIN_LOCK(NThreading::TReaderWriterSpinLock, ProcessesLock_);
+    THashMap<TQueryId, TYqlExecutorProcessPtr> RunningYqlQueries_;
+    ::NThreading::TBlockingQueue<TYqlExecutorProcessPtr> StandbyProcessesQueue_;
+
+    NProfiling::TGauge StandbyProcessesGauge_;
+    NProfiling::TGauge ActiveProcessesGauge_;
+    NProfiling::TGauge ProcessesLimitGauge_;
+
+private:
```
psushin, line 263: redundant

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,525 @@
...
+    {
+        auto guard = NThreading::WriterGuard(ProcessesLock_);
+        if (queryId) {
+            RunningYqlQueries_.erase(*queryId);
+        }
+        ActiveProcessesGauge_.Update(RunningYqlQueries_.size());
+        StandbyProcessesGauge_.Update(StandbyProcessesQueue_.Size());
+    }
+
+    void CheckProcessIsActive(TYqlExecutorProcessPtr process, TGuid queryId)
```
psushin, line 312: TQueryId

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,525 @@
...
+        } catch (std::exception e) {
+            YT_LOG_ERROR(e, "Failed to initialize");
+        }
+    }
+
+    TFuture<void> StartPluginProcess(int slotIndex)
+    {
+        YT_LOG_INFO("Starting yql plugin process (SlotIndex: %v)", slotIndex);
+
+        return BIND_NO_PROPAGATE(&TProcessYqlPlugin::StartProcess, this, slotIndex)
```
psushin, line 389: These raw "this" (here and below) look dangerous.

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,525 @@
...
+        TString unixDomainSocketPath = GetUnixDomainSocketPath(slotIndex);
+
+        if (NFS::Exists(unixDomainSocketPath)) {
+            NFS::Remove(unixDomainSocketPath);
+        }
+
+        auto serverConfig = NBus::TBusServerConfig::CreateUds(unixDomainSocketPath);
+        auto clientConfig = NBus::TBusClientConfig::CreateUds(unixDomainSocketPath);
+
+        TProcessYqlPluginInternalConfigPtr config = BuildProcessConfig(slotIndex, serverConfig);
```
psushin, line 429: auto config

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,525 @@
...
+
+        auto client = NBus::CreateBusClient(clientConfig);
+
+        TProcessBasePtr process = New<TSimpleProcess>(YqlAgentProgramName);
```
psushin, line 433: auto process

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,525 @@
...
+            std::move(proxy),
+            unixDomainSocketPath,
+            process,
+            process->Spawn(),
+            Config_->ProcessPluginConfig->RunRequestTimeout);
+    }
+
+    TProcessYqlPluginInternalConfigPtr BuildProcessConfig(int slotIndex, NBus::TBusServerConfigPtr serverConfig)
+    {
+        TProcessYqlPluginInternalConfigPtr config = NYTree::CloneYsonStruct(ConfigTemplate_);
```
psushin, line 458: auto config

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,525 @@
...
+        YT_LOG_DEBUG("Yql plugin config successfully written (SlotIndex: %v, ConfigPath: %v)", config->SlotIndex, configPath);
+    }
+
+    static TProcessYqlPluginInternalConfigPtr BuildPluginConfigTemplate(
+        TYqlPluginConfigPtr config,
+        TSingletonsConfigPtr singletonsConfig,
+        TConnectionCompoundConfigPtr clusterConnectionConfig,
+        TString maxSupportedYqlVersion)
+    {
+        TProcessYqlPluginInternalConfigPtr result = New<TProcessYqlPluginInternalConfig>();
```
psushin, line 507: auto

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,525 @@
...
+
+        result->PluginConfig = config;
+        result->MaxSupportedYqlVersion = maxSupportedYqlVersion;
+
+        return result;
+    }
+
+    void InitializeDqControllerYqlPlugin(TSingletonsConfigPtr singletonsConfig, std::string maxSupportedYqlVersion)
+    {
+        TYqlPluginOptions options = ConvertToOptions(
```
psushin, line 524: auto options

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,525 @@
...
+            NYT::NLogging::CreateArcadiaLogBackend(NLogging::TLogger("YqlPlugin")),
+            maxSupportedYqlVersion,
+            true);
+        DqControllerYqlPlugin_ = CreateBridgeYqlPlugin(std::move(options));
+    }
+};
+
+////////////////////////////////////////////////////////////////////////////////
+
+TString TProcessYqlPlugin::SocketName = "yql-plugin.sock";
```
psushin, line 536: May be make it just a regular static const variable in the beginning of the file?

yt/yql/plugin/process/plugin.cpp
```diff
@@ -0,0 +1,525 @@
...
+std::unique_ptr<IYqlPlugin> CreateProcessYqlPlugin(
+    TYqlPluginConfigPtr pluginConfig,
+    TSingletonsConfigPtr singletonsConfig,
+    TConnectionCompoundConfigPtr clusterConnectionConfig,
+    TString maxSupportedYqlVersion,
+    const NProfiling::TProfiler& profiler)
+{
+    return std::make_unique<TProcessYqlPlugin>(std::move(pluginConfig), singletonsConfig, clusterConnectionConfig, maxSupportedYqlVersion, profiler);
+}
+
```
psushin, line 549: delimiter

yt/yql/plugin/process/plugin_service.cpp
```diff
@@ -0,0 +1,187 @@
...
+                         .SetInvoker(QueryActionQueue_->GetInvoker()));
+        RegisterMethod(RPC_SERVICE_METHOD_DESC(AbortQuery));
+        RegisterMethod(RPC_SERVICE_METHOD_DESC(GetQueryProgress));
+        RegisterMethod(RPC_SERVICE_METHOD_DESC(Start));
+        RegisterMethod(RPC_SERVICE_METHOD_DESC(GetUsedClusters));
+        RegisterMethod(RPC_SERVICE_METHOD_DESC(OnDynamicConfigChanged));
+    }
+
+    DECLARE_RPC_SERVICE_METHOD(NYqlPlugin::NProto, Start)
+    {
```
psushin: SetRequestInfo in each rpc call

yt/yql/plugin/process/process.cpp
```diff
@@ -0,0 +1,271 @@
...
+    TFuture<void> processFinishFuture,
+    TDuration runRequestTimeout)
+    : SlotIndex_(slotIndex)
+    , DynamicConfigVersion_(dynamicConfigVersion)
+    , PluginProxy_(std::move(pluginProxy))
+    , UnixSocketPath_(unixSocketPath)
+    , YqlPluginProcess_(yqlPluginProcess)
+    , ProcessFinishFuture_(std::move(processFinishFuture))
+    , RunRequestTimeout_(runRequestTimeout)
+    { }
```
psushin, line 68: Braces should be without indent

yt/yql/plugin/process/process.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+
+TClustersResult TYqlExecutorProcess::GetUsedClusters(
+        TQueryId queryId,
```
psushin, line 73: too large indent

yt/yql/plugin/process/process.cpp
```diff
@@ -0,0 +1,271 @@
...
+    runQueryReq->SetTimeout(RunRequestTimeout_);
+
+    ToProto(runQueryReq->mutable_query_id(), queryId);
+    runQueryReq->set_user(user);
+    runQueryReq->set_credentials(credentials.ToString());
+    runQueryReq->set_query_text(queryText);
+    runQueryReq->set_settings(settings.ToString());
+
+    for (const auto& file : files) {
+      auto queryFile = runQueryReq->add_files();
```
psushin, line 135: too small indent

yt/yql/plugin/process/process.cpp
```diff
@@ -0,0 +1,271 @@
...
+    auto response = WaitFor(runQueryReq->Invoke());
+    if (!response.IsOK()) {
+        YT_LOG_ERROR(response, "Failed to run query in subprocess (QueryId: %v, SlotIndex %v)", queryId, SlotIndex_);
+        return ToErrorResponse<TQueryResult>("Failed to run query in subprocess", response);
+    }
+
+    return ToQueryResult(response.Value()->response());
+}
+
+////////////////////////////////////////////////////////////////////////////////
```
psushin, line 152: well, here some delimiters (between the methods of the same class) are redundant in my opinion :)

yt/yql/plugin/process/process.cpp
```diff
@@ -0,0 +1,271 @@
...
+        YT_LOG_ERROR("Failed to get query progress from subprocess (QueryId: %v, SlotIndex %v)", queryId, SlotIndex_);
+        return ToErrorResponse<TQueryResult>("Failed to get query result from subprocess", response);
+    }
+
+    return ToQueryResult(response.Value()->response());
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+TAbortResult TYqlExecutorProcess::Abort(TQueryId queryId) {
```
psushin, line 172: brace on the next line

yt/yql/plugin/process/process.cpp
```diff
@@ -0,0 +1,271 @@
...
+    if (response.Value()->has_error()) {
+        abortResult.YsonError = response.Value()->error();
+    }
+    return abortResult;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+template<typename T, typename R>
+T TYqlExecutorProcess::ToErrorResponse(const TFormatString<>& errorMessage, const TErrorOr<R>& response) const {
```
psushin, line 221: brace on the next line

yt/yql/plugin/process/process.cpp
```diff
@@ -0,0 +1,271 @@
...
+        << TErrorAttribute("slot_index", SlotIndex_);
+
+    return T{
+        .YsonError = ConvertToYsonString<TError>(error).ToString()
+    };
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+int TYqlExecutorProcess::SlotIndex() const {
```
psushin, line 233: 1. Opening brace on the next line (here and later)
2. You could use macros like DECLARE/DEFINE_BYVAL_RO_PROPERTY

yt/yql/plugin/process/process.cpp
```diff
@@ -0,0 +1,271 @@
...
+    return SlotIndex_;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+int TYqlExecutorProcess::DynamicConfigVersion() const {
```
psushin, line 239: Opening brace here and in the next couple of methods

yt/yql/plugin/process/process.h
```diff
@@ -0,0 +1,98 @@
...
+    std::shared_ptr<IRetryPolicy<const std::exception&>> StartPluginRetryPolicy_ = IRetryPolicy<const std::exception&>::GetExponentialBackoffPolicy(
+        /*retryClassFunction*/ [](const std::exception&) {
+            return ERetryErrorClass::LongRetry;
+        },
+        /*minDelay*/ TDuration::Seconds(5),
+        /*minLongRetryDelay*/ TDuration::Seconds(5),
+        /*maxDelay*/ TDuration::Seconds(30),
+        /*maxRetries*/ 5);
+
+private:
```
psushin, line 94: we don't write "private" twice

yt/yql/plugin/public.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#include <library/cpp/yt/memory/ref_counted.h>
+
+namespace NYT::NYqlPlugin {
+
```
psushin, line 6: delimiters

yt/yt/ytlib/yql_plugin/yql_plugin_proxy.h
```diff
@@ -0,0 +1,28 @@
...
+namespace NYT::NYqlPlugin {
+
+////////////////////////////////////////////////////////////////////////////////
+
+class TYqlPluginProxy
+    : public NRpc::TProxyBase
+{
+public:
+    DEFINE_RPC_PROXY(TYqlPluginProxy, TYqlPluginService,
+                     .SetProtocolVersion(0));
```
psushin, line 16: wrong indent

psushin:
I'm ok with overall structure and implementation, there are just some cosmetic remarks.
We should resolve the conflicts and run internal tests - if they are fine, we can merge the PR and then do the refactoring and cosmetics.

Krisha11 (approved):

PR #1354, opened 2025-07-04
Use text representation of decimals.

Tony-Romanov:
Type: fix
Component: query-tracker


Krisha11:
Should not be merged. At the moment this code leads to problems in other parts of YT.

PR #1355, opened 2025-07-04
Update go sdk: add secrets.

Tony-Romanov:
Type: feature
Component: go-sdk


yt/go/yt/interface.go
```diff
@@ -1569,6 +1569,13 @@ type QueryTrackerOptions struct {
 	Stage *string `http:"stage,omitnil"`
 }
 
+type Secret struct {
```
tinarsky, line 1572: Let's add the field names in the "yson" tag for each field. Otherwise, the field names will be used as keys.

tinarsky (changes requested):

PR #1358, opened 2025-07-04
mTLS: add public rpc server to rpc proxies

koct9i:
- **yt/rpc-proxy: add dedicated bus server for public use**
- **yt/driver: add option for rpc-proxy default address type**
- **yt/environment: separate internal and public CA**
- **yt/environment: prefer https for rpc proxy discovery**
- **yt/test: use rpc proxy public rpc port for tls setup**

---

* Changelog entry
Type: feature
Component: misc-server

Add public RPC server for RPC proxies with own TLS certificate.


yalekseev:
Shipped!

yalekseev:
Shipped!

Issue #1359, opened 2025-07-05
Go HTTP Client goroutines leak on Go >= 1.23


mtvch:
After upgrading to `Go 1.23` from `Go 1.22` I noticed increase in number of goroutines caused by YT client. In my case it was around 100k goroutines that hanged in `net.(*sysDialer).dialParallel -> net.(*netFD).connect.func2`.

I think it's caused by this change in `net/http`: https://github.com/golang/go/commit/334ce510046ad30b1be466634cf313aad3040892

Configuring `MaxConnsPerHost` to small number (e.g. 10) helped to prevent the leak.

https://github.com/ytsaurus/ytsaurus/blob/9ad3055cd6f8a599ac433b42e4de676916baad02/yt/go/yt/internal/httpclient/client.go#L743

However, as for now Go client does not provide a way to configure this value.

I suggest modifying client constructor to allow user pass http client and net dialer configs. I can create a PR for that.

PR #1360, opened 2025-07-05
Go SDK: Pass custom http.Client through config

mtvch:
Feature for https://github.com/ytsaurus/ytsaurus/issues/1359

* Changelog entry
Type: feature
Component: go-sdk

Allow user to pass custom http.Client through yt.Config

### DIFF
--- a/yt/go/yt/internal/rpcclient/client.go
+++ b/yt/go/yt/internal/rpcclient/client.go
@@ -93,7 +93,7 @@ func NewClient(conf *yt.Config) (*client, error) {
 
 	transport, ok := c.httpClient.Transport.(*http.Transport)
 	if !ok {
-		return nil, errors.New("http.Transport is not supported")
+		return nil, xerrors.Errorf("expected *http.Transport, got %T: rpc client does not support other transports", c.httpClient.Transport)
 	}
 
 	if conf.Credentials != nil {


tinarsky (changes requested):
I suggest setting the value `MaxConnsPerHost`=`MaxIdleConnsPerHost`=100 by default in the library.

If you want to use an HTTP client with your own settings, you can pass `RoundTripper` (like `http.Transport`) in the context — https://github.com/ytsaurus/ytsaurus/blob/main/yt/go/yt/internal/httpclient/testing.go#L10

koct9i (approved):

mtvch:
> I suggest setting the value `MaxConnsPerHost`=`MaxIdleConnsPerHost`=100 by default in the library.
> 
> If you want to use an HTTP client with your own settings, you can pass `RoundTripper` (like `http.Transport`) in the context — https://github.com/ytsaurus/ytsaurus/blob/main/yt/go/yt/internal/httpclient/testing.go#L10

Sorry for the late reply! I changed default value to 100 as you suggested.

Regarding passing http.Transport via context - It seems to me that WithRoundTripper was meant only for testing (as it's defined in testing.go file). I think that config way is more intuitive and that's where users are going to look if they want to modify http client configuration. In many cases it makes sense to override default http config once where it's built in NewHTTPClient instead of modifying context each time you call a client function.

yt/go/yt/internal/rpcclient/client.go
```diff
@@ -77,6 +69,33 @@ func NewClient(conf *yt.Config) (*client, error) {
...
+	}
+
+	c.httpClient, err = BuildHTTPClient(conf)
+	if err != nil {
+		return nil, err
+	}
+
+	transport, ok := c.httpClient.Transport.(*http.Transport)
+	if !ok {
+		return nil, errors.New("http.Transport is not supported")
```
tinarsky: It looks like it should be a message with a type that is not supported?

tinarsky (changes requested):

yt/go/yt/internal/rpcclient/client.go
mtvch: Sure, fixed error message

tinarsky (approved):

PR #1363, opened 2025-07-07
Add internal_registry_regex to controller agents config

Krock21:
Sometimes, `internal_registry_alternative_addresses` are not enough, and a custom regex is required


Kontakter:
Shipped!

PR #1366, opened 2025-07-08
Use Poco::XML library in S3 client

pavel-bash:
# Description

This PR rewrites the internals of S3 client to use the `Poco::XML` library instead of the `libxml` to avoid having `libiconv` as a dependency to this module of the system.


pavel-bash:
Ready to be reviewed by @achulkov2.

yt/yt/library/s3/client.h
```diff
@@ -7,7 +7,16 @@
 
 #include <yt/yt/core/actions/future.h>
 
-#include <library/cpp/xml/document/xml-document.h>
+// Forward declaration for Poco::XML::Node.
+namespace Poco::XML {
+
+////////////////////////////////////////////////////////////////////////////////
```
achulkov2: In this particular case I would not use "otbivkas", because it is more annoying than not. Forward declaration because header is large?

yt/yt/library/s3/client.cpp
```diff
@@ -6,41 +6,81 @@
...
+#include <contrib/libs/poco/XML/include/Poco/DOM/DOMParser.h>
+#include <contrib/libs/poco/XML/include/Poco/DOM/Document.h>
+#include <contrib/libs/poco/XML/include/Poco/DOM/Node.h>
+
 namespace NYT::NS3 {
 
 using namespace NConcurrency;
 using namespace NNet;
 
+namespace NPocoXml = Poco::XML;
```
achulkov2: Why this variable? To unify namespace name formats? I think we tend to use the original namespace names for contribs.

yt/yt/library/s3/client.cpp
```diff
@@ -6,41 +6,81 @@
...
+        if (child->nodeType() == NPocoXml::Node::ELEMENT_NODE) {
+            return child;
+        }
+    }
+    THROW_ERROR_EXCEPTION("The document does not have a non-comment root node");
+}
+
+//! Returns a pair <ParsedDocument, RootNode>. We need both of them as the first will manage the lifetime
+//! of the parsed document, while the second one is usually what we actually want to access.
+std::pair<TPocoXmlDocumentPtr, NPocoXml::Node*> ParseXmlDocument(TSharedRef responseBody)
```
achulkov2: Let's make this a named struct. Or even a small wrapper that stores both and overrides operator* and -> to return node stuff. 

Where is an aliasing constructor when you need one :(

yt/yt/library/s3/client.cpp
```diff
@@ -82,17 +118,15 @@ void TListObjectsRequest::Serialize(THttpRequest* request) const
...
+    auto [document, rootNode] = ParseXmlDocument(response->ReadAll());
+    for (auto* child = rootNode->firstChild(); child != nullptr; child = child->nextSibling()) {
+        if (child->nodeName() != "Contents") {
+            continue;
+        }
+        Objects.emplace_back().Deserialize(*child);
     }
-    if (auto nextToken = node.FirstChild("NextContinuationToken"); !nextToken.IsNull()) {
-        NextContinuationToken = nextToken.Value<TString>();
+    if (auto nextToken = TryGetChildByName(*rootNode, "NextContinuationToken"); nextToken != nullptr) {
```
achulkov2: just `nextToken` instead of `nextToken != nullptr`?

yt/yt/library/s3/client.h
pavel-bash: Well, it's normal size, but it also has lots of includes, etc... so I've decided it'd be better to just have it forward-declared.

yt/yt/library/s3/client.cpp
pavel-bash: Yes, to unify and avoid the extra namespace to be written every time. But if contribs are used with the original ones, I'll of course remove this.

pavel-bash:
@achulkov2 ready for another look.

yt/yt/library/s3/client.cpp
```diff
@@ -6,41 +6,107 @@
...
+    const Poco::XML::Node* operator->() const
+    {
+        return RootNode;
+    }
+
+    //! Parsed document; this pointer also manages the lifetime.
+    TPocoXmlDocumentPtr Document;
+
+    //! Pointer to the root node of this document.
+    Poco::XML::Node* RootNode;
```
achulkov2: IMO this should be more generic, no reason for the node to be the root node per se. It could be any node within the document, and it would still be memory-safe, right?

And maybe rename this class then, since it acts like a node? Like `TXmlNodeHolder` maybe, not the perfect name, but works.

yt/yt/library/s3/client.cpp
```diff
@@ -6,41 +6,107 @@
...
+    THROW_ERROR_EXCEPTION("The document does not have a non-comment root node");
+}
+
+struct TXmlParsedDocument
+{
+    Poco::XML::Node& operator*()
+    {
+        return *RootNode;
+    }
+    const Poco::XML::Node& operator*() const
```
achulkov2, line 42: We prefer to separate struct/class methods by newlines.

pavel-bash:
@Kontakter this PR is ready to be reviewed.

yt/yt/library/s3/client.cpp
```diff
@@ -6,41 +6,110 @@
...
-NXml::TDocument ParseXmlDocument(TSharedRef responseBody)
+Poco::XML::Node* GetRootNode(TPocoXmlDocumentPtr document)
+{
+    for (auto* child = document->firstChild(); child != nullptr; child = child->nextSibling()) {
+        // Skip the comment nodes.
+        if (child->nodeType() == Poco::XML::Node::ELEMENT_NODE) {
+            return child;
+        }
+    }
+    THROW_ERROR_EXCEPTION("The document does not have a non-comment root node");
```
achulkov2: Maybe `XML document does not...`?

yt/yt/library/s3/client.cpp
```diff
@@ -6,41 +6,110 @@
...
+    TPocoXmlDocumentPtr parsedDocument = Poco::XML::DOMParser{}.parseString(responseString);
+    return {
+        .Document = parsedDocument,
+        .Node = GetRootNode(parsedDocument),
+    };
+}
+
+Poco::XML::Node* GetChildByName(const Poco::XML::Node& node, const std::string& childName)
+{
+    for (auto* child = node.firstChild(); child != nullptr; child = child->nextSibling()) {
```
achulkov2: Asked about `!= nullptr` with respect to our codestyle. I do actually get the extra clarity, sometimes I also want to write it explicitly, but up until this point I have tried to simplify it. 

Let's wait for responses and change/keep as is depending on the result. (there are a few usages like this across the PR).

achulkov2:
Hi, @savnadya can you take a quick look/delegate to someone? Should be an easy one.

yt/yt/library/s3/client.cpp
Kontakter: I'd prefer something like this: `Failed to find non-comment root node in XML document`

yt/yt/library/s3/client.cpp
```diff
@@ -6,41 +6,110 @@
...
+        if (child->nodeName() == childName) {
+            return child;
+        }
+    }
+    THROW_ERROR_EXCEPTION("Child with name %Qv not found in node %Qv", childName, node.nodeName());
+}
+
+Poco::XML::Node* TryGetChildByName(const Poco::XML::Node& node, const std::string& childName)
```
Kontakter: We normally call such methods with verb `Find`: `FindChildByName`

yt/yt/library/s3/client.cpp
```diff
@@ -6,41 +6,110 @@
...
-        NXml::TDocument::Source::String);
+    std::string responseString(responseBody.ToStringBuf());
+    TPocoXmlDocumentPtr parsedDocument = Poco::XML::DOMParser{}.parseString(responseString);
+    return {
+        .Document = parsedDocument,
+        .Node = GetRootNode(parsedDocument),
+    };
+}
+
+Poco::XML::Node* GetChildByName(const Poco::XML::Node& node, const std::string& childName)
```
Kontakter: `GetChildByNameOrThrow`

Kontakter:
Shipped!

Kontakter:
(to run tests in internal CI)

Kontakter:
We have discuss it in SIG-chat, I am waiting proper improvements of naming

pavel-bash:
@Kontakter I've fixed the names of the methods.

Kontakter:
Shipped!

savnadya:
LGTM. However, it would be cleaner to split the new Delete method and the XML library change into separate PRs.

Kontakter:
> LGTM. However, it would be cleaner to split the new Delete method and the XML library change into separate PRs.

Discussed internally, and decided to merge this PR as is

PR #1369, opened 2025-07-10
fix NPE in GenerateDockerAuthFromToken

faucct:
It was introduced in https://github.com/ytsaurus/ytsaurus/pull/1047.
Still no tests, as I could not test CRI last time.

### DIFF
--- a/yt/yt/server/controller_agent/controllers/helpers.cpp
+++ b/yt/yt/server/controller_agent/controllers/helpers.cpp
@@ -401,7 +401,9 @@ void GenerateDockerAuthFromToken(
     const std::string& authenticatedUser,
     NControllerAgent::NProto::TUserJobSpec* jobSpec)
 {
-    if (!secureVault) return;
+    if (!secureVault) {
+        return;
+    }
     auto findEnv = [&] (const TStringBuf& key) -> std::optional<TString> {
         auto child = secureVault->FindChild(std::string(key));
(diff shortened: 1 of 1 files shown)


yt/yt/server/controller_agent/controllers/helpers.cpp
```diff
@@ -405,7 +405,7 @@ void GenerateDockerAuthFromToken(
         auto child = secureVault->FindChild(std::string(key));
         return child && child->GetType() == ENodeType::String ? std::optional(child->AsString()->GetValue()) : std::nullopt;
     };
-    if (!findEnv(DockerAuthEnv)) {
+    if (secureVault && !findEnv(DockerAuthEnv)) {
```
achulkov2: Control flow is kind of non-obvious, findEnv is silently capturing and using secureValut, and we check it as a pre-condition without it being directly passed to `findEnv`. (and yes it is right above, but still it is a hidden knowledge moment).

IMO `if (!secureVault) return` in the very beginning will look better.

yt/yt/server/controller_agent/controllers/helpers.cpp
```diff
@@ -401,6 +401,7 @@ void GenerateDockerAuthFromToken(
     const std::string& authenticatedUser,
     NControllerAgent::NProto::TUserJobSpec* jobSpec)
 {
+    if (!secureVault) return;
```
Kontakter: Use style guide please.

```cpp
if (!secureVault) {
    return;
}
```

Kontakter:
Shipped!

PR #1373, opened 2025-07-15
Wrap slot location HeavyIO action queue in 500ms watchdog invoker

Krock21:
We saw that HeavyIO action_queue was overloaded and actions were waiting for 100+ seconds. Wrapping it in watchdog invoker will help debugging such cases


koct9i:
I'm not sure that 500ms timeout is a good choice here.
Probably 30-60 seconds is a more reasonable value.

yt/yt/server/node/exec_node/slot_location.cpp
```diff
@@ -76,7 +76,10 @@ TSlotLocation::TSlotLocation(
     , SlotIndexToUserId_(slotIndexToUserId)
     , HeavyLocationQueue_(New<TActionQueue>(Format("HeavyIO:%v", id)))
     , LightLocationQueue_(New<TActionQueue>(Format("LightIO:%v", id)))
-    , HeavyInvoker_(HeavyLocationQueue_->GetInvoker())
+    , HeavyInvoker_(CreateWatchdogInvoker(
+        HeavyLocationQueue_->GetInvoker(),
+        ExecNodeLogger(),
+        TDuration::MilliSeconds(500)))
```
Kontakter: I think this value should be added to config

Krock21:
Fixed comments:
Extracted timeout to config, set default to 30 seconds

PR #1374, opened 2025-07-15
Add require_administer_for_password_set option

achulkov2:
Useful for restricting usage of passwords by non-admins/in non-special circumstances in favor of SSO authentication.

* Changelog entry
Type: feature
Component: proxy

Add `require_administer_for_password_set` option to allow forcing the `administer` permission requirement for running `set-user-password`.

### DIFF
--- a/yt/yt/client/driver/driver.cpp
+++ b/yt/yt/client/driver/driver.cpp
@@ -455,6 +455,7 @@ public:
 
         auto options = TClientOptions::FromAuthenticationIdentity(identity);
         options.RequirePasswordInAuthenticationCommands = Config_->RequirePasswordInAuthenticationCommands;
+        options.RequireAdministerForPasswordSet = Config_->RequireAdministerForPasswordSet;
         options.Token = request.UserToken;
         options.ServiceTicketAuth = request.ServiceTicket
             ? std::make_optional(New<NAuth::TServiceTicketFixedAuth>(*request.ServiceTicket))
--- a/yt/yt/ytlib/api/native/client_authentication_impl.cpp
+++ b/yt/yt/ytlib/api/native/client_authentication_impl.cpp
@@ -51,7 +51,8 @@ void TClient::DoSetUserPassword(
         "Password change",
         user,
         currentPasswordSha256,
-        options);
+        options,
+        Options_.RequireAdministerForPasswordSet);
 
     constexpr int PasswordSaltLength = 16;
     auto newPasswordSaltBytes = GenerateCryptoStrongRandomString(PasswordSaltLength);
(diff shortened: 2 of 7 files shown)


ilyaibraev:
We are waiting for @dim-an, he will watch PR in a week.

dim-an:
I'm sorry for delay. I'll take a look at this PR this week.

yt/yt/ytlib/api/native/client_authentication_impl.cpp
```diff
@@ -410,6 +415,15 @@ void TClient::ValidateAuthenticationCommandPermissions(
...
 
+    if (requireAdminister && !canAdminister) {
+        THROW_ERROR_EXCEPTION(
+            "%v can be performed only by a user having %Qlv permission on the user",
+            action,
+            EPermission::Administer)
+            << TErrorAttribute("user", user)
+            << TErrorAttribute("authenticated_user", Options_.User);
+    }
+
```
dim-an, line 426: Don't we want to move this `if` down inside `if (!canAdminister)`

yt/yt/client/driver/driver.cpp
```diff
@@ -455,6 +455,7 @@ class TDriver
 
         auto options = TClientOptions::FromAuthenticationIdentity(identity);
         options.RequirePasswordInAuthenticationCommands = Config_->RequirePasswordInAuthenticationCommands;
+        options.RequireAdministerForPasswordSet = Config_->RequireAdministerForPasswordSet;
```
dim-an, line 458: Looks like user still can change their password using RPC proxies?

yt/yt/ytlib/api/native/client_authentication_impl.cpp
achulkov2: Sure, I may have thought it was more concise as a separate if, but I agree with you on second thought.

yt/yt/client/driver/driver.cpp
achulkov2: You would think so, but [muahaha](https://github.com/ytsaurus/ytsaurus/blob/d2715d428605c2f9654d9a200cd0156577e49dc1/yt/yt/client/api/rpc_proxy/client_impl.cpp#L2652)!

If implemented, there are some existing options that might have to move.

dim-an:
If I understand correctly you want to patch your static configuration for http proxies. It looks to me that the better place would be some dynconfig that is checked inside SetPassword.

Dynamic config would work automatically when rpc is implemented and there would be single place for this behaviour configuration.

dim-an:
What do you think about putting an option into `TConnectionDynamicConfig`?

achulkov2:
> What do you think about putting an option into `TConnectionDynamicConfig`?

The reason that I placed the new option into the static config was that there already is a similar option right nearby — `require_password_in_authentication_commands`.
I do agree with you in principle that it is a somewhat weird place for both of these options, but our initial reasoning for this change was to comply with some audit requirements, so taking the path of least resistance was the way to go.

If that is what you want, I will take another look and relocate these two knobs, but not earlier than next quarter, since this one has more pressing issues.

dim-an:
Closing for now, can be reopened after rework.

PR #1376, opened 2025-07-16
yt/admin_service: add method "ping_node"

koct9i:
This method verifies cluster interconnect connectivity and latency.
It is useful for testing mTLS and complex distributed clusters.

Ping request could include chain of addresses to check connectivity
between nodes within cluster and measure latency of overall call chain.

This method is allows only for superuser, chain is limited with 10 hops.
Size of payload attachment is limited with 10 MiB.

---

* Changelog entry
Type: feature
Component: misc-server

Add method "ping_node" for checking cluster interconnect connectivity.


Kontakter:
Will you submit it later?

PR #1378, opened 2025-07-22
Go SDK: Added support for multilookup requests

mtvch:
* Changelog entry
Type: feature
Component: go-sdk

Added support for multilookup requests.

### DIFF
--- a/yt/go/yt/interface.go
+++ b/yt/go/yt/interface.go
@@ -1261,6 +1261,30 @@ type LookupRowsOptions struct {
 	*TransactionOptions
 }
 
+type MultiLookupSubrequest struct {
+	Path                ypath.Path
+	KeepMissingRows     *bool
+	EnablePartialResult *bool
+	UseLookupCache      *bool
+	Columns             []string
+	Keys                []any
+}
+
+type TabletReadOptions struct {
+	ReadFrom                  *TabletReadKind
+	CachedSyncReplicasTimeout *uint64
+}
+
+type MultiLookupRowsOptions struct {
+	Timestamp          *Timestamp
+	RetentionTimestamp *Timestamp
+	TabletReadOptions  *TabletReadOptions
+	ReplicaConsistency *ReplicaConsistency
+	MultiplexingBand   *MultiplexingBand
+
+	*TransactionOptions
+}
+
 type InsertRowsOptions struct {
 	Atomicity          *Atomicity `http:"atomicity,omitnil"`
 	RequireSyncReplica *bool      `http:"require_sync_replica,omitnil"`
@@ -1345,6 +1369,19 @@ type TabletClient interface {
 		keys []any,
 		options *LookupRowsOptions,
 	) (r TableReader, err error)
+
+	// MultiLookupRows performs lookup by key across multiple tables.
+	//
+	// This method allows performing lookup operations on multiple tables in a single request,
+	// which can be more efficient than making separate LookupRows calls.
+	//
+	// Returns a slice of TableReaders, one for each subrequest in the same order.
+	// Each TableReader contains the lookup results for the corresponding table.
+	MultiLookupRows(
+		ctx context.Context,
+		subrequests []MultiLookupSubrequest,
+		options *MultiLookupRowsOptions,
+	) (readers []TableReader, err error)
 
 	// LockRows acquires lock for given keys, without changing row values.
 	//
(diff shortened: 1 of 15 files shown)


tinarsky:
Hello, I'll check the pr this week.

yt/go/yt/interface.go
```diff
@@ -1261,6 +1261,30 @@ type LookupRowsOptions struct {
...
+	Keys                []any
+}
+
+type TabletReadOptions struct {
+	ReadFrom                  *TabletReadKind
+	CachedSyncReplicasTimeout *uint64
+}
+
+type MultiLookupRowsOptions struct {
+	Timestamp          *Timestamp
```
tinarsky: Let's specify the field names in the `http` tag (see `LookupRowsOptions`). And then regenerate params_gen.go.

yt/go/yt/interface.go
```diff
@@ -1261,6 +1261,30 @@ type LookupRowsOptions struct {
...
+	RetentionTimestamp *Timestamp
+	TabletReadOptions  *TabletReadOptions
```
tinarsky: Let's embed `TabletReadOptions` as `TransactionOptions`

yt/go/yt/interface.go
```diff
@@ -1261,6 +1261,30 @@ type LookupRowsOptions struct {
...
+	Path                ypath.Path
+	KeepMissingRows     *bool
+	EnablePartialResult *bool
+	UseLookupCache      *bool
+	Columns             []string
+	Keys                []any
+}
+
+type TabletReadOptions struct {
+	ReadFrom                  *TabletReadKind
```
tinarsky, line 1276: Same: let's add `http` tags.

yt/go/yt/internal/rpcclient/tx.go
```diff
@@ -214,3 +215,21 @@ func (t *TxInterceptor) ReadRow(
...
+	ctx context.Context,
+	call *Call,
+	next MultiLookupInvoker,
+	rsp ProtoMultiLookupResp,
+) (readers []yt.TableReader, err error) {
+	if err = t.setTxID(call); err != nil {
+		return
+	}
+
+	readers, err = next(ctx, call, rsp)
```
tinarsky: `return next(ctx, call, rsp)`?

tinarsky (changes requested):

yt/go/yt/interface.go
mtvch: Why do we need http tags if multilookup is only implemented in RPC proxy?

yt/go/yt/interface.go
mtvch: Why do you think it would be better? As I see `tablet_read_options` are used only for lookups in rpc proxy and transaction options are more general

yt/go/yt/internal/rpcclient/tx.go
mtvch: Sure, fixed

yt/go/yt/interface.go
tinarsky: Ok, I agree

yt/go/yt/interface.go
tinarsky: In this file all options are embedded, so I suggest keeping the same style (e.g. [TabletRangeOptions](https://github.com/ytsaurus/ytsaurus/blob/main/yt/go/yt/interface.go#L1160)).

yt/go/yt/interface.go
```diff
@@ -1261,6 +1261,30 @@ type LookupRowsOptions struct {
...
+	CachedSyncReplicasTimeout *uint64
+}
+
+type MultiLookupRowsOptions struct {
+	Timestamp          *Timestamp
+	RetentionTimestamp *Timestamp
+	TabletReadOptions  *TabletReadOptions
+	ReplicaConsistency *ReplicaConsistency
```
tinarsky: Let's move `ReplicaConsistency` and `RetentionTimestamp` to TabletReadOptions, similar to the C++ API — https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/client/api/client_common.h#L128

yt/go/yt/interface.go
```diff
@@ -1261,6 +1261,30 @@ type LookupRowsOptions struct {
...
+	MultiplexingBand   *MultiplexingBand
```
tinarsky: I propose to extract MultiplexingBand into [MultiplexingBandOptions](https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/client/api/client_common.h#L29) (similar to the C++ API, perhaps this is done for future extensibility) and embed it as well.

yt/go/yt/interface.go
mtvch: Got it, I misunderstood you

yt/go/yt/interface.go
mtvch: Done

yt/go/yt/interface.go
mtvch: Done

tinarsky (approved):

PR #1390, opened 2025-08-01
Fix read results with yson fields by reference.

Tony-Romanov:
Type: fix
Component: query-tracker


Krock21:
Could you please describe the issue you are fixing and add tests(if possible in reasonable time) that demonstrate it?
Why other table readers with default `.EnableAnyUnpacking = true` don't have this issue? For example, in read-table

Tony-Romanov:
> Could you please describe the issue you are fixing and add tests(if possible in reasonable time) that demonstrate it? Why other table readers with default `.EnableAnyUnpacking = true` don't have this issue? For example, in read-table

The values of TUnversionedValue obtained as a result of this reading are written as is in the `rowset` field. It is obvious that the `rowset` field should not change depending on the method of transferring the result: by value or by reference. And now, when transferring by reference, the values of TUnversionedValue with the `Any` type are converted to the `String` type because of `EnableAnyUnpackig` flag by default. Which additionally leads to errors in their parsing later if there was an empty string there.

yt/yql/agent/interop.cpp
```diff
@@ -164,7 +164,7 @@ TYqlRowset BuildRowsetByRef(
 
         YT_LOG_DEBUG("Opening static table reader (Path: %v)", path);
 
-        auto reader = WaitFor(client->CreateTableReader(path))
+        auto reader = WaitFor(client->CreateTableReader(path, NApi::TTableReaderOptions{.EnableAnyUnpacking = false}))
```
Krisha11, line 180: Maybe we can add some comment here for future readers?

PR #1396, opened 2025-08-07
yt/rpc_proxy: do not announce PublicRpc address fallback to InternalRpc

koct9i:
Old client are confused by new enum value in proxy discovery response.
Hide RPC proxy PublicRpc address unless we really have separate port.

---

* Changelog entry
Type: fix
Component: rpc-proxy

Remove RPC proxy PublicRpc address fallback to InternalRpc


yalekseev:
Shipped!

PR #1399, opened 2025-08-12
[stable/25.1] cherry-pick CRI/CRI-O patches

koct9i:
- **Revalidate CRI Image Cache on every cache hit**
- **Add option for CRI runtime endpoint in yt-local and integration tests**
- **Handle CRI error reply on getting container status**
- **Update CRI API proto file**
- **Add docker image id into TCriImageDescriptor**
- **Fix package.json syntax for ytsaurus docker image**
- **Add CRI-O and crictl into ytserver docker image**


Kontakter:
Shipped!

Issue #1402, opened 2025-08-13
Go SDK does not support deserialization of composite types


mtvch:
When calling `TableReader.Scan` for a row that has a column of composite type, I get following error:
`unable to decode response from wire format: unable to deserialize attachments: invalid wire type 0x12`

Need to add support for deserialization of composite types in Go SDK. I can create a PR for that

tinarsky:
Hello, it would be great if you create a pr

tinarsky:
The [pull request](https://github.com/ytsaurus/ytsaurus/pull/1403) has been merged, I am closing the issue. Thanks for your contribution!

PR #1403, opened 2025-08-13
Go SDK: Deserialization of composite types

mtvch:
Fixes for https://github.com/ytsaurus/ytsaurus/issues/1402

* Changelog entry
Type: feature
Component: go-sdk

Added deserialization of composite types

### DIFF
--- a/yt/go/schema/converter.go
+++ b/yt/go/schema/converter.go
@@ -8,7 +8,7 @@ import (
 	"go.ytsaurus.tech/yt/go/yson"
 )
 
-func ConvertFromRPCProxy(rpcSchema *rpc_proxy.TTableSchema) (Schema, error) {
+func ConvertFromProto(rpcSchema *rpc_proxy.TTableSchema) (Schema, error) {
 	schema := Schema{
 		Strict:     ptr.Bool(rpcSchema.GetStrict()),
 		UniqueKeys: rpcSchema.GetUniqueKeys(),
--- a/yt/go/schema/converter.go
+++ b/yt/go/schema/converter.go
@@ -63,7 +63,7 @@ func convertType(valueType *int32) Type {
 	// Based on ESimpleLogicalValueType from yt/yt/client/table_client/row_base.h
 	switch *valueType {
 	case 0x02: // Null
-		return TypeAny // No direct null type in schema, use Any
+		return TypeNull
 	case 0x03:
 		return TypeInt64
 	case 0x04:
--- a/yt/go/schema/converter.go
+++ b/yt/go/schema/converter.go
@@ -10,8 +10,8 @@ import (
 
 func ConvertFromRPCProxy(rpcSchema *rpc_proxy.TTableSchema) (Schema, error) {
 	schema := Schema{
-		Strict:     ptr.Bool(*rpcSchema.Strict),
-		UniqueKeys: *rpcSchema.UniqueKeys,
+		Strict:     ptr.Bool(rpcSchema.GetStrict()),
+		UniqueKeys: rpcSchema.GetUniqueKeys(),
 		Columns:    make([]Column, len(rpcSchema.Columns)),
 	}
 
@@ -24,14 +24,14 @@ func ConvertFromRPCProxy(rpcSchema *rpc_proxy.TTableSchema) (Schema, error) {
 		}
 
 		schema.Columns[i] = Column{
-			Name:        *col.Name,
+			Name:        col.GetName(),
 			Type:        convertType(col.Type),
 			ComplexType: complexType,
-			Required:    *col.Required,
-			Group:       *col.Group,
-			Expression:  *col.Expression,
-			Aggregate:   AggregateFunction(*col.Aggregate),
-			Lock:        *col.Lock,
+			Required:    col.GetRequired(),
+			Group:       col.GetGroup(),
+			Expression:  col.GetExpression(),
+			Aggregate:   AggregateFunction(col.GetAggregate()),
+			Lock:        col.GetLock(),
 			SortOrder:   convertSortOrder(col.SortOrder),
 		}
 	}
--- a/yt/go/schema/converter.go
+++ b/yt/go/schema/converter.go
@@ -1,0 +1,108 @@
+// Package schema defines schema of YT tables.
+package schema
+
+import (
+	"fmt"
+
+	"go.ytsaurus.tech/library/go/ptr"
+	"go.ytsaurus.tech/yt/go/proto/client/api/rpc_proxy"
+	"go.ytsaurus.tech/yt/go/yson"
+)
+
+func ConvertFromRPCProxy(rpcSchema *rpc_proxy.TTableSchema) (Schema, error) {
+	schema := Schema{
+		Strict:     ptr.Bool(*rpcSchema.Strict),
+		UniqueKeys: *rpcSchema.UniqueKeys,
+		Columns:    make([]Column, len(rpcSchema.Columns)),
+	}
+
+	for i, col := range rpcSchema.Columns {
+		var complexType ComplexType
+		if len(col.GetTypeV3()) > 0 {
+			if err := yson.Unmarshal(col.GetTypeV3(), &complexType); err != nil {
+				return Schema{}, fmt.Errorf("failed to unmarshal complex type: %w", err)
+			}
+		}
+
+		schema.Columns[i] = Column{
+			Name:        *col.Name,
+			Type:        convertType(col.Type),
+			ComplexType: complexType,
+			Required:    *col.Required,
+			Group:       *col.Group,
+			Expression:  *col.Expression,
+			Aggregate:   AggregateFunction(*col.Aggregate),
+			Lock:        *col.Lock,
+			SortOrder:   convertSortOrder(col.SortOrder),
+		}
+	}
+
+	return schema, nil
+}
+
+func convertSortOrder(sortOrder *int32) SortOrder {
+	if sortOrder == nil {
+		return SortNone
+	}
+
+	switch *sortOrder {
+	case 0:
+		return SortAscending
+	case 1:
+		return SortDescending
+	}
+
+	return SortNone
+}
+
+func convertType(valueType *int32) Type {
+	if valueType == nil {
+		return TypeAny
+	}
+
+	// Map RPC Schema enum values to schema.Type
+	// Based on ESimpleLogicalValueType from yt/yt/client/table_client/row_base.h
+	switch *valueType {
+	case 0x02: // Null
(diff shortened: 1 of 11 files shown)


mtvch:
What I have done is simply treating composite types as YSON. But it seems that there are some optimizations: e.g. structs are stored as YSON lists and we need schema information to parse lists into structs, isn't it?

mtvch:
> What I have done is simply treating composite types as YSON. But it seems that there are some optimizations: e.g. structs are stored as YSON lists and we need schema information to parse lists into structs, isn't it?

So, I found this specification. I will try to implement deserialization according to it for types that are currently supported by Go client.

https://ytsaurus.tech/docs/en/user-guide/storage/data-types#formats

mtvch:
> > What I have done is simply treating composite types as YSON. But it seems that there are some optimizations: e.g. structs are stored as YSON lists and we need schema information to parse lists into structs, isn't it?
> 
> 
> 
> So, I found this specification. I will try to implement deserialization according to it for types that are currently supported by Go client.
> 
> 
> 
> https://ytsaurus.tech/docs/en/user-guide/storage/data-types#formats

Done

tinarsky:
Hello, I will check this pr in the next few days

yt/go/wire/decode.go
```diff
@@ -183,10 +188,11 @@ func (e *ReflectTypeError) Error() string {
 
 type WireDecoder struct {
 	NameTable NameTable
+	Schema    *rpc_proxy.TTableSchema
 }
 
-func NewDecoder(table NameTable) *WireDecoder {
-	return &WireDecoder{NameTable: table}
+func NewDecoder(table NameTable, schema *rpc_proxy.TTableSchema) *WireDecoder {
```
tinarsky: I suggest using [schema.Schema](https://github.com/ytsaurus/ytsaurus/blob/main/yt/go/schema/schema.go#L161) here (and creating a converter from rpc_proxy.TTableSchema to schema.Schema in package schema).

yt/go/wire/decode.go
```diff
@@ -265,13 +271,337 @@ func (d *WireDecoder) decodeValueGeneric(value Value, v *any) (err error) {
...
+		*v = result
+		return nil
+	}
+
+	return fmt.Errorf("expected list or map for dict, got %T", data)
+}
+
+func (d *WireDecoder) decodeVariant(data any, t schema.Variant, v *any) error {
+	// Variant is represented as a YSON list of length 2: [alternative, value]
+	if ysonList, ok := data.([]any); ok && len(ysonList) == 2 {
```
tinarsky: Please reduce the number of nesting levels here.

yt/go/wire/decode.go
```diff
@@ -265,13 +271,337 @@ func (d *WireDecoder) decodeValueGeneric(value Value, v *any) (err error) {
...
+
+func (d *WireDecoder) decodePrimitiveType(data any, t schema.Type, v *any) error {
+	switch t {
+	case schema.TypeDate:
+		return d.decodeDate(data, v)
+	case schema.TypeDatetime:
+		return d.decodeDatetime(data, v)
+	case schema.TypeTimestamp:
+		return d.decodeTimestamp(data, v)
+	default:
```
tinarsky, line 345: case schema.TypeInterval?

yt/go/wire/decode_test.go
```diff
@@ -348,8 +350,729 @@ func TestDecoder_UnmarshalRow(t *testing.T) {
...
+
+func TestDecoder_decodeReflect(t *testing.T) {
+	nameTable := NameTable{
+		{Name: "composite_field"},
+	}
+
+	d := NewDecoder(nameTable, nil)
+
+	type testStruct struct {
+		CompositeField any `yson:"composite_field"`
```
tinarsky: Let's add a test with fields of type map[string]any, []any, struct, and schema.Timestamp.

yt/go/wire/decode.go
```diff
@@ -265,13 +271,337 @@ func (d *WireDecoder) decodeValueGeneric(value Value, v *any) (err error) {
...
+
+		// named variants
+		if altName, ok := alternative.(string); ok {
+			for _, member := range t.Members {
+				if member.Name == altName {
+					var decodedValue any
+					if err := d.decodeSchemaType(value, member.Type, &decodedValue); err != nil {
+						return err
+					}
+					*v = decodedValue
```
tinarsky: The Variant should be returned in [this](https://github.com/ytsaurus/ytsaurus/blob/main/yt/go/skiff/decoder.go#L529) format. Also, please make sure that deserialization of the other types matches the deserialization in skiff/yson.

yt/go/wire/decode.go
mtvch: Done

yt/go/wire/decode.go
mtvch: Interval is simply int64 and can be decoded as primitive type (in default case), isn't it? But since you asked for readability I added explicit decoding for interval

yt/go/wire/decode.go
mtvch: Done.

Found 1 inconsistency: in `wire/decode` `schema.Dict` is decoded into `map[any][any]` while in `skiff/decoder` its decoded into `[]any` (where each element is `[]any{key, value}`).

I find `wire/decode` format more convenient for end user, can we keep it? Or do we need strict consistency here? Are there any reasons why it's decoded like this in `skiff/decoder`?

yt/go/wire/decode.go
mtvch: Done

yt/go/wire/decode_test.go
mtvch: I've added tests that show decoding of composite types into structure fields of types map[string]any, []any, and schema.Timestamp.

But my implementation does not support decoding into field of struct type as schema struct is decoded into map[string]any.

I don't think that this is really needed at this stage, it can be implemented later if there will be such need.

tinarsky: Ok, what error will we return now if the user passes a structure with a field of type struct? Let's return an error in this case that the field of this type is not supported.

yt/go/wire/decode.go
tinarsky: Yes, we need to keep consistency here.

> Are there any reasons why it's decoded like this in skiff/decoder?

It’s decoded this way because `yson.decodeGeneric` handles schema.Dict the same way.

mtvch: Ok, I guess it makes sense since keys are not unique in YTsaurus dicts
https://ytsaurus.tech/docs/en/user-guide/storage/data-types#schema_dict

I also removed support for decoding dicts with format string_keyed_dict_mode=named because it's impossible to receive this format from RPC proxy as you can't pass YSON format in request.

I also noticed that HTTP client uses string_keyed_dict_mode=named and there dicts will be decoded as map. Honestly, I find all this confusing

yt/go/schema/converter.go
```diff
@@ -0,0 +1,106 @@
...
+				return Schema{}, fmt.Errorf("failed to unmarshal complex type: %w", err)
+			}
+		}
+
+		schema.Columns[i] = Column{
+			Name:        *col.Name,
+			Type:        convertType(col.Type),
+			ComplexType: complexType,
+			Required:    *col.Required,
+			Group:       *col.Group,
```
heromantor: Plz, use getters for access to protobuf message fields. Deref col members sometime panics

mtvch: Sure, thank you!

yt/go/wire/decode_test.go
mtvch: Done, now type mismatch error will be returned

tinarsky:
I will review the pull request again in the next few days.

yt/go/schema/converter.go
```diff
@@ -0,0 +1,106 @@
...
+
+import (
+	"fmt"
+
+	"go.ytsaurus.tech/library/go/ptr"
+	"go.ytsaurus.tech/yt/go/proto/client/api/rpc_proxy"
+	"go.ytsaurus.tech/yt/go/yson"
+)
+
+func ConvertFromRPCProxy(rpcSchema *rpc_proxy.TTableSchema) (Schema, error) {
```
tinarsky: I propose to rename this function to `ConvertFromProto`.

yt/go/schema/converter.go
```diff
@@ -0,0 +1,106 @@
...
+func convertType(valueType *int32) Type {
+	if valueType == nil {
+		return TypeAny
+	}
+
+	// Map RPC Schema enum values to schema.Type
+	// Based on ESimpleLogicalValueType from yt/yt/client/table_client/row_base.h
+	switch *valueType {
+	case 0x02: // Null
+		return TypeAny // No direct null type in schema, use Any
```
tinarsky: Let's add `TypeNull`

yt/go/schema/converter.go
```diff
@@ -0,0 +1,106 @@
+package schema
```
tinarsky, line 1: schema/converter.go and schema/converter_test.go are missing from yt/go/schema/ya.make. Please add them there.

yt/go/wire/decode.go
```diff
@@ -265,13 +270,328 @@ func (d *WireDecoder) decodeValueGeneric(value Value, v *any) (err error) {
...
 
+func (d *WireDecoder) decodeValueComposite(value Value, v *any) error {
+	entry, err := d.getNameTableEntry(value)
+	if err != nil {
+		return err
+	}
+
+	var columnSchema *schema.Column
+	if d.Schema != nil {
+		for _, col := range d.Schema.Columns {
```
tinarsky, line 290: Let's create a `map[string]*schema.Column` in the decoder constructor so we don't have to search for the column in the loop.

mtvch: Do you think it really makes sense here? It won't make difference in performance even if we have table with 100s of columns, but it sure will make solution a bit more complicated

tinarsky: Ok, let's not do such optimization for now, the diff of pr is already big.

yt/go/schema/converter.go
mtvch: Done

yt/go/schema/converter.go
mtvch: Done

yt/go/schema/converter.go
mtvch: Done. I added TypeNull to schema and also added convertion for skiff wire type to TypeNothing. idk if that's correct and if this change should be considered anywhere else in the codebase. Double check that, please.

tinarsky (approved):

PR #1357, opened 2025-07-04
yt/core/https: verify server certificate

koct9i:
- **yt/yt/core/test_framework: update and refactor test keys**

    Present certificate are expired:
    
       Issuer: C = CN, ST = Some-State, O = Internet Widgits Pty Ltd
       Validity
           Not Before: Nov  1 06:59:16 2014 GMT
           Not After : Oct 29 06:59:16 2024 GMT
       Subject: C = CN, ST = Some-State, O = Internet Widgits Pty Ltd
    
       Issuer: C = US, O = Let's Encrypt, CN = Let's Encrypt Authority X3
       Validity
           Not Before: Feb 12 19:27:22 2018 GMT
           Not After : May 13 19:27:22 2018 GMT
       Subject: CN = cpp.manytask.org
    
       Issuer: O = Digital Signature Trust Co., CN = DST Root CA X3
       Validity
           Not Before: Mar 17 16:40:46 2016 GMT
           Not After : Mar 17 16:40:46 2021 GMT
       Subject: C = US, O = Let's Encrypt, CN = Let's Encrypt Authority X3
    
    Reuse test certificate from yt/yt/core/bus for "localhost".
    
       Issuer: C = RU, ST = Moscow, O = Yandex, CN = CA@localhost
       Validity
           Not Before: Jul 13 11:51:17 2023 GMT
           Not After : Jul 10 11:51:17 2033 GMT
       Subject: C = RU, ST = Moscow, O = Yandex, CN = localhost
    
       Issuer: C = RU, ST = Moscow, O = Yandex, CN = CA@localhost
       Validity
           Not Before: Jul 13 11:51:11 2023 GMT
           Not After : Jul 10 11:51:11 2033 GMT
       Subject: C = RU, ST = Moscow, O = Yandex, CN = CA@localhost

- **yt/core/https: verify server certificate**

NYT::NCrypto::TTlsConnection should verify server certificate and hostname.

---

* Changelog entry
Type: fix
Component: misc-server

Verify server certificate

PR #1419, opened 2025-09-01
yt/yt/core/https: make secure by default

koct9i:
Let's make HTTPS secure by default. Like it always supposed to be.

Link: https://github.com/ytsaurus/ytsaurus/pull/1357

---

* Changelog entry
Type: Fix
Component: core

Make HTTPS client secure by default.


Kontakter:
Shipped!

Kontakter:
Unfortunately some non-YT tests fail, I need some time to investigate the reasons

PR #1410, opened 2025-08-18
yt/environment: enable mutual TLS for cluster connection

koct9i:
Mutual TLS is a simple way to secure internal RPC once and for all.
Once it is working there is no much reason to use it partially
(i.e. only HTTPS or RPC proxy BUS TLS without internal RPC mTLS).

Environment "YT_ENABLE_TLS=1" activates mTLS for all integration tests.

Signed-off-by: Konstantin Khlebnikov <khlebnikov@tracto.ai>

**Depends on: https://github.com/ytsaurus/ytsaurus/pull/1333**

PR #1420, opened 2025-09-01
yt/environment: enable mutual TLS for cluster connection

koct9i:
Mutual TLS is a simple way to secure internal RPC once and for all.
If it is working there is no much reason to use it partially
(i.e. only HTTPS or RPC proxy BUS TLS without internal RPC mTLS).

Environment "YT_ENABLE_TLS=1" activates mTLS for all integration tests.

### DIFF
--- a/yt/python/yt/environment/api/__init__.py
+++ b/yt/python/yt/environment/api/__init__.py
@@ -81,6 +81,8 @@ class LocalYtConfig(object):
     public_ca_cert_key = attr.ib(None)
     rpc_cert = attr.ib(None)
     rpc_cert_key = attr.ib(None)
+    rpc_client_cert = attr.ib(None)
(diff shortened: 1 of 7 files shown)


koct9i:
Resend of #1410

Kontakter:
Shipped!

Kontakter:
(to run tests in internal CI)

Kontakter:
There are still errors in CHYT tests, I have made an investigation and realized that it can be reproduced only in internal CI and caused by using new code from this PR with enabled TVM in tests.

Future investigation will be done by @buyval01

Kontakter:
Shipped!

Kontakter:
@yalekseev have promised to take a look to this PR, after that we are ready to merge it.

PR #1424, opened 2025-09-04
Increase MaxMediumCount from 120 to 64'000

achulkov2:
On the road to S3 media it becomes clear that the current medium count limit of 120 per cluster is not enough once media become more user-facing objects, e.g. a client can configure their own bucket.

Conveniently, this limit is no longer justified by any real physical limitations, and it is relatively safe to increase. Still, it requires a few changes around the place.

From the looks of it we are lucky and no extra actions are required for backwards compatibility of new native clients with old masters before this change. Of course, newer masters with actual media with larger indexes won't work with older native clients, but that has never worked well.

* Changelog entry
Type: feature
Component: master

Increase MaxMediumCount from 120 to 64'000 and medium index size from 8 bits to 16 bits.

### DIFF
--- a/library/cpp/yt/compact_containers/compact_map-inl.h
+++ b/library/cpp/yt/compact_containers/compact_map-inl.h
@@ -1,0 +1,634 @@
+#ifndef COMPACT_MAP_INL_H_
+#error "Direct inclusion of this file is not allowed, include compact_map.h"
+#include "compact_map.h"
+#endif
+
+namespace NYT {
+
+////////////////////////////////////////////////////////////////////////////////
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+template <bool IsConst>
+class TCompactMap<TKey, TValue, N, TCompare, TAllocator>::TIteratorBase
+{
+private:
+    friend class TCompactMap<TKey, TValue, N, TCompare, TAllocator>;
+    friend class TIteratorBase<!IsConst>;
+    friend class iterator;
+    friend class const_iterator;
+
+    using TVecIter = std::conditional_t<IsConst, TVectorConstIterator, TVectorIterator>;
+    using TMapIter = std::conditional_t<IsConst, TMapConstIterator, TMapIterator>;
+    using TValueType = typename TCompactMap::value_type;
+    using TVectorValue = typename TCompactMap::TVectorValue;
+    using TSmallValue = std::conditional_t<IsConst, const TVectorValue, TVectorValue>;
+    using TSmallPointer = std::conditional_t<IsConst, const TVectorValue*, TVectorValue*>;
+    using TReference = std::conditional_t<IsConst, const TValueType&, TValueType&>;
+    using TPointer = std::conditional_t<IsConst, const TValueType*, TValueType*>;
+
+    static_assert(
+        sizeof(TVectorValue) == sizeof(TValueType),
+        "TCompactMap relies on identical layout of internal and public pair types");
+    static_assert(
+        alignof(TVectorValue) == alignof(TValueType),
+        "TCompactMap relies on identical layout of internal and public pair types");
+
+    static TReference VectorRef(TSmallValue& value)
+    {
+        // Technically, this violates strict aliasing rules, and is, therefore, undefined behavior.
+        // In practice, see the comment regarding value_type in the class header. It is defined
+        // with `may_alias` attribute to suppress optimizations based on type-based alias analysis.
+        return reinterpret_cast<TReference>(value);
+    }
+
+    static TPointer VectorPtr(TSmallPointer ptr)
+    {
+        // Technically, this violates strict aliasing rules, and is, therefore, undefined behavior.
+        // In practice, see the comment regarding value_type in the class header. It is defined
+        // with `may_alias` attribute to suppress optimizations based on type-based alias analysis.
+        return reinterpret_cast<TPointer>(ptr);
+    }
+
+    union
+    {
+        TVecIter VIter;
+        TMapIter MIter;
+    };
+    bool Small_;
+
+    explicit TIteratorBase(TVecIter it)
+        : VIter(it)
+        , Small_(true)
+    { }
+
+    explicit TIteratorBase(TMapIter it)
+        : MIter(it)
+        , Small_(false)
+    { }
+
+    // Conversion constructor for const_iterator from iterator.
+    template <bool WasConst, typename = std::enable_if_t<IsConst && !WasConst>>
+    TIteratorBase(const TIteratorBase<WasConst>& other)
+        : Small_(other.Small_)
+    {
+        if (Small_) {
+            VIter = other.VIter;
+        } else {
+            MIter = other.MIter;
+        }
+    }
+
+public:
+    using difference_type = std::ptrdiff_t;
+    using value_type = TValueType;
+    using reference = TReference;
+    using pointer = TPointer;
+    using iterator_category = std::bidirectional_iterator_tag;
+
+    reference operator*() const
+    {
+        if (Small_) {
+            return VectorRef(*VIter);
+        }
+        return *MIter;
+    }
+
+    pointer operator->() const
+    {
+        if (Small_) {
+            return VectorPtr(&*VIter);
+        }
+        return &*MIter;
+    }
+
+    TIteratorBase& operator++()
+    {
+        if (Small_) {
+            ++VIter;
+        } else {
+            ++MIter;
+        }
+        return *this;
+    }
+
+    TIteratorBase operator++(int)
+    {
+        TIteratorBase tmp = *this;
+        ++*this;
+        return tmp;
+    }
+
+    TIteratorBase& operator--()
+    {
+        if (Small_) {
+            --VIter;
+        } else {
+            --MIter;
+        }
+        return *this;
+    }
+
+    TIteratorBase operator--(int)
+    {
+        TIteratorBase tmp = *this;
+        --*this;
+        return tmp;
+    }
+
+    bool operator==(const TIteratorBase& other) const
+    {
+        if (Small_ != other.Small_) {
+            return false;
+        }
+        return Small_ ? VIter == other.VIter : MIter == other.MIter;
+    }
+
+    bool operator!=(const TIteratorBase& other) const
+    {
+        return !(*this == other);
+    }
+};
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+class TCompactMap<TKey, TValue, N, TCompare, TAllocator>::iterator
+    : public TCompactMap<TKey, TValue, N, TCompare, TAllocator>::template TIteratorBase<false>
+{
+public:
+    iterator() = default;
+
+private:
+    friend class TCompactMap<TKey, TValue, N, TCompare, TAllocator>;
+    friend class const_iterator;
+    using TBase = typename TCompactMap<TKey, TValue, N, TCompare, TAllocator>::template TIteratorBase<false>;
+
+    explicit iterator(TVectorIterator it)
+        : TBase(it)
+    { }
+
+    explicit iterator(TMapIterator it)
+        : TBase(it)
+    { }
+};
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+class TCompactMap<TKey, TValue, N, TCompare, TAllocator>::const_iterator
+    : public TCompactMap<TKey, TValue, N, TCompare, TAllocator>::template TIteratorBase<true>
+{
+public:
+    const_iterator() = default;
+    
+    // Allow implicit conversion from iterator to const_iterator.
+    const_iterator(const iterator& other)
+        : TBase(other.Small_ ? TBase(other.VIter) : TBase(other.MIter))
+    { }
+
+private:
+    friend class TCompactMap<TKey, TValue, N, TCompare, TAllocator>;
+    using TBase = typename TCompactMap<TKey, TValue, N, TCompare, TAllocator>::template TIteratorBase<true>;
+
+    explicit const_iterator(TVectorConstIterator it)
+        : TBase(it)
+    { }
+
+    explicit const_iterator(TMapConstIterator it)
+        : TBase(it)
+    { }
+};
+
+////////////////////////////////////////////////////////////////////////////////
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+TCompactMap<TKey, TValue, N, TCompare, TAllocator>::TCompactMap(const allocator_type& alloc)
+    : Alloc_(alloc)
+    , Map_(key_compare{}, alloc)
+{ }
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+template <class TIt>
+TCompactMap<TKey, TValue, N, TCompare, TAllocator>::TCompactMap(TIt first, TIt last)
+{
+    insert(first, last);
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+TCompactMap<TKey, TValue, N, TCompare, TAllocator>::TCompactMap(std::initializer_list<value_type> init)
+{
+    insert(init.begin(), init.end());
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+bool TCompactMap<TKey, TValue, N, TCompare, TAllocator>::empty() const
+{
+    return IsSmall() ? Vector_.empty() : Map_.empty();
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+typename TCompactMap<TKey, TValue, N, TCompare, TAllocator>::size_type
+TCompactMap<TKey, TValue, N, TCompare, TAllocator>::size() const
+{
+    return IsSmall() ? Vector_.size() : Map_.size();
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+void TCompactMap<TKey, TValue, N, TCompare, TAllocator>::clear()
+{
+    Vector_.clear();
+    Map_.clear();
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+auto TCompactMap<TKey, TValue, N, TCompare, TAllocator>::begin() -> iterator
+{
+    return BeginImpl(this);
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+auto TCompactMap<TKey, TValue, N, TCompare, TAllocator>::begin() const -> const_iterator
+{
+    return BeginImpl(this);
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+auto TCompactMap<TKey, TValue, N, TCompare, TAllocator>::cbegin() const -> const_iterator
+{
+    return begin();
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+auto TCompactMap<TKey, TValue, N, TCompare, TAllocator>::end() -> iterator
+{
+    return EndImpl(this);
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+auto TCompactMap<TKey, TValue, N, TCompare, TAllocator>::end() const -> const_iterator
+{
+    return EndImpl(this);
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+auto TCompactMap<TKey, TValue, N, TCompare, TAllocator>::cend() const -> const_iterator
+{
+    return end();
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+bool TCompactMap<TKey, TValue, N, TCompare, TAllocator>::IsSmall() const
+{
+    return Map_.empty();
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+void TCompactMap<TKey, TValue, N, TCompare, TAllocator>::UpgradeToMap()
+{
+    if (!IsSmall()) {
+        return;
+    }
+
+    Map_ = TMap(Compare_, Alloc_);
+    for (auto& item : Vector_) {
+        Map_.emplace(item.first, item.second);
+    }
+
+    Vector_.clear();
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+std::pair<typename TCompactMap<TKey, TValue, N, TCompare, TAllocator>::TVectorConstIterator, bool>
+TCompactMap<TKey, TValue, N, TCompare, TAllocator>::VectorLowerBound(const key_type& key) const
+{
+    return VectorLowerBoundImpl(Vector_.begin(), Vector_.end(), key, Compare_);
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+std::pair<typename TCompactMap<TKey, TValue, N, TCompare, TAllocator>::TVectorIterator, bool>
+TCompactMap<TKey, TValue, N, TCompare, TAllocator>::VectorLowerBound(const key_type& key)
+{
+    return VectorLowerBoundImpl(Vector_.begin(), Vector_.end(), key, Compare_);
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+template <class... TArgs>
+auto TCompactMap<TKey, TValue, N, TCompare, TAllocator>::EmplaceSmall(TVectorConstIterator pos, TArgs&&... args) -> TVectorIterator
+{
+    return Vector_.emplace(pos, std::forward<TArgs>(args)...);
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+template <class TVectorIteratorType>
+std::pair<TVectorIteratorType, bool>
+TCompactMap<TKey, TValue, N, TCompare, TAllocator>::VectorLowerBoundImpl(
+    TVectorIteratorType begin,
+    TVectorIteratorType end,
+    const key_type& key,
+    const key_compare& compare)
+{
+    auto comp = [&compare] (const auto& pair, const key_type& k) {
+        return compare(pair.first, k);
+    };
+
+    auto it = std::lower_bound(begin, end, key, comp);
+    bool found = (it != end && !compare(key, it->first));
+    return {it, found};
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+template <class TKeyParam>
+typename TCompactMap<TKey, TValue, N, TCompare, TAllocator>::mapped_type&
+TCompactMap<TKey, TValue, N, TCompare, TAllocator>::Subscript(TKeyParam&& key)
+{
+    if (IsSmall()) {
+        auto [vectorIt, found] = VectorLowerBound(key);
+
+        if (found) {
+            return vectorIt->second;
+        }
+
+        if (Vector_.size() < N) {
+            auto inserted = EmplaceSmall(vectorIt, std::forward<TKeyParam>(key), mapped_type());
+            return inserted->second;
+        }
+
+        UpgradeToMap();
+    }
+
+    return Map_[std::forward<TKeyParam>(key)];
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+template <class TSelf>
+auto TCompactMap<TKey, TValue, N, TCompare, TAllocator>::BeginImpl(TSelf* self)
+    -> std::conditional_t<std::is_const_v<TSelf>, const_iterator, iterator>
+{
+    using TResult = std::conditional_t<std::is_const_v<TSelf>, const_iterator, iterator>;
+    if (self->IsSmall()) {
+        return TResult(self->Vector_.begin());
+    }
+    return TResult(self->Map_.begin());
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+template <class TSelf>
+auto TCompactMap<TKey, TValue, N, TCompare, TAllocator>::EndImpl(TSelf* self)
+    -> std::conditional_t<std::is_const_v<TSelf>, const_iterator, iterator>
+{
+    using TResult = std::conditional_t<std::is_const_v<TSelf>, const_iterator, iterator>;
+    if (self->IsSmall()) {
+        return TResult(self->Vector_.end());
+    }
+    return TResult(self->Map_.end());
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+template <class TSelf>
+auto TCompactMap<TKey, TValue, N, TCompare, TAllocator>::FindImpl(TSelf* self, const key_type& key)
+    -> std::conditional_t<std::is_const_v<TSelf>, const_iterator, iterator>
+{
+    using TResult = std::conditional_t<std::is_const_v<TSelf>, const_iterator, iterator>;
+    if (self->IsSmall()) {
+        auto [it, found] = self->VectorLowerBound(key);
+        return found ? TResult(it) : TResult(self->Vector_.end());
+    }
+    return TResult(self->Map_.find(key));
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+template <class TSelf>
+auto TCompactMap<TKey, TValue, N, TCompare, TAllocator>::AtImpl(TSelf* self, const key_type& key)
+    -> std::conditional_t<std::is_const_v<TSelf>, const mapped_type&, mapped_type&>
+{
+    auto it = FindImpl(self, key);
+    if (it == EndImpl(self)) {
+        throw std::out_of_range("TCompactMap::at");
+    }
+    return it->second;
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+typename TCompactMap<TKey, TValue, N, TCompare, TAllocator>::size_type
+TCompactMap<TKey, TValue, N, TCompare, TAllocator>::count(const key_type& key) const
+{
+    return FindImpl(this, key) == EndImpl(this) ? 0 : 1;
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+bool TCompactMap<TKey, TValue, N, TCompare, TAllocator>::contains(const key_type& key) const
+{
+    return FindImpl(this, key) != EndImpl(this);
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+auto TCompactMap<TKey, TValue, N, TCompare, TAllocator>::find(const key_type& key) -> iterator
+{
+    return FindImpl(this, key);
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+auto TCompactMap<TKey, TValue, N, TCompare, TAllocator>::find(const key_type& key) const -> const_iterator
+{
+    return FindImpl(this, key);
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+std::pair<typename TCompactMap<TKey, TValue, N, TCompare, TAllocator>::iterator, bool>
+TCompactMap<TKey, TValue, N, TCompare, TAllocator>::insert(const value_type& v)
+{
+    if (IsSmall()) {
+        auto [vectorIt, found] = VectorLowerBound(v.first);
+
+        if (found) {
+            return {iterator(vectorIt), false};
+        }
+
+        if (Vector_.size() < N) {
+            auto inserted = EmplaceSmall(vectorIt, v.first, v.second);
+            return {iterator(inserted), true};
+        }
+
+        UpgradeToMap();
+    }
+
+    auto [mapIt, inserted] = Map_.emplace(v.first, v.second);
+    return {iterator(mapIt), inserted};
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+template <class... TArgs>
+std::pair<typename TCompactMap<TKey, TValue, N, TCompare, TAllocator>::iterator, bool>
+TCompactMap<TKey, TValue, N, TCompare, TAllocator>::emplace(TArgs&&... args)
+{
+    if (IsSmall()) {
+        TVectorValue value(std::forward<TArgs>(args)...);
+        auto [vectorIt, found] = VectorLowerBound(value.first);
+
+        if (found) {
+            return {iterator(vectorIt), false};
+        }
+
+        if (Vector_.size() < N) {
+            auto inserted = EmplaceSmall(vectorIt, std::move(value));
+            return {iterator(inserted), true};
+        }
+
+        UpgradeToMap();
+
+        // We already constructed the value above, so we cannot forward args again below.
+        auto [mapIt, inserted] = Map_.emplace(std::move(value));
+        return {iterator(mapIt), inserted};
+    }
+
+    auto [mapIt, inserted] = Map_.emplace(std::forward<TArgs>(args)...);
+    return {iterator(mapIt), inserted};
+}
+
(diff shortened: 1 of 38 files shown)


yt/yt/server/master/chunk_server/chunk_replica-inl.h
```diff
@@ -26,135 +26,192 @@ namespace NYT::NChunkServer {
...
+        YT_ASSERT(secondIndex >= 0 && secondIndex <= 0xff);
+    }
+
+    if constexpr (ExtraIndexCount >= 1) {
+        // If extra index is requested, it is always the second index.
+        ExtraIndex1_ = secondIndex;
+    }
+
+    // Should be true for all cases in practice, so doesn't hurt to check.
+    YT_ASSERT((reinterpret_cast<uintptr_t>(ptr) & 0xffff000000000003LL) == 0);
```
achulkov2, line 135: Changed `0xffff000000000000LL` -> `0xffff000000000003LL` here, looks like it was wrong originally.

yt/yt/tests/integration/master/test_media.py
```diff
@@ -439,9 +500,8 @@ def test_table_medium(self):
             assert replica.attributes["medium"] == self.NON_DEFAULT_MEDIUM
 
     @authors("babenko", "shakurov")
-    @pytest.mark.ignore_in_opensource_ci
```
achulkov2, line 440: There are a few other places like this, and it is best removed in a separate PR, but I want to get as many tests running in this one as possible, so will do it here for test_media.py.

achulkov2 (commented):
I have tried running a few test files locally, including `master/test_media.py`, `master/test_node_tracker.py`, `master/test_chunk_locations.py`, `master/test_chunk_server.py` and `node/test_locations.py`, and they seem to pass, but tests on the whole code base will be crucial.

I am also not entirely sure we got rid of all the potential O(MaxMediumCount^2) places, but I feel like this is good enough for now. If anyone remembers others, please mention them and we can evaluate their importance.

yt/yt/client/chunk_client/public.h
```diff
@@ -118,12 +118,29 @@ constexpr int MaxReplicationFactor = 20;
...
+//! NB: The bound on medium indexes for actual master media
+//! objects is *not* equal to this value. Be very careful.
+constexpr int MaxMediumCount = 64000;
+//! The average reasonable number of media on a cluster.
+//! Used for initializing compact containers.
+constexpr int TypicalMediumCount = 4;
+//! The actual upper bound on medium indexes of real media, accounting for
+//! historical sentinels |GenericMediumIndex| and |AllMediaIndex|.
+//! NB: This value is *not* equal to |MaxMediumCount|. Be very careful.
+constexpr int RealMediumIndexBound = MaxMediumCount + 2;
```
raundn, line 154: Right now, it a takes a good deal of head-scratching and parsing these constants to understand how all this stuff is arranged. A simple ascii-art diagram would've helped a great deal.

For example, this is what tripped me off. First of all, it's not clear what's meant by a "real medium". Reading this, I have to assume it stands for "non-sentinel medium" but that's never stated explicitly.  Second of all, reading the declaration of `RealMediumIndexBound`, it's reasonable to assume that real media (whatever they are) fall into range [0, RealMediumIndexBound], but then what about the fact that `GenericMediumIndex` is 126, and `AllMediaIndex` is 127?

I'm being intentionally nitpicky here. I get that the sentinels got stuck in the middle of the real media index range for compatibility reasons but let's make sure everyone else has a chance to get that, too.

yt/yt/core/misc/mex_set.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+namespace NYT {
+
+////////////////////////////////////////////////////////////////////////////////
+
+namespace {
+
+////////////////////////////////////////////////////////////////////////////////
+
+// This could have been a privavte class helper, but we are fooled by const-ness issues.
```
raundn: priva~~v~~te

yt/yt/core/misc/mex_set.h
```diff
@@ -0,0 +1,52 @@
...
+#include "public.h"
+
+namespace NYT {
+
+////////////////////////////////////////////////////////////////////////////////
+
+//! A simple set of integers supporting fast MEX (minimum excluded value) queries.
+//! Implemented as a set of disjoint intervals. See complexities of operations below.
+//! Feel free to extend however needed, or to replace with a more efficient implementation,
+//! there are about a million ways to do MEX with different tradeoffs. This one make sense
```
raundn: makes

yt/yt/core/misc/mex_set.h
```diff
@@ -0,0 +1,52 @@
...
+//! if the set is densely and consecutively populated, e.g. it is often just a few intervals.
+class TMexSet
```
raundn: Unless you want to turn it into a template, I suggest calling this `TMexIntSet`.

yt/yt/core/misc/mex_set.h
```diff
@@ -0,0 +1,52 @@
...
+
+    //! Returns whether the set contains the value.
+    //! Complexity: O(log N), where N is the number of disjoint intervals in the set.
+    bool Contains(int value) const;
+
+    //! Clears the set and resets MEX to 0.
+    //! Complexity: O(N), where N is the number of disjoint intervals in the set.
+    void Clear();
+
+    //! Returns the minimum non-negative integer not contained in the set.
```
raundn, line 40: It's a bit weird that non-negativity is first mentioned this late into the class definition.

yt/yt/core/misc/mex_set.cpp
raundn: How about turning this into static private helper method then?

yt/yt/core/misc/mex_set.cpp
```diff
@@ -0,0 +1,136 @@
...
+    if (value == Mex_) {
+        if (rightAdjacent) {
+            Mex_ = rightIt->second;
+        } else {
+            ++Mex_;
+        }
+    }
+
+    if (leftAdjacent && rightAdjacent) {
+        // Merge both intervals.
```
raundn: "Merge both" sounds weird, as if it's possible to merge only one interval =)

yt/yt/core/misc/mex_set.cpp
```diff
@@ -0,0 +1,136 @@
...
+        leftIt->second = rightIt->second;
+        Intervals_.erase(rightIt);
+    } else if (leftAdjacent) {
+        // Extend left interval to the right.
+        leftIt->second += 1;
+    } else if (rightAdjacent) {
+        // Extend right interval to the left.
+        auto newRight = rightIt->second;
+        Intervals_.erase(rightIt);
+        Intervals_[value] = newRight;
```
raundn: emplace?

yt/yt/core/misc/mex_set.cpp
```diff
@@ -0,0 +1,136 @@
...
+    } else {
+        // Create a new interval.
+        Intervals_[value] = value + 1;
```
raundn: emplace?

yt/yt/core/misc/mex_set.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+int TMexSet::GetMex() const
+{
+    return Mex_;
+}
+
+void TMexSet::ValidateBounds(int value)
+{
+    YT_VERIFY(value >= 0);
+    YT_VERIFY(value < std::numeric_limits<int>::max());
```
raundn: I think it's better to throw here. I don't believe it's reasonable to count on the client code to guarantee not to pass INT_MAX.

yt/yt/server/lib/misc/max_min_balancer-inl.h
```diff
@@ -28,80 +20,105 @@ TDecayingMaxMinBalancer<T, W>::TDecayingMaxMinBalancer(
...
+    auto it = ContenderToWeight_.find(winner);
+    YT_VERIFY(it != ContenderToWeight_.end());
 
-    it->Weight += extraWeight;
+    it->second += extraWeight;
 }
 
 template <typename T, typename W>
-void TDecayingMaxMinBalancer<T, W>::ResetWeights()
+void TDecayingMaxMinBalancer<T, W>::AddWeightWithDefault(T winner, W extraWeight, W defaultWeight)
```
raundn: Is there a good reason for this to be the fused method this is? Wouldn't `TryAddContender` + `AddWeight` have sufficed?

yt/yt/server/master/chunk_server/chunk_manager.cpp
```diff
@@ -2148,6 +2179,18 @@ class TChunkManager
             THROW_ERROR_EXCEPTION("Medium count limit %v is reached",
                 MaxMediumCount);
         }
+
+        if (hintIndex && !IsValidRealMediumIndex(*hintIndex)) {
+            THROW_ERROR_EXCEPTION("Requested medium index %v is not allowed for actual media objects",
```
raundn: "actual" or "real"?

yt/yt/server/master/chunk_server/chunk_manager.cpp
```diff
@@ -6452,12 +6493,12 @@ class TChunkManager
 
     int GetFreeMediumIndex()
     {
-        for (int index = 0; index < MaxMediumCount; ++index) {
-            if (!UsedMediumIndexes_[index]) {
-                return index;
-            }
-        }
-        YT_ABORT();
+        YT_VERIFY(IndexToMediumMap_.size() < MaxMediumCount);
```
raundn, line 6409: I haven't read ahead yet but I assume there's protective code out there somewhere that won't allow the user to create too many media. If not, we should add it.

yt/yt/client/chunk_client/public.h
```diff
@@ -118,12 +118,29 @@ constexpr int MaxReplicationFactor = 20;
...
+//! Returns a list of all sentinel medium indexes: |GenericMediumIndex|, |AllMediaIndex| and |[RealMediumIndexBound, MediumIndexBound)|.
+std::vector<int> GetSentinelMediumIndexes();
+//! Returns true if |mediumIndex| is a valid medium index: in range |[0, RealMediumIndexBound)|
+//! and not equal to any of the reserved sentinel values (see |IsSentinelMediumIndex|).
+bool IsValidRealMediumIndex(int mediumIndex);
 
 template <typename T>
 using TMediumMap = THashMap<int, T>;
-template <typename T>
-using TCompactMediumMap = TCompactFlatMap<int, T, 4>;
```
raundn, line 130: Why the deletion? This compact map avoids *a lot* of allocations, as it gets passed around quite a bit inside the replicator. I'm afraid getting rid of it will be a *hard* performance hit.

And in any case, `TCompactMediumMap`s are often use to store per-medium info *for a particular chunk*. And an individual chunk rarely is (or going to be) placed at more that 4 media simultaneously. Hence the memory allocation savings.

yt/yt/server/master/chunk_server/chunk_replicator.cpp
```diff
@@ -561,19 +555,19 @@ TChunkReplicator::TChunkStatistics TChunkReplicator::ComputeErasureChunkStatisti
 
     auto* codec = NErasure::GetCodec(chunk->GetErasureCodec());
 
-    TCompactMediumMap<std::array<TChunkLocationList, ChunkReplicaIndexBound>> decommissionedReplicas;
-    TCompactMediumMap<std::array<ui8, RackIndexBound>> perRackReplicaCounters;
+    TMediumMap<std::array<TChunkLocationList, ChunkReplicaIndexBound>> decommissionedReplicas;
+    TMediumMap<std::array<ui8, RackIndexBound>> perRackReplicaCounters;
     // TODO(gritukan): YT-16557.
-    TCompactMediumMap<THashMap<const TDataCenter*, ui8>> perDataCenterReplicaCounters;
```
raundn: This is a very hot code, dealing with an individual chunk. Which means these maps will almost always hold just a few items. The removal of `TCompactMediumMap` is hard to argue for.

yt/yt/server/master/chunk_server/chunk_replicator.cpp
```diff
@@ -901,15 +895,11 @@ void TChunkReplicator::ComputeErasureChunkStatisticsCrossMedia(
...
-
-    std::bitset<MaxMediumCount> transientMedia;
-    if (allMediaTransient) {
-        transientMedia.flip();
-    } else {
-        for (const auto& mediumIdAndPtrPair : Bootstrap_->GetChunkManager()->Media()) {
-            auto* medium = mediumIdAndPtrPair.second;
-            if (medium->IsDomestic()) {
-                transientMedia.set(medium->GetIndex(), medium->AsDomestic()->GetTransient());
+    TMediumSet transientMedia;
```
raundn, line 938: This set, as well, will only ever hold just a couple of items. Using a compact container is advised to avoid allocations.

yt/yt/server/master/chunk_server/chunk_replicator.cpp
```diff
@@ -1581,13 +1568,15 @@ EMisscheduleReason TChunkReplicator::TryScheduleReplicationJob(
...
+        "(JobId: %v, JobEpoch: %v, Address: %v, ChunkId: %v, TargetAddresses: %v, IsPullReplicationJob: %v, TargetMediumIndex: %v, TargetMediumName: %v)",
         job->GetJobId(),
         job->GetJobEpoch(),
         sourceNode->GetDefaultAddress(),
         chunkWithIndex,
         MakeFormattableView(targetNodes, TNodePtrAddressFormatter()),
-        isPullReplicationJob);
+        isPullReplicationJob,
+        targetMediumIndex,
+        targetMedium->GetName());
```
raundn, line 1641: The is disputable, but I think logging just the medium name *or* its index is enough. (As a rule, I prefer names to indices unless logging the name is hard to do for some technical reason or other.)

raundn (changes requested):

yt/yt/core/misc/mex_set.h
achulkov2: Added it into a few comments.

yt/yt/core/misc/mex_set.cpp
achulkov2: Sure, but is there actually a difference?

yt/yt/server/master/chunk_server/chunk_manager.cpp
achulkov2: For some reason I thought that would be easier to understand for the user, but let's keep the terminology consistent.

yt/yt/server/master/chunk_server/chunk_manager.cpp
achulkov2: Yeah:
```
if (MediumMap_.GetSize() >= MaxMediumCount) {
    THROW_ERROR_EXCEPTION("Medium count limit %v is reached",
        MaxMediumCount);
}
```

yt/yt/server/master/chunk_server/chunk_replicator.cpp
achulkov2: Ok.

yt/yt/client/chunk_client/public.h
achulkov2: As discussed during a meeting, the main reason was that compact flat map has the following caveat:
```Unlike SmallSet, however, this one never falls back to std::map (or
 *  set) for larger sizes. This means that the flat map is only useful
 *    - at small sizes, when there's absolutely no chance of it getting big, or
 *    - when it's filled once and is then only read from.```
 
I will refactor to using `TCompactSet`, which does fall back to std::set for a large number of elements.

File: yt/yt/core/misc/mex_set.cpp (hunk 11 above)
Comment (achulkov2) [reply yt/yt/core/misc/mex_set.cpp]: Will throw on any value out of bounds here.

File: yt/yt/server/master/chunk_server/chunk_replicator.cpp (hunk 17 above)
Comment (achulkov2) [reply yt/yt/server/master/chunk_server/chunk_replicator.cpp:938]: `TMediumSet` is already `TCompactSet<int, NChunkClient::TypicalMediumCount>`, so nothing to do here.

File: yt/yt/client/chunk_client/public.h (hunk 3 above)
Comment (achulkov2) [reply yt/yt/client/chunk_client/public.h:154]: Did my best :)

File: yt/yt/client/chunk_client/public.h (hunk 15 above)
Comment (achulkov2) [reply yt/yt/client/chunk_client/public.h:130]: So, this turned out much harder than expected. 

I ended up introducing a new class, `TCompactMap` (not to be confused with `TCompactFlatMap`!) (all jokes aside, there is a comment about the difference), which is also a map over `TCompactVector`, but unlike `TCompactFlatMap` it does fallback to `std::map` when the number of elements grows larger than the amount stored inline.

Sounds easy, but there is the annoying caveat that `TCompactVector` can only store copy-assignable types, so we cannot store `std::pair<const TKey, TValue>` there, which is the obvious `value_type` of `std::map`, only `std::pair<TKey, TValue>`. 
This annoying difference makes it very hard to provide a proper single iterator over both containers. I spent some time battling C++ bosses to try to make it standard compliant via various proxy types, but it adds a huge amount of extra complexity, so I gave up and used `reinterpret_cast`.
Technically, it is a strict aliasing violation, so I used `__attribute__((__may_alias__))`, which we already have in the form of the `alias_hack` macro, as a form of sacrifice to the aforementioned gods.
Hopefully, they will bless us with correct binary code!

From my research, I expect this scenario combined with the compiler instruction to be good enough to make it safe.

Comment (achulkov2) [reply yt/yt/client/chunk_client/public.h:130]: To be discreet: a lot of the boilerplate code/unit tests for the new container was written with the help of (you would never guess!) AI.

File: yt/yt/server/master/chunk_server/chunk_replicator.cpp (hunk 16 above)
Comment (achulkov2) [reply yt/yt/server/master/chunk_server/chunk_replicator.cpp]: Also removed `TCompactVector -> std::vector` in `mediaOnWhichLost` in favour of `TMediumSet = TCompactSet`.

File: library/cpp/yt/compact_containers/compact_map.h [hunk 19]
```diff
@@ -0,0 +1,169 @@
...
+    size_t N,
+    class TCompare = std::less<TKey>,
+    class TAllocator = std::allocator<std::pair<const TKey, TValue>>>
+class TCompactMap
+{
+public:
+    using key_type = TKey;
+    using mapped_type = TValue;
+    // We store std::pair<TKey, TValue> in TCompactVector because of its copy-assignability requirement.
+    // The underlying storage type of std::map is, obviosly, std::pair<const TKey, TValue>.
```
raundn: obvioUsly

library/cpp/yt/compact_containers/compact_map.h
```diff
@@ -0,0 +1,169 @@
...
+    // To provide a single iterator interface, we use reinterpret_cast to convert between pointers
+    // to values of these types, which should have the same layout on all sane compilers.
+    // However, this is still technically undefined behavior and a strict aliasing violation.
```
raundn: Yikes. I'm sorry, but I don't think this is ok. I'd suggest giving up compact vector altogether and using std::array or something like that. We don't need compact vector's ability to reallocate to the heap here, anyway.

yt/yt/client/chunk_client/public.cpp
```diff
@@ -20,4 +20,26 @@ const std::string DefaultSlotsMediumName("default");
 
 ////////////////////////////////////////////////////////////////////////////////
 
+static_assert(TypicalMediumCount <= MaxMediumCount, "Typical medium count exceeds max medium count");
+static_assert(MaxMediumCount <= RealMediumIndexBound, "Max medium count exceeds bound on real medium indexes");
+static_assert(RealMediumIndexBound <= MediumIndexBound, "Real medium index bound exceeds medium index bound");
+
+std::vector<int> GetSentinelMediumIndexes()
```
raundn: I suspect this can (and therefore should) be made constexpr.

yt/yt/client/chunk_client/public.h
```diff
@@ -120,12 +121,50 @@ constexpr int MaxReplicationFactor = 20;
...
+/*
+ *   Non-sentinel media slots (MaxMediumCount allowed entries)    Sentinel-only slots
+ *                        ∨ ∨ ∨                                         ∨ ∨ ∨
+ *                              GenericMediumIndex=126                                   MediumIndexBound
+ *   0                          ∨                                                        ∨
+ *   |--------------------------XX---------------------------|---------------------------|
+ *                               ^                           ^
+ *                               AllMediaIndex=127           RealMediumIndexBound
+ *                              ^^
+ *                              Legacy sentinels embedded into the real range for compatibility
```
raundn, line 136: Beautiful. Unironically, beautiful.

yt/yt/client/chunk_client/public.h
```diff
@@ -120,12 +121,50 @@ constexpr int MaxReplicationFactor = 20;
...
+ *
+ * “Real” media correspond to actual master medium objects. Sentinels are special indexes
```
raundn: Non-ascii quotes.

yt/yt/client/chunk_client/public.h
```diff
@@ -120,12 +121,50 @@ constexpr int MaxReplicationFactor = 20;
...
+ * that do not have associated medium objects and are used for special purposes across APIs.
+ * Because GenericMediumIndex/AllMediaIndex sit inside the real range, |RealMediumIndexBound|
+ * exceeds the plain count limit by 2, to keep room for |MaxMediumCount| real media.
```
raundn, line 141: It's probably a good idea to add that "Because GenericMediumIndex/AllMediaIndex sit inside ... one must avoid using comparison for checking the realness of a medium and must use IsValidRealMediumIndex instead". Or something along those lines.

yt/yt/server/lib/misc/max_min_balancer-inl.h
```diff
@@ -28,82 +20,99 @@ TDecayingMaxMinBalancer<T, W>::TDecayingMaxMinBalancer(
...
-        Contenders_.end(),
-        [&] (const TWeighedContender& w) { return pred(w.Contender); });
+        ContenderToWeight_.begin(),
+        ContenderToWeight_.end(),
+        [&] (const auto& w) { return pred(w.first); });
 
-    for (auto it = resultIt; it != Contenders_.end(); ++it) {
-        if (pred(it->Contender) && it->Weight < resultIt->Weight) {
+    for (auto it = resultIt; it != ContenderToWeight_.end(); ++it) {
+        if (pred(it->first) && it->second < resultIt->second) {
```
raundn: I think these days this is better expressed via `views::filter` but I'm not going to insist.

yt/yt/server/master/chunk_server/chunk_manager.cpp
```diff
@@ -210,6 +212,35 @@ struct TChunkToLinkedListNode
 
 ////////////////////////////////////////////////////////////////////////////////
 
+//! Initializes itself with all sentinels. Required because history has left us with
+//! sentinels inside the valid medium index range.
+class TUsedMediumIndexSet
```
raundn: Let's move this to a separate file. Yes, the class is small but it may turn out useful elsewhere in the code.

raundn (changes requested):

yt/yt/client/chunk_client/public.h
achulkov2: Added the last part as a continuation sentence.

yt/yt/server/lib/misc/max_min_balancer-inl.h
achulkov2: Old code, but improved this part.

yt/yt/server/master/chunk_server/chunk_manager.cpp
achulkov2: Ok. Put it in `client/chunk_client` so that it is near the definitions themselves.

library/cpp/yt/compact_containers/compact_map.h
```diff
@@ -0,0 +1,187 @@
...
+        std::is_nothrow_move_constructible_v<key_compare> &&
+        std::allocator_traits<allocator_type>::is_always_equal::value);
+    ~TCompactMap();
+
+    template <class TIt>
+    TCompactMap(TIt first, TIt last);
+
+    TCompactMap(std::initializer_list<value_type> init);
+
+    bool empty() const;
```
raundn, line 59: [[nodiscard]]

library/cpp/yt/compact_containers/compact_map-inl.h
```diff
@@ -0,0 +1,826 @@
...
+    using TArrayValue = typename TCompactMap::TArrayValue;
+    using TSmallValue = std::conditional_t<IsConst, const TArrayValue, TArrayValue>;
+    using TSmallPointer = std::conditional_t<IsConst, const TArrayValue*, TArrayValue*>;
+    using TReference = std::conditional_t<IsConst, const TValueType&, TValueType&>;
+    using TPointer = std::conditional_t<IsConst, const TValueType*, TValueType*>;
+
+    union
+    {
+        TVecIter VIter;
+        TMapIter MIter;
```
raundn, line 33: `VIter_`, `MIter_`

library/cpp/yt/compact_containers/compact_map-inl.h
```diff
@@ -0,0 +1,826 @@
...
+    };
+    bool Small_;
+
+    explicit TIteratorBase(TVecIter it)
+        : VIter(it)
```
raundn, line 38: Not that it matters but there's no reason no to `std::move(it)`.

library/cpp/yt/compact_containers/compact_map-inl.h
```diff
@@ -0,0 +1,826 @@
...
+        , Small_(true)
+    { }
+
+    explicit TIteratorBase(TMapIter it)
+        : MIter(it)
```
raundn, line 43: Same here.

library/cpp/yt/compact_containers/compact_map-inl.h
```diff
@@ -0,0 +1,826 @@
...
+    using reference = TReference;
+    using pointer = TPointer;
+    using iterator_category = std::bidirectional_iterator_tag;
+
+    reference operator*() const
+    {
+        if (Small_) {
+            return *VIter;
+        }
+        return *MIter;
```
raundn, line 71: This looks like a perfect place for a ternary.

library/cpp/yt/compact_containers/compact_map-inl.h
```diff
@@ -0,0 +1,826 @@
...
+    }
+
+    pointer operator->() const
+    {
+        if (Small_) {
+            return &*VIter;
+        }
+        return &*MIter;
```
raundn, line 79: Same here.

library/cpp/yt/compact_containers/compact_map-inl.h
```diff
@@ -0,0 +1,826 @@
...
+
+    bool operator==(const TIteratorBase& other) const
+    {
+        if (Small_ != other.Small_) {
+            return false;
+        }
+        return Small_ ? VIter == other.VIter : MIter == other.MIter;
+    }
+
+    bool operator!=(const TIteratorBase& other) const
```
raundn, line 124: Fun fact: since C++ 20, there's actually no need to define `operator!=` when `operator==` is defined: the compiler will rewrite any occurrences of `a != b` with  `!(a == b)`.

The only reason to define `operator!=` is if there's a chance it'll be needed as an actual function (e.g. we'll want to take its address). But even then `bool operator!=(const TIteratorBase& other) const = default;` will be enough here.

library/cpp/yt/compact_containers/compact_map-inl.h
```diff
@@ -0,0 +1,826 @@
...
+public:
+    iterator() = default;
+
+private:
+    friend class TCompactMap<TKey, TValue, N, TCompare, TAllocator>;
+    friend class const_iterator;
+    using TBase = typename TCompactMap<TKey, TValue, N, TCompare, TAllocator>::template TIteratorBase<false>;
+
+    explicit iterator(TArrayIterator it)
+        : TBase(it)
```
raundn, line 143: No reason no to `std::move(it)` here.

library/cpp/yt/compact_containers/compact_map.h
```diff
@@ -0,0 +1,187 @@
...
+
+    mapped_type& at(const key_type& key);
+    const mapped_type& at(const key_type& key) const;
+
+private:
+    template <bool IsConst>
+    class TIteratorBase;
+
+    using TArrayValue = value_type;
+    using TStorage = std::aligned_storage_t<sizeof(TArrayValue), alignof(TArrayValue)>;
```
raundn, line 112: Another fun fact: `aligned_storage_t` is deprecated in C++23. Let's do without.

library/cpp/yt/compact_containers/compact_map-inl.h
```diff
@@ -0,0 +1,826 @@
...
+{
+    return ArrayData() + ArraySize_;
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+void TCompactMap<TKey, TValue, N, TCompare, TAllocator>::ClearArray()
+{
+    auto* data = ArrayData();
+    for (size_type index = 0; index < ArraySize_; ++index) {
+        std::destroy_at(data + index);
```
raundn, line 221: I hate to ask this, but... what if this throws? We can't really demand `is_nothrow_destructible` ('cause nobody bothers to actually write all those `noexcept`s), but if an exception actually happens here, we're screwed. All in all, I think `ClearArray` should be made noexcept (and thus crash if anything goes awry). It's still better than to proceed with a half-destroyed container. But maybe you'll come up with a better idea.

library/cpp/yt/compact_containers/compact_map-inl.h
```diff
@@ -0,0 +1,826 @@
...
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+void TCompactMap<TKey, TValue, N, TCompare, TAllocator>::MoveArrayFrom(TCompactMap&& other)
+{
+    auto* data = ArrayData();
+    auto* otherData = other.ArrayData();
+
+    for (size_type index = 0; index < other.ArraySize_; ++index) {
+        std::construct_at(data + index, std::move(otherData[index]));
```
raundn, line 245: Static-asserting all those `is_nothrow_move_constructible_v` etc. here would be a great boon for readability. Otherwise one is left to wonder about the consequences of an exception in the middle of this loop. 

(Another idea is to replicate those conditional `noexcept`s on this method.)

library/cpp/yt/compact_containers/compact_map-inl.h
```diff
@@ -0,0 +1,826 @@
...
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+void TCompactMap<TKey, TValue, N, TCompare, TAllocator>::CopyArrayFrom(const TCompactMap& other)
+{
+    auto* data = ArrayData();
+    auto* otherData = other.ArrayData();
+
+    for (size_type index = 0; index < other.ArraySize_; ++index) {
+        std::construct_at(data + index, otherData[index]);
```
raundn, line 233: Unlike `MoveArrayFrom`, this *can* be rolled back - especially if items are `is_nothrow_destructible` or if we decide to make `ClearArray` `noexcept`.

 (Obviously, when this is called from constructor, there's no need to rollback anything. Assignment, however, is another matter.)

library/cpp/yt/compact_containers/compact_map.h
```diff
@@ -0,0 +1,187 @@
...
+        std::allocator_traits<allocator_type>::is_always_equal::value);
+    ~TCompactMap();
+
+    template <class TIt>
+    TCompactMap(TIt first, TIt last);
+
+    TCompactMap(std::initializer_list<value_type> init);
+
+    bool empty() const;
+    size_type size() const;
```
raundn, line 60: Let's support `ssize` as well.

library/cpp/yt/compact_containers/compact_map-inl.h
```diff
@@ -0,0 +1,826 @@
...
+bool TCompactMap<TKey, TValue, N, TCompare, TAllocator>::IsSmall() const
+{
+    return Map_.empty();
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+void TCompactMap<TKey, TValue, N, TCompare, TAllocator>::UpgradeToMap()
+{
+    if (!IsSmall()) {
+        return;
```
raundn, line 421: I believe this check is redundant. (If it's not, I'd suggest at least renaming `UpgradeToMap` to `EnsureUpgradedToMap`/`EnsureLarge` or something along those lines.)

library/cpp/yt/compact_containers/compact_map-inl.h
```diff
@@ -0,0 +1,826 @@
...
+        std::construct_at(data + ArraySize_, std::forward<TArgs>(args)...);
+        ++ArraySize_;
+        return data + index;
+    }
+
+    // Shift elements to make space at index.
+    std::construct_at(data + ArraySize_, std::move(data[ArraySize_ - 1]));
+    for (size_type i = ArraySize_ - 1; i > index; --i) {
+        std::destroy_at(data + i);
+        std::construct_at(data + i, std::move(data[i - 1]));
```
raundn, line 466: This is cursed from the exception safety standpoint, and something probably should be done about that. (I don't think every call to `EmplaceSmall` comes from a `noexcept(is_nothrow_move_constructible_v ...)` method at the moment.)

library/cpp/yt/compact_containers/compact_map.h
```diff
@@ -0,0 +1,187 @@
...
+
+    TInlineStorage ArrayStorage_;
+    size_type ArraySize_ = 0;
+    TMap Map_;
+    key_compare Compare_{};
+    allocator_type Alloc_{};
+
+    TArrayIterator ArrayData();
+    TArrayConstIterator ArrayData() const;
+    TArrayIterator ArrayBegin();
```
raundn, line 128: I think `ArrayData` should return `TArrayValue*` instead of `TArrayIterator`. I understand that these are the same type but, semantically, there's a difference. In fact it's *the* difference between `ArrayData` and `ArrayBegin`.

library/cpp/yt/compact_containers/compact_map.h
```diff
@@ -0,0 +1,187 @@
...
+    TArrayIterator ArrayEnd();
+    TArrayConstIterator ArrayEnd() const;
+    void ClearArray();
+    void CopyArrayFrom(const TCompactMap& other);
+    void MoveArrayFrom(TCompactMap&& other);
+
+    // Returns if map part is empty (i.e. we are still using inline array).
+    bool IsSmall() const;
+    // Moves data from inline array to map.
+    void UpgradeToMap();
```
raundn, line 139: I haven't read the code ahead yet, but... do we ever downgrade to small? If not, that should probably be mentioned in the commentary above.

library/cpp/yt/compact_containers/compact_map-inl.h
```diff
@@ -0,0 +1,826 @@
...
+template <bool IsConst>
+class TCompactMap<TKey, TValue, N, TCompare, TAllocator>::TIteratorBase
+{
+private:
+    friend class TCompactMap<TKey, TValue, N, TCompare, TAllocator>;
+    friend class TIteratorBase<!IsConst>;
+    friend class iterator;
+    friend class const_iterator;
+
+    using TVecIter = std::conditional_t<IsConst, TArrayConstIterator, TArrayIterator>;
```
raundn, line 21: Not a big deal, but why is it called `TVecIter`, not `TArrayIter`? Same question relating `VIter` below. This is bit confusing when there're `TArrayValue`s all around.

library/cpp/yt/compact_containers/compact_map-inl.h
```diff
@@ -0,0 +1,826 @@
...
+        , Small_(true)
+    { }
+
+    explicit TIteratorBase(TMapIter it)
+        : MIter(it)
+        , Small_(false)
+    { }
+
+    // Conversion constructor for const_iterator from iterator.
+    template <bool WasConst, typename = std::enable_if_t<IsConst && !WasConst>>
```
raundn, line 48: I this this is better expressed via `requires` these days.

library/cpp/yt/compact_containers/compact_map-inl.h
```diff
@@ -0,0 +1,826 @@
...
+{
+    static_assert(sizeof(TStorage) == sizeof(TArrayValue), "Inline storage size mismatch");
+    static_assert(alignof(TStorage) == alignof(TArrayValue), "Inline storage alignment mismatch");
+    return reinterpret_cast<TArrayIterator>(ArrayStorage_.data());
+}
+
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+auto TCompactMap<TKey, TValue, N, TCompare, TAllocator>::ArrayData() const -> TArrayConstIterator
+{
+    return reinterpret_cast<TArrayConstIterator>(ArrayStorage_.data());
```
raundn, line 189: These casts would (will?) go away if we get rid of the aligned storage.

library/cpp/yt/compact_containers/compact_map-inl.h
```diff
@@ -0,0 +1,826 @@
...
+    if (IsSmall()) {
+        auto [arrayIt, found] = ArrayLowerBound(key);
+
+        if (found) {
+            return arrayIt->second;
+        }
+
+        if (ArraySize_ < N) {
+            auto inserted = EmplaceSmall(arrayIt, std::forward<TKeyParam>(key), mapped_type());
+            return inserted->second;
```
raundn, line 541: Let's call this `insertResult` (or `emplaceResult`) and reserve `inserted` specifically for bools indicating whether an insertion has actually happened or not.

library/cpp/yt/compact_containers/compact_map-inl.h
```diff
@@ -0,0 +1,826 @@
...
+template <class TKey, class TValue, size_t N, class TCompare, class TAllocator>
+template <class TSelf>
+auto TCompactMap<TKey, TValue, N, TCompare, TAllocator>::EndImpl(TSelf* self)
+    -> std::conditional_t<std::is_const_v<TSelf>, const_iterator, iterator>
+{
+    using TResult = std::conditional_t<std::is_const_v<TSelf>, const_iterator, iterator>;
+    if (self->IsSmall()) {
+        return TResult(self->ArrayEnd());
+    }
+    return TResult(self->Map_.end());
```
raundn, line 571: By the way, are these two "end" iterators (one wrapping array's end and the other one wrapping map's end) considered equal? I'm pretty sure they aren't. 

One the one hand, this is completely reasonable (after all, one has to modify `TCompactMap` to get to differing "ends"). But on the other hand, this may be quite surprising. I mean, a mental model for this class is "array + std::map", and neither implies the need to ever invalidate `end()`.

I think this should be either fixed or documented explicitly in the commentary before the class.

yt/yt/server/master/chunk_server/chunk_replica.h
```diff
@@ -121,35 +137,40 @@ template <class T>
 using TPtrWithMediumIndex = TAugmentedPtr<
     T,
     /*WithReplicaState*/ false,
-    /*IndexCount*/ 1,
+    /*CompactIndexCount*/ 0,
+    /*ExtraIndexCount*/ 1,
```
raundn, line 141: I think it should be explained somewhere here that extra index is wider than compact index. It should also be mentioned that compact indices are 8 bit wide regardless of `CompactIndexCount`.

Otherwise it's impossible to guess why we're using extra index when compact index is available.

yt/yt/server/master/chunk_server/chunk_replica.h
```diff
@@ -17,38 +17,46 @@ namespace NYT::NChunkServer {
 ////////////////////////////////////////////////////////////////////////////////
 
 //! A compact representation for |(T*, index1, index2, replica_state)|.
-template <class T, bool WithReplicaState, int IndexCount, template <class> class TAugmentationAccessor>
+//! Compact indices are stored in the upper 16 bits of the pointer value.
+//! Extra indices are stored as plain integer fields.
```
raundn, line 21: It should probably be mentioned that just a single extra index is supported at the moment.

PR #1426, opened 2025-09-07
Refactoring: apply patches from inner repo

cherepasshka:
* Changelog entry
Type: feature
Component: master
Refactoring around ReplicasData in chunk

### DIFF
--- a/yt/yt/server/master/chunk_server/chunk.cpp
+++ b/yt/yt/server/master/chunk_server/chunk.cpp
@@ -280,10 +280,16 @@ bool TChunk::HasParents() const
 }
 
 void TChunk::AddReplica(
-    TChunkLocationPtrWithReplicaInfo replica,
-    const TDomesticMedium* medium,
+    TStoredReplica replica,
+    const TMedium* medium,
     bool approved)
 {
+    if (medium->IsOffshore()) {
+        auto* data = MutableReplicasData();
+        data->AddStoredReplica(replica);
+        return;
+    }
+    auto* domesticMedium = medium->AsDomestic();
     auto* data = MutableReplicasData();
     if (IsJournal()) {
         for (auto& existingReplica : data->MutableStoredReplicas()) {
@@ -299,9 +305,9 @@ void TChunk::AddReplica(
     }
 
     data->AddStoredReplica(replica);
-    if (!medium->GetTransient()) {
+    if (!domesticMedium->GetTransient()) {
         auto lastSeenReplicas = data->MutableLastSeenReplicas();
-        auto nodeId = replica.GetPtr()->GetNode()->GetId();
+        auto nodeId = replica.GetNodeId();
         if (IsErasure()) {
             lastSeenReplicas[replica.GetReplicaIndex()] = nodeId;
         } else {
@@ -332,7 +338,10 @@ void TChunk::RemoveReplica(
     auto storedReplicas = data->GetStoredReplicas();
     for (int replicaIndex = 0; replicaIndex < std::ssize(storedReplicas); ++replicaIndex) {
         auto existingReplica = storedReplicas[replicaIndex];
-        if (existingReplica.GetPtr() == replica.GetPtr() && existingReplica.GetReplicaIndex() == replica.GetReplicaIndex()) {
+        if (!existingReplica.IsChunkLocation()) {
+            continue;
+        }
+        if (existingReplica.AsChunkLocation().GetPtr() == replica.GetPtr() && existingReplica.GetReplicaIndex() == replica.GetReplicaIndex()) {
             data->RemoveStoredReplica(replicaIndex);
             return;
         }
@@ -340,7 +349,7 @@ void TChunk::RemoveReplica(
     YT_ABORT();
 }
 
-void TChunk::ApproveReplica(TChunkLocationPtrWithReplicaInfo replica)
+void TChunk::ApproveReplica(TStoredReplica replica)
 {
     auto* data = MutableReplicasData();
     ++data->ApprovedReplicaCount;
@@ -826,19 +835,19 @@ TChunk::TReplicasData<TypicalStoredReplicaCount, MaxLastSeenReplicaCount>::TRepl
 }
 
 template <size_t TypicalStoredReplicaCount, size_t MaxLastSeenReplicaCount>
-TRange<TChunkLocationPtrWithReplicaInfo> TChunk::TReplicasData<TypicalStoredReplicaCount, MaxLastSeenReplicaCount>::GetStoredReplicas() const
+TRange<TStoredReplica> TChunk::TReplicasData<TypicalStoredReplicaCount, MaxLastSeenReplicaCount>::GetStoredReplicas() const
 {
     return TRange(StoredReplicas);
 }
 
 template <size_t TypicalStoredReplicaCount, size_t MaxLastSeenReplicaCount>
-TMutableRange<TChunkLocationPtrWithReplicaInfo> TChunk::TReplicasData<TypicalStoredReplicaCount, MaxLastSeenReplicaCount>::MutableStoredReplicas()
+TMutableRange<TStoredReplica> TChunk::TReplicasData<TypicalStoredReplicaCount, MaxLastSeenReplicaCount>::MutableStoredReplicas()
 {
     return TMutableRange(StoredReplicas);
 }
 
 template <size_t TypicalStoredReplicaCount, size_t MaxLastSeenReplicaCount>
-void TChunk::TReplicasData<TypicalStoredReplicaCount, MaxLastSeenReplicaCount>::AddStoredReplica(TChunkLocationPtrWithReplicaInfo replica)
+void TChunk::TReplicasData<TypicalStoredReplicaCount, MaxLastSeenReplicaCount>::AddStoredReplica(TStoredReplica replica)
 {
     StoredReplicas.push_back(replica);
 }
--- a/yt/yt/server/master/chunk_server/chunk_manager.cpp
+++ b/yt/yt/server/master/chunk_server/chunk_manager.cpp
@@ -766,12 +766,10 @@ public:
             this);
     }
 
-
-
     TNodeList AllocateWriteTargets(
         TDomesticMedium* medium,
         TChunk* chunk,
-        const TChunkLocationPtrWithReplicaInfoList& replicas,
+        const TStoredReplicaList& replicas,
         int desiredCount,
         int minCount,
         std::optional<int> replicationFactorOverride,
@@ -795,7 +793,7 @@ public:
     TNodeList AllocateWriteTargets(
         TDomesticMedium* medium,
         TDummyNbdChunk* chunk,
-        const TChunkLocationPtrWithReplicaInfoList& replicas,
+        const TStoredReplicaList& replicas,
         int desiredCount,
         int minCount,
         std::optional<int> replicationFactorOverride,
@@ -819,7 +817,7 @@ public:
     TNodeList AllocateWriteTargets(
         TDomesticMedium* medium,
         TChunk* chunk,
-        const TChunkLocationPtrWithReplicaInfoList& replicas,
+        const TStoredReplicaList& replicas,
         int replicaIndex,
         int desiredCount,
         int minCount,
@@ -841,7 +839,7 @@ public:
     TNodeList AllocateWriteTargets(
         TDomesticMedium* medium,
         TDummyNbdChunk* chunk,
-        const TChunkLocationPtrWithReplicaInfoList& replicas,
+        const TStoredReplicaList& replicas,
         int replicaIndex,
         int desiredCount,
         int minCount,
@@ -1178,12 +1176,13 @@ public:
         YT_LOG_DEBUG(
             "Chunk created "
             "(ChunkId: %v, ChunkListId: %v, TransactionId: %v, Account: %v, Medium: %v, "
-            "ReplicationFactor: %v, ErasureCodec: %v, Movable: %v, Vital: %v%v%v)",
+            "RequisitionIndex: %v, ReplicationFactor: %v, ErasureCodec: %v, Movable: %v, Vital: %v%v%v)",
             chunk->GetId(),
             GetObjectId(chunkList),
             transaction->GetId(),
             account->GetName(),
             medium->GetName(),
+            requisitionIndex,
             replicationFactor,
             erasureCodecId,
             movable,
@@ -1299,8 +1298,12 @@ public:
         // Unregister chunk replicas from all known locations.
         // Schedule removal jobs.
         for (auto storedReplica : chunk->StoredReplicas()) {
+            if (!storedReplica.IsChunkLocation()) {
+                // TODO(cherepashka): handle this, when offshore mediums will be supported.
+                continue;
+            }
             // TODO(aleksandra-zh): skip Sequoia replicas here.
-            auto* location = storedReplica.GetPtr();
+            auto* location = storedReplica.AsChunkLocation().GetPtr();
             auto replicaIndex = storedReplica.GetReplicaIndex();
             TChunkPtrWithReplicaIndex replica(chunk, replicaIndex);
             if (!location->RemoveReplica(replica)) {
@@ -1311,7 +1314,7 @@ public:
                 TSequoiaChunkReplica replica;
                 replica.ChunkId = chunk->GetId();
                 replica.ReplicaIndex = replicaIndex;
-                replica.NodeId = storedReplica.GetPtr()->GetNode()->GetId();
+                replica.NodeId = location->GetNode()->GetId();
                 replica.LocationIndex = location->GetIndex();
                 sequoiaReplicas.push_back(replica);
             } else {
@@ -1687,11 +1690,15 @@ public:
         const auto& replicas = replicasOrError.Value();
         TNodePtrWithReplicaAndMediumIndexList result;
         for (auto replica : replicas) {
+            if (!replica.IsChunkLocation()) {
+                // TODO(cherepashka): handle this, when offshore mediums will be supported.
+                continue;
+            }
             if (replicaIndex != GenericChunkReplicaIndex && replica.GetReplicaIndex() != replicaIndex) {
                 continue;
             }
 
-            auto* chunkLocation = replica.GetPtr();
+            auto* chunkLocation = replica.AsChunkLocation().GetPtr();
             result.emplace_back(chunkLocation->GetNode(), replica.GetReplicaIndex(), chunkLocation->GetEffectiveMediumIndex());
         }
 
@@ -3179,13 +3186,13 @@ private:
         TChunkLocation* locationWithMaxId = nullptr;
 
         for (auto replica : chunk->StoredReplicas()) {
-            auto* medium = FindMediumByIndex(replica.GetPtr()->GetEffectiveMediumIndex());
-            if (!medium) {
+            auto* medium = FindMediumByIndex(replica.GetEffectiveMediumIndex());
+            if (!medium || !replica.IsChunkLocation()) {
                 continue;
             }
 
             // We do not care about approvedness.
-            auto location = replica.GetPtr();
+            auto location = replica.AsChunkLocation().GetPtr();
             if (!locationWithMaxId || location->GetId() > locationWithMaxId->GetId()) {
                 locationWithMaxId = location;
             }
@@ -4722,7 +4729,13 @@ private:
         auto* medium = GetMediumByNameOrThrow(mediumName);
         int mediumIndex = medium->GetIndex();
 
-        auto replicationFactor = isErasure ? 1 : subrequest->replication_factor();
+        // NB(achulkov2): It makes no sense to place erasure chunks on offshore media, since offshore media ensure durability on their own.
+        // NB: Instead, we could allow utilizing optimizations of underlying storage systems, e.g. the same erasure coding.
+        THROW_ERROR_EXCEPTION_IF(medium->IsOffshore() && isErasure, "Erasure chunks cannot be placed on offshore media");
+        // NB: Not supported for now, but theoretically possible, if the underlying storage supports appending to files.
+        THROW_ERROR_EXCEPTION_IF(medium->IsOffshore() && isJournal, "Journal chunks cannot be placed on offshore media");
+
+        auto replicationFactor = (isErasure || medium->IsOffshore()) ? 1 : subrequest->replication_factor();
         ValidateReplicationFactor(replicationFactor);
 
         auto transactionId = FromProto<TTransactionId>(subrequest->transaction_id());
@@ -5098,7 +5111,10 @@ private:
                 YT_ASSERT_THREAD_AFFINITY_ANY();
 
                 for (auto replica : chunk->StoredReplicas()) {
-                    auto* location = replica.GetPtr();
+                    if (!replica.IsChunkLocation()) {
+                        continue;
+                    }
+                    auto* location = replica.AsChunkLocation().GetPtr();
                     auto* scratchData = location->GetLoadScratchData();
                     scratchData->Replicas[scratchData->CurrentReplicaIndex++] = TChunkPtrWithReplicaInfo(
                         chunk,
--- a/yt/yt/server/master/chunk_server/chunk_owner_node_proxy.cpp
+++ b/yt/yt/server/master/chunk_server/chunk_owner_node_proxy.cpp
@@ -139,8 +139,8 @@ void CanonizeCellTags(TCellTagList* cellTags)
         cellTags->end());
 }
 
-void PopulateChunkSpecWithReplicas(
-    const TChunkLocationPtrWithReplicaInfoList& chunkReplicas,
+static void PopulateChunkSpecWithReplicas(
+    const TStoredReplicaList& chunkReplicas,
     bool fetchParityReplicas,
     NNodeTrackerServer::TNodeDirectoryBuilder* nodeDirectoryBuilder,
     NChunkClient::NProto::TChunkSpec* chunkSpec)
@@ -153,13 +153,16 @@ void PopulateChunkSpecWithReplicas(
         ? std::numeric_limits<int>::max() // all replicas are feasible
         : NErasure::GetCodec(erasureCodecId)->GetDataPartCount();
 
-    auto addReplica = [&] (TChunkLocationPtrWithReplicaInfo replica)  {
+    auto addReplica = [&] (TStoredReplica replica)  {
+        if (!replica.IsChunkLocation()) {
+            return false;
+        }
         if (replica.GetReplicaIndex() >= firstInfeasibleReplicaIndex) {
             return false;
         }
-        const auto* location = replica.GetPtr();
-        replicas.emplace_back(location->GetNode(), replica.GetReplicaIndex(), location->GetEffectiveMediumIndex());
-        nodeDirectoryBuilder->Add(replica);
+        const auto* location = replica.AsChunkLocation().GetPtr();
+        replicas.emplace_back(location->GetNode(), replica.GetReplicaIndex(), replica.GetEffectiveMediumIndex());
+        nodeDirectoryBuilder->Add(replica.AsChunkLocation());
         return true;
     };
 
@@ -266,7 +269,7 @@ void BuildReplicalessChunkSpec(
 void BuildChunkSpec(
     TBootstrap* bootstrap,
     TChunk* chunk,
-    const TChunkLocationPtrWithReplicaInfoList& chunkReplicas,
+    const TStoredReplicaList& chunkReplicas,
     std::optional<i64> rowIndex,
     std::optional<int> tabletIndex,
     const TReadLimit& lowerLimit,
@@ -443,9 +446,9 @@ private:
                 return false;
             }
 
-            const auto& replicas = chunkReplicasOrError.Value();
+            const auto& chunkReplicas = chunkReplicasOrError.Value();
             PopulateChunkSpecWithReplicas(
-                replicas,
+                chunkReplicas,
                 FetchContext_.FetchParityReplicas,
                 &NodeDirectoryBuilder_,
                 &chunkSpec);
@@ -1050,7 +1053,7 @@ TFuture<TYsonString> TChunkOwnerNodeProxy::GetBuiltinAttributeAsync(TInternedAtt
                     int chosenMediumReplicaCount = 0;
 
                     for (auto replica : replicas) {
-                        int mediumIndex = replica.GetPtr()->GetEffectiveMediumIndex();
+                        int mediumIndex = replica.GetEffectiveMediumIndex();
                         if (mediumIndex == chosenMediumIndex || chosenMediumReplicaCount == 0) {
                             chosenMediumIndex = mediumIndex;
                             ++chosenMediumReplicaCount;
--- a/yt/yt/server/master/chunk_server/chunk_placement.cpp
+++ b/yt/yt/server/master/chunk_server/chunk_placement.cpp
@@ -437,7 +437,7 @@ TNode* TChunkPlacement::FindPreferredNode(
 std::optional<TNodeList> TChunkPlacement::FindConsistentPlacementWriteTargets(
     TDomesticMedium* medium,
     TChunk* chunk,
-    const TChunkLocationPtrWithReplicaInfoList& replicas,
+    const TStoredReplicaList& replicas,
     const TChunkReplicaIndexList& replicaIndexes,
     int desiredCount,
     int minCount,
@@ -532,16 +532,21 @@ std::optional<TNodeList> TChunkPlacement::FindConsistentPlacementWriteTargets(
 
     auto isNodeConsistent = [&] (TNode* node, int replicaIndex) {
         for (auto replica : replicas) {
-            if (replica.GetPtr()->GetEffectiveMediumIndex() != mediumIndex) {
+            if (!replica.IsChunkLocation()) {
+                continue;
+            }
+            if (replica.GetEffectiveMediumIndex() != mediumIndex) {
                 continue;
             }
 
+            auto locationReplica = replica.AsChunkLocation();
+
             if (replicaIndex == NChunkClient::GenericChunkReplicaIndex) {
-                if (replica.GetPtr()->GetNode() == node) {
+                if (locationReplica.GetPtr()->GetNode() == node) {
                     return true;
                 }
             } else if (replica.GetReplicaIndex() == replicaIndex) {
-                return replica.GetPtr()->GetNode() == node;
+                return locationReplica.GetPtr()->GetNode() == node;
             }
         }
 
--- a/yt/yt/server/master/chunk_server/chunk_proxy.cpp
+++ b/yt/yt/server/master/chunk_server/chunk_proxy.cpp
@@ -105,6 +105,14 @@ static_assert(IsLeaderChunkScanKind(EChunkScanKind::GlobalStatisticsCollector));
 
 ////////////////////////////////////////////////////////////////////////////////
 
+DEFINE_BIT_ENUM(EStoredReplicasType,
+    ((None)        (0x0000))
+    ((Master)      (0x0001))
+    ((Offshore)    (0x0002))
+);
+
+////////////////////////////////////////////////////////////////////////////////
+
 class TChunkProxy
     : public TNonversionedObjectProxyBase<TChunk>
 {
@@ -131,6 +139,9 @@ private:
             .SetPresent(!isForeign)
             .SetOpaque(true));
         descriptors->push_back(TAttributeDescriptor(EInternedAttributeKey::StoredMasterReplicas)
+            .SetPresent(!isForeign));
+        // NB: Offshore replicas are not stored in sequoia and thus don't have to be opaque for now.
+        descriptors->push_back(TAttributeDescriptor(EInternedAttributeKey::StoredOffshoreReplicas)
             .SetPresent(!isForeign));
         descriptors->push_back(TAttributeDescriptor(EInternedAttributeKey::StoredSequoiaReplicas)
             .SetPresent(!isForeign)
@@ -298,7 +309,7 @@ private:
                     fluent
                         .Item("location_uuid").Value(location->GetUuid());
                 })
-                .DoIf(node->IsDecommissioned(), [&] (TFluentMap fluent) {
+                .DoIf(node && node->IsDecommissioned(), [&] (TFluentMap fluent) {
                     fluent
                         .Item("decommissioned").Value(true);
                 })
@@ -311,27 +322,41 @@ private:
                         .Item("state").Value(replicaState);
                 })
             .EndAttributes()
-            .Value(node->GetDefaultAddress());
+            // Maybe not the cleanest approach, but good enough for now.
+            .Value(node ? node->GetDefaultAddress() : TString(medium->GetName()));
     };
 
     void BuildYsonReplicas(
         IYsonConsumer* consumer,
         const IChunkManagerPtr& chunkManager,
         TChunkId chunkId,
-        TChunkLocationPtrWithReplicaInfoList replicas)
+        TStoredReplicaList replicas,
+        EStoredReplicasType storedReplicasType)
     {
-        SortBy(replicas, [] (TChunkLocationPtrWithReplicaInfo replica) {
-            return std::tuple(replica.GetReplicaIndex(), replica.GetPtr()->GetEffectiveMediumIndex());
+        SortBy(replicas, [] (TStoredReplica replica) {
+            return std::tuple(replica.GetReplicaIndex(), replica.GetEffectiveMediumIndex());
         });
 
         BuildYsonFluently(consumer)
-            .DoListFor(replicas, [&] (TFluentList fluent, TChunkLocationPtrWithReplicaInfo replica) {
-                const auto* location = replica.GetPtr();
+            .DoListFor(replicas, [&] (TFluentList fluent, TStoredReplica replica) {
+                if (None(storedReplicasType & EStoredReplicasType::Master) && replica.IsChunkLocation()) {
+                    return;
+                }
+                if (None(storedReplicasType & EStoredReplicasType::Offshore) && replica.IsMedium()) {
+                    return;
+                }
+
+                TChunkLocation* location = nullptr;
+                TNodeRawPtr node = nullptr;
+                if (replica.IsChunkLocation()) {
+                    location = replica.AsChunkLocation().GetPtr();
+                    node = location->GetNode();
+                }
                 SerializeReplica(
                     fluent,
                     chunkManager,
                     chunkId,
-                    location->GetNode(),
+                    node,
                     location,
                     replica.GetReplicaIndex(),
                     replica.GetReplicaState(),
@@ -358,9 +383,29 @@ private:
                     break;
                 }
 
-                auto masterReplicas = chunk->StoredReplicas();
-                TChunkLocationPtrWithReplicaInfoList replicaList(masterReplicas.begin(), masterReplicas.end());
-                BuildYsonReplicas(consumer, chunkManager, chunk->GetId(), replicaList);
+                auto storedReplicas = chunk->StoredReplicas();
+                TStoredReplicaList replicaList(storedReplicas.begin(), storedReplicas.end());
+                BuildYsonReplicas(
+                    consumer,
+                    chunkManager,
+                    chunk->GetId(),
+                    replicaList,
+                    EStoredReplicasType::Master);
+                return true;
+            }
+
+             case EInternedAttributeKey::StoredOffshoreReplicas: {
+                if (isForeign) {
+                    break;
+                }
+
+                auto storedReplicas = chunk->StoredReplicas();
+                TStoredReplicaList replicaList(storedReplicas.begin(), storedReplicas.end());
+                BuildYsonReplicas(consumer,
+                    chunkManager,
+                    chunk->GetId(),
+                    replicaList,
+                    EStoredReplicasType::Offshore);
                 return true;
             }
 
@@ -1124,7 +1169,12 @@ private:
                         auto aliveReplicas = chunkReplicaFetcher->FilterAliveReplicas(replicas);
                         return BuildYsonStringFluently()
                             .Do([&] (auto fluent) {
-                                BuildYsonReplicas(fluent.GetConsumer(), chunkManager, chunkId, aliveReplicas);
+                                BuildYsonReplicas(
+                                    fluent.GetConsumer(),
+                                    chunkManager,
+                                    chunkId,
+                                    aliveReplicas,
+                                    EStoredReplicasType::Master | EStoredReplicasType::Offshore);
                             });
                     }));
             }
@@ -1140,10 +1190,15 @@ private:
                         auto it = replicas.find(chunkId);
                         const auto& chunkReplicas = it != replicas.end()
                             ? chunkReplicaFetcher->FilterAliveReplicas(it->second)
-                            : TChunkLocationPtrWithReplicaInfoList();
+                            : TStoredReplicaList();
                         return BuildYsonStringFluently()
                             .Do([&] (auto fluent) {
-                                BuildYsonReplicas(fluent.GetConsumer(), chunkManager, chunkId, chunkReplicas);
+                                BuildYsonReplicas(
+                                    fluent.GetConsumer(),
+                                    chunkManager,
+                                    chunkId,
+                                    chunkReplicas,
+                                    EStoredReplicasType::Master);
                             });
                     }));
             }
--- a/yt/yt/server/master/chunk_server/chunk_replica.cpp
+++ b/yt/yt/server/master/chunk_server/chunk_replica.cpp
@@ -1,5 +1,6 @@
 #include "chunk_replica.h"
 #include "chunk.h"
+#include "medium_base.h"
 
 #include <yt/yt/server/master/node_tracker_server/node.h>
 
@@ -105,6 +106,40 @@ void ToProto(ui32* protoValue, TChunkLocationPtrWithReplicaIndex value)
 void ToProto(ui32* protoValue, TChunkLocationPtrWithReplicaInfo value)
 {
     ToProto(protoValue, TChunkLocationPtrWithReplicaIndex(value.GetPtr(), value.GetReplicaIndex()));
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+void FormatValue(TStringBuilderBase* builder, TMediumPtrWithReplicaInfo value, TStringBuf /*spec*/)
+{
+    auto* medium = value.GetPtr();
+
+    builder->AppendFormat("%v", medium->GetName());
+
+    if (value.GetReplicaIndex() != GenericChunkReplicaIndex) {
+        builder->AppendFormat("/%v", value.GetReplicaIndex());
+    }
+
+    if (value.GetReplicaState() != EChunkReplicaState::Generic) {
+        builder->AppendFormat(":%v", value.GetReplicaState());
+    }
+
+    builder->AppendFormat("@%v", medium->GetIndex());
+}
+
+void ToProto(ui64* protoValue, TMediumPtrWithReplicaInfo value)
+{
+    // Only applicable to offshore replicas. Domestic replicas require real node ids.
+    THROW_ERROR_EXCEPTION_UNLESS(
+        value.GetPtr()->IsOffshore(),
+        "Failed to serialize domestic medium %v",
+        value.GetPtr()->GetName());
+
+    TChunkReplicaWithMedium replica(
+        NNodeTrackerClient::OffshoreNodeId,
+        value.GetReplicaIndex(),
+        value.GetPtr()->GetIndex());
+    NChunkClient::ToProto(protoValue, replica);
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/yt/yt/server/master/chunk_server/chunk_replica_fetcher.cpp
+++ b/yt/yt/server/master/chunk_server/chunk_replica_fetcher.cpp
@@ -209,7 +209,7 @@ public:
             }));
     }
 
-    TErrorOr<TChunkLocationPtrWithReplicaInfoList> GetChunkReplicas(
+    TErrorOr<TStoredReplicaList> GetChunkReplicas(
         const TEphemeralObjectPtr<TChunk>& chunk,
         bool includeUnapproved) const override
     {
@@ -222,7 +222,7 @@ public:
         auto result = GetChunkReplicas(chunks, includeUnapproved);
         return GetOrCrash(result, chunk->GetId());
     }
-    TChunkToLocationPtrWithReplicaInfoList GetChunkReplicas(
+    TChunkToStoredReplicaList GetChunkReplicas(
         const std::vector<TEphemeralObjectPtr<TChunk>>& chunks,
         bool includeUnapproved) const override
     {
@@ -263,16 +263,14 @@ public:
             auto masterReplicas = chunk->StoredReplicas();
             std::vector<TSequoiaChunkReplica> replicas;
             for (const auto& masterReplica : masterReplicas) {
-                auto location = masterReplica.GetPtr();
                 TSequoiaChunkReplica replica;
                 replica.ChunkId = chunk->GetId();
                 replica.ReplicaIndex = masterReplica.GetReplicaIndex();
-                replica.NodeId = location->GetNode()->GetId();
-                replica.LocationIndex = location->GetIndex();
+                replica.NodeId = masterReplica.GetNodeId();
+                replica.LocationIndex = masterReplica.GetChunkLocationIndex();
                 replica.ReplicaState = masterReplica.GetReplicaState();
                 replicas.push_back(replica);
             }
-            TChunkLocationPtrWithReplicaInfoList replicaList(masterReplicas.begin(), masterReplicas.end());
             EmplaceOrCrash(result, chunk->GetId(), replicas);
         }
 
@@ -472,10 +470,10 @@ private:
         return Bootstrap_->GetConfigManager()->GetConfig()->ChunkManager->SequoiaChunkReplicas;
     }
 
-    TChunkLocationPtrWithReplicaInfoList FilterAliveReplicas(const std::vector<TSequoiaChunkReplica>& replicas) const override
+    TStoredReplicaList FilterAliveReplicas(const std::vector<TSequoiaChunkReplica>& replicas) const override
     {
         const auto& dataNodeTracker = Bootstrap_->GetDataNodeTracker();
-        TChunkLocationPtrWithReplicaInfoList aliveReplicas;
+        TStoredReplicaList aliveReplicas;
         for (const auto& replica : replicas) {
             auto chunkId = replica.ChunkId;
             auto locationIndex = replica.LocationIndex;
@@ -487,19 +485,19 @@ private:
                 continue;
             }
 
-            aliveReplicas.emplace_back(location, replica.ReplicaIndex, replica.ReplicaState);
+            aliveReplicas.emplace_back(TChunkLocationPtrWithReplicaInfo(location, replica.ReplicaIndex, replica.ReplicaState));
         }
         return aliveReplicas;
     }
 
-    TChunkToLocationPtrWithReplicaInfoList GetReplicas(
+    TChunkToStoredReplicaList GetReplicas(
         const std::vector<TEphemeralObjectPtr<TChunk>>& chunks,
         const TErrorOr<THashMap<TChunkId, std::vector<TSequoiaChunkReplica>>>& sequoiaReplicasOrError,
         const std::vector<TChunkId>& sequoiaChunkIds) const
     {
         VerifyPersistentStateRead();
 
-        TChunkToLocationPtrWithReplicaInfoList result;
+        TChunkToStoredReplicaList result;
         if (!sequoiaReplicasOrError.IsOK()) {
             for (auto chunkId : sequoiaChunkIds) {
                 EmplaceOrCrash(result, chunkId, TError(sequoiaReplicasOrError));
@@ -512,7 +510,7 @@ private:
 
         for (const auto& chunk : chunks) {
             auto masterReplicas = chunk->StoredReplicas();
-            TChunkLocationPtrWithReplicaInfoList replicaList(masterReplicas.begin(), masterReplicas.end());
+            TStoredReplicaList replicaList(masterReplicas.begin(), masterReplicas.end());
             auto [it, inserted] = result.emplace(chunk->GetId(), replicaList);
 
             if (inserted) {
@@ -526,12 +524,10 @@ private:
             auto& replicas = it->second.Value();
             replicas.insert(replicas.end(), masterReplicas.begin(), masterReplicas.end());
 
-            SortUniqueBy(replicas, [] (const auto& replica) {
-                auto location = replica.GetPtr();
-
+            SortUniqueBy(replicas, [] (const TStoredReplica& replica) {
                 auto replicaIndex = replica.GetReplicaIndex();
-                auto nodeId = location->GetNode()->GetId();
-                auto locationUuid = location->GetUuid();
+                auto nodeId = replica.GetNodeId();
+                auto locationUuid = replica.GetLocationUuid();
                 return std::tuple(replicaIndex, nodeId, locationUuid);
             });
         }
--- a/yt/yt/server/master/chunk_server/chunk_replicator.cpp
+++ b/yt/yt/server/master/chunk_server/chunk_replicator.cpp
@@ -518,7 +518,7 @@ void TChunkReplicator::TouchChunk(TChunk* chunk)
 
 TCompactMediumMap<EChunkStatus> TChunkReplicator::ComputeChunkStatuses(
     TChunk* chunk,
-    const TChunkLocationPtrWithReplicaInfoList& replicas)
+    const TStoredReplicaList& replicas)
 {
     TCompactMediumMap<EChunkStatus> result;
 
@@ -533,15 +533,29 @@ TCompactMediumMap<EChunkStatus> TChunkReplicator::ComputeChunkStatuses(
 
 ECrossMediumChunkStatus TChunkReplicator::ComputeCrossMediumChunkStatus(
     TChunk* chunk,
-    const TChunkLocationPtrWithReplicaInfoList& replicas)
+    const TStoredReplicaList& replicas)
 {
     return ComputeChunkStatistics(chunk, replicas).Status;
 }
 
 TChunkReplicator::TChunkStatistics TChunkReplicator::ComputeChunkStatistics(
     const TChunk* chunk,
-    const TChunkLocationPtrWithReplicaInfoList& replicas)
+    const TStoredReplicaList& replicas)
 {
+    std::vector<TMediumPtrWithReplicaInfo> offshoreReplicas;
+    for (const auto& replica: replicas) {
+        if (replica.IsMedium()) {
+            offshoreReplicas.push_back(replica.AsMedium());
+        }
+    }
+    if (chunk->IsErasure() && !offshoreReplicas.empty()) {
+        YT_LOG_ALERT(
+            "Erasure chunk has offshore replicas (ChunkId: %v, Replicas: %v, OffshoreReplicas: %v)",
+            chunk->GetId(),
+            replicas,
+            offshoreReplicas);
+    }
+
     auto result = chunk->IsErasure()
         ? ComputeErasureChunkStatistics(chunk, replicas)
         : ComputeRegularChunkStatistics(chunk, replicas);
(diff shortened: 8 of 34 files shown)


yt/yt/server/master/chunk_server/stored_chunk_replica.h
```diff
@@ -0,0 +1,60 @@
...
+
+class TStoredReplica
+{
+public:
+    TStoredReplica();
+    TStoredReplica(const TChunkLocationPtrWithReplicaInfo& chunkLocation);
+    TStoredReplica(const TMediumPtrWithReplicaInfo& medium);
+
+    bool operator==(const TStoredReplica& other) const = default;
+    bool operator<(const TStoredReplica& other) const;
```
raundn: Seems like `bool operator<=>(const TStoredReplica& other) const = default;` would work here.

yt/yt/server/master/chunk_server/stored_chunk_replica.h
```diff
@@ -0,0 +1,60 @@
...
+    EChunkReplicaState GetReplicaState() const;
+
+    NNodeTrackerClient::TChunkLocationIndex GetChunkLocationIndex() const;
+    NChunkClient::TChunkLocationUuid GetLocationUuid() const;
+    NNodeTrackerClient::TNodeId GetNodeId() const;
+
+    void Save(NCellMaster::TSaveContext& context) const;
+    void Load(NCellMaster::TLoadContext& context);
+private:
+    std::variant<TChunkLocationPtrWithReplicaInfo, TMediumPtrWithReplicaInfo> ReplicaInfo_;
```
raundn: We'll have to do a back-of-the-envelope calculation and see whether this will be prohibitively expensive (memory-wise). Before the changes, each replica was represented by just a pointer (bit-packing inside `TChunkLocationPtrWithReplicaInfo` notwithstanding), now its larger (due to type tag inside std::variant). 

There's a trick to be proposed here: using a bit (or two) inside `TSomethingPtrWithSomething` as a type tag but let's discuss this before going that route =)

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
```diff
@@ -0,0 +1,198 @@
...
+}
+
+bool TStoredReplica::IsChunkLocation() const
+{
+    return std::holds_alternative<TChunkLocationPtrWithReplicaInfo>(ReplicaInfo_);
+}
+
+TChunkLocationPtrWithReplicaInfo& TStoredReplica::AsChunkLocation()
+{
+    return std::get<0>(ReplicaInfo_);
```
raundn: Better to use `get_if` and `VERIFY`. If the code uses `TStoredReplica` correctly, there's no difference. If not, it's better to crash than throw bad_variant_access nobody expects.

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
```diff
@@ -0,0 +1,198 @@
...
+}
+
+const TChunkLocationPtrWithReplicaInfo& TStoredReplica::AsChunkLocation() const
+{
+    return std::get<0>(ReplicaInfo_);
```
raundn: Same here and so on.

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
```diff
@@ -0,0 +1,198 @@
...
+namespace NYT::NChunkServer {
+
+////////////////////////////////////////////////////////////////////////////////
+
+TStoredReplica::TStoredReplica()
+{ }
+
+TStoredReplica::TStoredReplica(const TChunkLocationPtrWithReplicaInfo& chunkLocation)
+    : ReplicaInfo_(chunkLocation)
+{ }
```
raundn: `YT_VERIFY(chunkLocation)`?

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+TStoredReplica::TStoredReplica(const TMediumPtrWithReplicaInfo& medium)
+    : ReplicaInfo_(medium)
+{ }
```
raundn: `YT_VERIFY(medium)`?

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
```diff
@@ -0,0 +1,198 @@
...
+        });
+}
+
+NNodeTrackerClient::TChunkLocationIndex TStoredReplica::GetChunkLocationIndex() const
+{
+    return Visit(ReplicaInfo_,
+        [] (const TChunkLocationPtrWithReplicaInfo& chunkLocation) {
+            auto* location = chunkLocation.GetPtr();
+            if (!NObjectServer::IsObjectAlive(location)) {
+                return NNodeTrackerClient::InvalidChunkLocationIndex;
```
raundn: Hm... Why are we concerned with location being alive here but not, for example, in `GetEffectiveMediumIndex`?

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
```diff
@@ -0,0 +1,198 @@
...
+            FormatValue(builder, chunkLocation, spec);
+        },
+        [&] (const TMediumPtrWithReplicaInfo& medium) {
+            FormatValue(builder, medium, spec);
+        });
+}
+
+void ToProto(ui64* protoValue, TStoredReplica value)
+{
+
```
raundn, line 248: Extraneous empty line.

yt/yt/server/master/chunk_server/chunk.cpp
```diff
@@ -280,10 +280,16 @@ bool TChunk::HasParents() const
...
 
 void TChunk::AddReplica(
-    TChunkLocationPtrWithReplicaInfo replica,
-    const TDomesticMedium* medium,
+    TStoredReplica replica,
+    const TMedium* medium,
     bool approved)
 {
+    if (medium->IsOffshore()) {
+        auto* data = MutableReplicasData();
```
raundn: Let's pull this above `if`. We'll need it in either case.

yt/yt/server/master/chunk_server/chunk.cpp
```diff
@@ -280,10 +280,16 @@ bool TChunk::HasParents() const
...
+        data->AddStoredReplica(replica);
```
raundn: Not having read the code ahead, I have to wonder: is it guaranteed (by the caller) that `replica` is actually a pointer to medium? And that it points the the same medium as `medium` does.

yt/yt/server/master/chunk_server/chunk_manager.cpp
```diff
@@ -1299,8 +1298,12 @@ class TChunkManager
         // Unregister chunk replicas from all known locations.
         // Schedule removal jobs.
         for (auto storedReplica : chunk->StoredReplicas()) {
+            if (!storedReplica.IsChunkLocation()) {
+                // TODO(cherepashka): handle this, when offshore mediums will be supported.
```
raundn: - "when offshore media"
- the comma is not really necessary

yt/yt/server/master/chunk_server/chunk_manager.cpp
```diff
@@ -1299,8 +1298,12 @@ class TChunkManager
         // Unregister chunk replicas from all known locations.
         // Schedule removal jobs.
         for (auto storedReplica : chunk->StoredReplicas()) {
+            if (!storedReplica.IsChunkLocation()) {
```
raundn: For now, this is ok. But strategically, these ifs are great markers: they precede chunks of code that should (eventually) be rendered polymorphic (not necessarily as virtual methods of `TStoredReplica` but something along those lines). The goal here, really, is to make it so that chunk manager and replicator code is expressed at a certain level of abstraction, unaware of the domestic-offshore distinction. I'm not sure this is achievable (in fact, I'm pretty sure it's not, at least not for the replicator) but we should strive to make the code that bothers with that distinction as small (and as localized) as possible.

yt/yt/server/master/chunk_server/chunk_manager.cpp
```diff
@@ -3179,13 +3186,13 @@ class TChunkManager
         TChunkLocation* locationWithMaxId = nullptr;
 
         for (auto replica : chunk->StoredReplicas()) {
-            auto* medium = FindMediumByIndex(replica.GetPtr()->GetEffectiveMediumIndex());
-            if (!medium) {
+            auto* medium = FindMediumByIndex(replica.GetEffectiveMediumIndex());
+            if (!medium || !replica.IsChunkLocation()) {
```
raundn: Let's split such ifs. `if ([!]replica.IsChunkLocations())` should always go separate, seeing as they are subject to future change.

yt/yt/server/master/chunk_server/chunk_manager.cpp
```diff
@@ -4722,7 +4729,13 @@ class TChunkManager
         auto* medium = GetMediumByNameOrThrow(mediumName);
         int mediumIndex = medium->GetIndex();
 
-        auto replicationFactor = isErasure ? 1 : subrequest->replication_factor();
+        // NB(achulkov2): It makes no sense to place erasure chunks on offshore media, since offshore media ensure durability on their own.
+        // NB: Instead, we could allow utilizing optimizations of underlying storage systems, e.g. the same erasure coding.
```
raundn, line 4660: This limitation is completely rational. It's also doomed from the start.

If we are to support offshore media in full, the replicator will have to be able to move chunks from domestic to offshore media (and the other way around). So this code is really a temporary protection. For the time being, you should probably leave it intact but be aware that erasure chunks on offshore media are an inevitability.

yt/yt/server/master/chunk_server/chunk_manager.cpp
```diff
@@ -4722,7 +4729,13 @@ class TChunkManager
...
+        THROW_ERROR_EXCEPTION_IF(medium->IsOffshore() && isErasure, "Erasure chunks cannot be placed on offshore media");
+        // NB: Not supported for now, but theoretically possible, if the underlying storage supports appending to files.
```
raundn, line 4662: This, on the other hand, is a good idea. Journal chunks are never intermixed with blob chunks, so prohibiting them on offshore media is a simple matter of 
  - forbidding creating journal chunks on offshore media, and
  - forbidding assigning offshore media to journal nodes.

Both of these measures should probably taken right now.

As a general note, I personally don't believe we'll ever need to put journal chunks on offshore media. The idea just doesn't seem reasonable. I could be wrong, though.

yt/yt/server/master/chunk_server/chunk_manager.cpp
```diff
@@ -4722,7 +4729,13 @@ class TChunkManager
...
+        THROW_ERROR_EXCEPTION_IF(medium->IsOffshore() && isJournal, "Journal chunks cannot be placed on offshore media");
+
+        auto replicationFactor = (isErasure || medium->IsOffshore()) ? 1 : subrequest->replication_factor();
```
raundn: Let's split these:

```cpp
auto replicationFactor = subrequest->replication_factor();
if (isErasure) {
    replicationFactor = 1;
}
if (medium->IsOffshore()) {
    replicationFactor = 1;
}
```

Semantically these are two very different reasons to cap RF at 1, so let's keep them separate.

yt/yt/server/master/chunk_server/chunk_proxy.cpp
```diff
@@ -132,6 +140,9 @@ class TChunkProxy
             .SetOpaque(true));
         descriptors->push_back(TAttributeDescriptor(EInternedAttributeKey::StoredMasterReplicas)
             .SetPresent(!isForeign));
+        // NB: Offshore replicas are not stored in sequoia and thus don't have to be opaque for now.
```
raundn: in Sequoia

yt/yt/server/master/chunk_server/chunk_proxy.cpp
```diff
@@ -105,6 +105,14 @@ static_assert(IsLeaderChunkScanKind(EChunkScanKind::GlobalStatisticsCollector));
 
 ////////////////////////////////////////////////////////////////////////////////
 
+DEFINE_BIT_ENUM(EStoredReplicasType,
+    ((None)        (0x0000))
+    ((Master)      (0x0001))
```
raundn: "Master replicas" usually means "replicas stored as part of persistent state" (as opposed to "replicas stored in dynamic tables", i.e. "Sequoia replicas"). How about `Domestic`?

yt/yt/server/master/chunk_server/chunk_proxy.cpp
```diff
@@ -311,27 +322,41 @@ class TChunkProxy
...
+                }
+                if (None(storedReplicasType & EStoredReplicasType::Offshore) && replica.IsMedium()) {
+                    return;
+                }
+
+                TChunkLocation* location = nullptr;
+                TNodeRawPtr node = nullptr;
+                if (replica.IsChunkLocation()) {
+                    location = replica.AsChunkLocation().GetPtr();
+                    node = location->GetNode();
```
raundn, line 339: If there's `TStoredReplica::GetNodeId` (that returns invalid node id for offshore), I don't see any reason not to have `TStoredReplica::GetNode` (returning nullptr in the offshore case). Maybe it's not worth it, though. Depends on how often we need to write this code here.

yt/yt/server/master/chunk_server/chunk_proxy.cpp
```diff
@@ -358,9 +383,29 @@ class TChunkProxy
...
+                BuildYsonReplicas(
+                    consumer,
+                    chunkManager,
+                    chunk->GetId(),
+                    replicaList,
+                    EStoredReplicasType::Master);
+                return true;
+            }
+
+             case EInternedAttributeKey::StoredOffshoreReplicas: {
```
raundn: The fact that `@stored_master_replicas` now really means "stored domestic master replicas" is unfortunate :-/ 

Also, it's would be logical (however inane) to support offshore replicas in Sequoia at some point in the future. This would lead to further naming troubles: we'll have to support `stored_{all,domestic,offshore}_{all,master,sequoia}_replicas`. This is starting to look like an overkill.

Anyway, I have an alternative proposal: get rid of `stored_offshore_replicas` and instead extend the format `SerializeReplica` produces with a new "offshore" attribute. The client code then may filter replicas however it pleases. 

To sum up, 
  - `@stored_replicas` - all the replicas (domestic, offshore, master, sequoia - all of them);
  - `@stored_master_replicas` - only master replicas (both domestic and, in theory, offshore);
  - `@stored_sequoia_replicas` - only Sequoia replicas (both domestic and, in theory, offshore);
  - no other attributes.

```bash
yt get "#$chunk_id/@stored_replicas
[
    <
        "medium" = "default";
        "location_uuid" = "2ae455-1c9ac8be-444ffd75-90adb1c3";
    > "sas4-5324-dat-node-hume.man-pre.yp-c.yandex.net:9012";
    <
        "medium" = "default";
        "location_uuid" = "70374d10-19decd31-a48a005-74e863d2";
    > "sas4-5257-dat-node-hume.man-pre.yp-c.yandex.net:9012";
    <
        "medium" = "dirt_cheap_s3";
        "offshore" = %true;
    > "dirt_cheap_s3";
]
```
Since most replicas are going to be domestic, only offshore replicas get extra attribute.

yt/yt/server/master/chunk_server/chunk_replica.cpp
```diff
@@ -109,6 +110,40 @@ void ToProto(ui32* protoValue, TChunkLocationPtrWithReplicaInfo value)
...
+    }
+
+    builder->AppendFormat("@%v", medium->GetIndex());
+}
+
+void ToProto(ui64* protoValue, TMediumPtrWithReplicaInfo value)
+{
+    // Only applicable to offshore replicas. Domestic replicas require real node ids.
+    THROW_ERROR_EXCEPTION_UNLESS(
+        value.GetPtr()->IsOffshore(),
```
raundn: Let's move the condition to the same line as the macro.

yt/yt/server/master/chunk_server/chunk_replica.cpp
```diff
@@ -109,6 +110,40 @@ void ToProto(ui32* protoValue, TChunkLocationPtrWithReplicaInfo value)
...
+    if (value.GetReplicaState() != EChunkReplicaState::Generic) {
+        builder->AppendFormat(":%v", value.GetReplicaState());
+    }
+
+    builder->AppendFormat("@%v", medium->GetIndex());
+}
+
+void ToProto(ui64* protoValue, TMediumPtrWithReplicaInfo value)
+{
+    // Only applicable to offshore replicas. Domestic replicas require real node ids.
```
raundn, line 143: This is a strange place: should the serialization of replicas be polymorphic, too? Let's talk this over.

yt/yt/server/master/chunk_server/chunk_replica_fetcher.cpp
```diff
@@ -487,19 +485,19 @@ class TChunkReplicaFetcher
                 continue;
             }
 
-            aliveReplicas.emplace_back(location, replica.ReplicaIndex, replica.ReplicaState);
+            aliveReplicas.emplace_back(TChunkLocationPtrWithReplicaInfo(location, replica.ReplicaIndex, replica.ReplicaState));
```
raundn: We could support `TStoredReplica` constructors accepting (location, replicaIndex, replicaState) and  (medium, replicaIndex, replicaState) and thus avoid the necessity to write `TChunkLocationPtrWithReplicaInfo` in cases like this. Up to your taste.

yt/yt/server/master/chunk_server/chunk_replicator.cpp
```diff
@@ -533,15 +533,29 @@ TCompactMediumMap<EChunkStatus> TChunkReplicator::ComputeChunkStatuses(
...
 {
+    std::vector<TMediumPtrWithReplicaInfo> offshoreReplicas;
+    for (const auto& replica: replicas) {
+        if (replica.IsMedium()) {
+            offshoreReplicas.push_back(replica.AsMedium());
+        }
+    }
+    if (chunk->IsErasure() && !offshoreReplicas.empty()) {
+        YT_LOG_ALERT(
+            "Erasure chunk has offshore replicas (ChunkId: %v, Replicas: %v, OffshoreReplicas: %v)",
```
raundn, line 558: The `ComputeChunkStatistics` method is hot, so I propose a small optimization: let's form `offshoreReplicas` only when we're sure we need to alert and use `std::find_if` instead of `!offshoreReplicas.empty()`.

yt/yt/server/master/chunk_server/chunk_replicator.cpp
```diff
@@ -590,10 +604,18 @@ TChunkReplicator::TChunkStatistics TChunkReplicator::ComputeErasureChunkStatisti
 
     const auto& chunkManager = Bootstrap_->GetChunkManager();
     for (auto replica : replicas) {
-        auto* chunkLocation = replica.GetPtr();
+        if (!replica.IsChunkLocation()) {
+            YT_LOG_ALERT("Non chunk location stored replica occured during computation statistics for erasure chunk "
```
raundn: Non-chunk

raundn: - occur**r**ed
- and I'd have said "encountered" rather than "occurred" anyway.

yt/yt/server/master/chunk_server/chunk.cpp
```diff
@@ -332,15 +338,18 @@ void TChunk::RemoveReplica(
...
         }
     }
     YT_ABORT();
 }
 
-void TChunk::ApproveReplica(TChunkLocationPtrWithReplicaInfo replica)
+void TChunk::ApproveReplica(TStoredReplica replica)
 {
     auto* data = MutableReplicasData();
     ++data->ApprovedReplicaCount;
```
achulkov2, line 388: Let's verify that we don't approve offshore replicas? There is no such plan as of right now.

yt/yt/server/master/chunk_server/chunk_owner_node_proxy.cpp
```diff
@@ -153,13 +153,16 @@ void PopulateChunkSpecWithReplicas(
         ? std::numeric_limits<int>::max() // all replicas are feasible
         : NErasure::GetCodec(erasureCodecId)->GetDataPartCount();
 
-    auto addReplica = [&] (TChunkLocationPtrWithReplicaInfo replica)  {
+    auto addReplica = [&] (TStoredReplica replica)  {
+        if (!replica.IsChunkLocation()) {
+            return false;
```
achulkov2, line 160: TODO to actually return these replicas in chunk specs, one more logic is here?

yt/yt/server/master/chunk_server/chunk_proxy.cpp
achulkov2: So `Master` here is only real master replicas, non-sequoia, right?

yt/yt/server/master/chunk_server/chunk_proxy.cpp
```diff
@@ -1140,10 +1190,15 @@ class TChunkProxy
...
+                            : TStoredReplicaList();
                         return BuildYsonStringFluently()
                             .Do([&] (auto fluent) {
-                                BuildYsonReplicas(fluent.GetConsumer(), chunkManager, chunkId, chunkReplicas);
+                                BuildYsonReplicas(
+                                    fluent.GetConsumer(),
+                                    chunkManager,
+                                    chunkId,
+                                    chunkReplicas,
+                                    EStoredReplicasType::Master);
```
achulkov2: Ok, now this is a bit confusing with `Master` type for sequoia replicas.

yt/yt/server/master/chunk_server/chunk_replicator.cpp
achulkov2: Shouldn't we log chunk id as well?

yt/yt/server/master/chunk_server/chunk_replicator.cpp
```diff
@@ -2401,6 +2428,20 @@ void TChunkReplicator::RefreshChunk(
 
     auto durabilityRequired = IsDurabilityRequired(chunk, chunkReplicas);
 
+    TChunkLocationPtrWithReplicaInfoList locationReplicas;
+    locationReplicas.reserve(chunkReplicas.size());
+    for (const auto& replica: chunkReplicas) {
+        if (!replica.IsChunkLocation()) {
+            YT_LOG_ALERT(
+                "Non chunk location stored replica was found during refresh of chunk "
```
achulkov2: ChunkId here as well.

yt/yt/server/master/node_tracker_server/node_directory_builder.cpp
```diff
@@ -51,6 +51,28 @@ void TNodeDirectoryBuilder::Add(TRange<TChunkLocationPtrWithReplicaInfo> locatio
     }
 }
 
+void TNodeDirectoryBuilder::Add(NChunkServer::TStoredReplica replica)
+{
+    if (replica.IsChunkLocation()) {
+        Add(replica.AsChunkLocation());
+    } else {
+        if (!ListedNodeIds_.insert(OffshoreNodeId).second) {
```
achulkov2: Maybe a TODO here, as right now this code is not very meaningful.

yt/yt/server/master/chunk_server/chunk_manager.cpp
achulkov2: Not relevant to this place in code, but what will happen if one now tries to confirm a chunk with offshore replicas in the confirm request? 

I think they will be skipped by 
```
auto* node = nodeTracker->FindNode(nodeId);
if (!IsObjectAlive(node)) {
```

But let's make it more explicit. Also maybe it should be `YT_LOG_ALERT`, because nobody should be sending such confirm requests at this moment.

yt/yt/server/master/chunk_server/chunk_replicator.cpp
```diff
@@ -2401,6 +2428,20 @@ void TChunkReplicator::RefreshChunk(
 
     auto durabilityRequired = IsDurabilityRequired(chunk, chunkReplicas);
 
+    TChunkLocationPtrWithReplicaInfoList locationReplicas;
+    locationReplicas.reserve(chunkReplicas.size());
+    for (const auto& replica: chunkReplicas) {
```
raundn: Looks like `const&` isn't really necessary.

yt/yt/server/master/chunk_server/chunk_replicator.cpp
```diff
@@ -2401,6 +2428,20 @@ void TChunkReplicator::RefreshChunk(
 
     auto durabilityRequired = IsDurabilityRequired(chunk, chunkReplicas);
 
+    TChunkLocationPtrWithReplicaInfoList locationReplicas;
+    locationReplicas.reserve(chunkReplicas.size());
```
raundn: This is a hot spot as well, so forming `locationReplicas` could be costly. I suggest first checking `chunkReplicas` for non-location replicas. If there aren't any, we may simply use `chunkReplicas` and avoid unnecessary work.

yt/yt/server/master/chunk_server/chunk_requisition.cpp
```diff
@@ -382,14 +382,16 @@ void TChunkRequisition::Load(NCellMaster::TLoadContext& context)
...
-void TChunkRequisition::ForceReplicationFactor(int replicationFactor)
+void TChunkRequisition::CorrectReplicationFactor(const IChunkManagerPtr& chunkManager, bool isErasureChunk)
 {
-    YT_ASSERT(replicationFactor > 0);
+    static_assert(MinReplicationFactor <= 1 && 1 <= MaxReplicationFactor,
+        "Replication factor limits are incorrect.");
 
     for (auto& entry : Entries_) {
-        auto& replicationPolicy = entry.ReplicationPolicy;
-        YT_VERIFY(replicationPolicy);
```
raundn, line 385: It seems this YT_VERIFY got dropped here. Let's bring it back and turn it into YT_LOG_ALERT while we're at it.

yt/yt/server/master/node_tracker_server/node_directory_builder.cpp
```diff
@@ -51,6 +51,28 @@ void TNodeDirectoryBuilder::Add(TRange<TChunkLocationPtrWithReplicaInfo> locatio
...
+            return;
+        }
+
+        auto* item = ProtoDirectory_->add_items();
```
raundn, line 60: This is a minor nitpick but this would be a bit more concise:

```cpp
    // ...
} else if (ListedNodeIds_.insert(OffshoreNodeId).second) {
    auto* item = ProtoDirectory_->add_items();
    // ....
```

yt/yt/server/master/node_tracker_server/node_tracker.cpp
```diff
@@ -1895,6 +1895,16 @@ class TNodeTracker
                 continue;
             }
 
+            // COMPAT(achulkov2, cherepashka): This is not supposed to happen on any real cluster. The code below is a safety precation to avoid
+            // damage to the cluster in such an unimaginalbe case. Updating a cluster from such a state will require local builds,
```
raundn: "local builds" are fine but I'd called them "custom builds". It's not locality that's important =)

yt/yt/server/master/node_tracker_server/node_tracker.cpp
```diff
@@ -1895,6 +1895,16 @@ class TNodeTracker
...
+            // but it is so unlikely that any cluster has had so many nodes, that we consider it impossible.
```
raundn: Extra comma.

yt/yt/server/master/node_tracker_server/node_tracker.cpp
```diff
@@ -1895,6 +1895,16 @@ class TNodeTracker
...
+            if (node->GetId() > MaxRealNodeId) {
+                YT_LOG_FATAL(
```
raundn, line 1956: BTW, I probably just forgot by the time I've reached the end of this PR but is there a protective counterpart to this YT_LOG_FATAL? I.e. is there a code that prevents a new node from being registered with too large of an ID?

raundn (changes requested):

yt/yt/server/master/chunk_server/chunk.cpp
cherepasshka: Right now only chunk location replicas are passed into this function, but I have added additional validations here

yt/yt/server/master/chunk_server/chunk_manager.cpp
cherepasshka: For now only TChunkReplicaWithLocationList is passed into confirm request: https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/server/master/chunk_server/chunk_manager.cpp#L932
Also there is alert in case node for confirming chunk is unknown: https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/server/master/chunk_server/chunk_manager.cpp#L878

yt/yt/server/master/chunk_server/chunk_replica.cpp
cherepasshka: The serialization for TMediumPtrWithReplicaInfo was placed in this file similarly to TChunkLocationPtrWithReplicaInfo's serialization
But serialization of TStoredReplica is in corresponding file: https://github.com/ytsaurus/ytsaurus/pull/1426/files#diff-d922524e4dbcd257c7b6b287eec5d530c5f4f8b25e839a126e22ae04996736caR56

yt/yt/server/master/chunk_server/chunk_replica_fetcher.cpp
cherepasshka: For now it is the only place, so I don't think this is necessary

yt/yt/server/master/chunk_server/chunk_proxy.cpp
cherepasshka: For now it is the only place, so I think it isn't worth it

yt/yt/server/master/chunk_server/chunk_manager.cpp
cherepasshka: Prohibiting to assigning offshore media is already implemented: https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/server/master/chunk_server/chunk_location_proxy.cpp#L139
I added todo to save throwing for journal nodes.
Prohibiting creating journal chunks is implemented right here

yt/yt/server/master/chunk_server/stored_chunk_replica.h
cherepasshka: (Stack) size of of TChunk was not increased, probably because of types alignment. Size of single stored replica increased from 8 to 16 bytes

yt/yt/server/master/chunk_server/chunk_manager.cpp
achulkov2: > For now only TChunkReplicaWithLocationList is passed into confirm request:
Ah, yes.

Then I guess it is fine, I can't see how we could pass OffshoreNodeId there.

It is just `YT_LOG_DEBUG` in the second link though, not alert.

yt/yt/server/master/chunk_server/stored_chunk_replica.h
achulkov2: Note that our [serializer](https://github.com/ytsaurus/ytsaurus/blob/6a9ec8d86133a48dacd6b27edf660d879e547910/yt/yt/core/misc/serialize-inl.h#L893) for `std::variant` unsurprisingly saves the index as well. So representation of replicas in snapshot will definitely increase, +4 bytes per replica.

[Here](https://gist.github.com/achulkov2/2ccba0c5e847b913ddcb6058c8721d90#how-can-we-dispatch-on-replica-type) are a few options how to avoid this.

But I will let Amir be the judge of whether we need to optimize for large clusters. If `TChunk` did not increase in size, it seems far less scary.

cherepasshka: Unfortunately brief calculation shows that on our largest cluster std::variant will cost additional +90 - + 420 gb to master memory, which is dreadful, so something like variant TAugmentedPtr will be introduced here

achulkov2: (jfyi I am also touching TAugmentedPtr in my max medium count PR, keep it in mind for the unlucky person who has to solve conflicts)

anyway, there is a free index in TAugementedPtr used for replicas, there quite a few bits to steal there
or you can even steal a few bits from the alignment part, right now less than 8 states are used there :)

cherepasshka: Yes, I am planning to use space that is supposed to be for extra index. But since we need to store variant type (TChunkLocation or TMedium) it is impossible to reuse TAugmentedPtr and I will create similar variant class

yt/yt/server/master/chunk_server/chunk.cpp
```diff
@@ -280,11 +280,41 @@ bool TChunk::HasParents() const
...
+    TStoredChunkReplicaPtrWithReplicaInfo replica,
+    const TMedium* medium,
     bool approved)
 {
     auto* data = MutableReplicasData();
+
+    if (medium->IsOffshore()) {
+        if (!replica.IsMediumPtr()) {
+            YT_LOG_ALERT(
+                "Attempted to add offshore medium to chunk location replica, ignored"
```
raundn: "...add medium to ... replica" sounds weird.

yt/yt/client/node_tracker_client/public.h
```diff
@@ -42,8 +42,12 @@ DEFINE_ENUM(EAddressType,
 );
 
 YT_DEFINE_STRONG_TYPEDEF(TNodeId, ui32);
-constexpr TNodeId InvalidNodeId = TNodeId(0);
+// Node ids of physical nodes are in the range [1, MaxRealNodeId].
```
raundn: "Physical" is just another synonym for "real" here, I doesn't really supply any additional information. If anything, it makes things blurrier: are contrainers physical? are virtualized machines?

It's better to explicitly state that "real" means "non-sentinel".

yt/yt/server/master/chunk_server/stored_chunk_replica.h
```diff
@@ -0,0 +1,103 @@
...
+
+    TStoredChunkReplicaPtrWithReplicaInfo(const TStoredChunkReplicaPtrWithReplicaInfo& other) = default;
+    TStoredChunkReplicaPtrWithReplicaInfo(TStoredChunkReplicaPtrWithReplicaInfo&& other) = default;
+
+    TStoredChunkReplicaPtrWithReplicaInfo& operator=(const TStoredChunkReplicaPtrWithReplicaInfo& other) = default;
+
+    bool HasPtr() const;
+
+    bool IsChunkLocationPtr() const;
+    bool IsMediumPtr() const;
```
raundn: This is a matter of taste but let me gently suggest removing these methods, making `GetStoredReplicaType` public and then replacing `IsChunkLocationPtr()` with `replica.GetStoredReplicaType() == EStoredReplicaType::ChunkLocation`. I won't insist on this but I think that'd make the code a bit more readable and extensible. (This would also require turning `As*Ptr` into a template method taking enum as its template parameter.)

yt/yt/server/master/chunk_server/stored_chunk_replica.h
```diff
@@ -0,0 +1,103 @@
...
+    template <class T>
+    TStoredChunkReplicaPtrWithReplicaInfo(T* ptr, int index, EChunkReplicaState replicaState = EChunkReplicaState::Generic)
+        requires ((std::is_same_v<T, TChunkLocation> || std::is_same_v<T, TMedium>));
+
+    TStoredChunkReplicaPtrWithReplicaInfo(const TStoredChunkReplicaPtrWithReplicaInfo& other) = default;
+    TStoredChunkReplicaPtrWithReplicaInfo(TStoredChunkReplicaPtrWithReplicaInfo&& other) = default;
+
+    TStoredChunkReplicaPtrWithReplicaInfo& operator=(const TStoredChunkReplicaPtrWithReplicaInfo& other) = default;
+
+    bool HasPtr() const;
```
raundn: Consider turning this into an `explicit operator bool() const`. I think it'll look more natural.

yt/yt/server/master/chunk_server/stored_chunk_replica.h
```diff
@@ -0,0 +1,103 @@
...
+
+    bool IsChunkLocationPtr() const;
+    bool IsMediumPtr() const;
+
+    TChunkLocation* AsChunkLocationPtr() const;
+    TMedium* AsMediumPtr() const;
+
+    bool operator==(TStoredChunkReplicaPtrWithReplicaInfo other) const;
```
raundn: I suspect you can `= default` this. (But you'll have to either change the type of `other` to `const&` or turn the method into a free function `friend bool operator==(TStoredChunkReplicaPtrWithReplicaInfo lhs, TStoredChunkReplicaPtrWithReplicaInfo rhs) = default`.)

yt/yt/server/master/chunk_server/stored_chunk_replica.h
```diff
@@ -0,0 +1,103 @@
...
+    bool operator<(TStoredChunkReplicaPtrWithReplicaInfo other) const;
+    bool operator<=(TStoredChunkReplicaPtrWithReplicaInfo other) const;
+    bool operator>(TStoredChunkReplicaPtrWithReplicaInfo other) const;
+    bool operator>=(TStoredChunkReplicaPtrWithReplicaInfo other) const;
```
raundn: Let's forget the past =) These four operators can be replaced with a single `operator <=>`. (The compiler will then rewrite `a < b` into `(a <=> b) < 0` etc., so you won't have to do anything at all about the four.)

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
```diff
@@ -0,0 +1,220 @@
...
+    auto thisStoredReplicaType = GetStoredReplicaType();
+    auto otherStoredReplicaType = other.GetStoredReplicaType();
+
+    auto thisReplicaIndex = GetReplicaIndex();
+    auto otherReplicaIndex = other.GetReplicaIndex();
+
+    auto thisState = GetReplicaState();
+    auto otherState = other.GetReplicaState();
+
+    return std::tuple(thisStoredReplicaType, thisReplicaIndex, thisState, GetId()) < std::tuple(otherStoredReplicaType, otherReplicaIndex, otherState, other.GetId());
```
raundn: While I'm here popularizing operator <=>, this is how you write a typical operator these days:

```cpp
if (auto cmp = GetStoredReplicaType() <=> other.GetStoredReplicaType(); cmp != 0) {
    return cmp;
}

if (auto cmp = GetReplicaIndex() <=> other.GetReplicaIndex(); cmp != 0) {
    return cmp;
}

if (auto cmp = GetReplicaState() <=> other.GetReplicaState(); cmp != 0) {
    return cmp;
}

return GetId() <=> other.GetId();
```

This is the idiomatic (and, I must say, awesome) way of doing comparison. You might, now and then, have to implement `operator <=>` for constituent parts of a class, but other than that - convenient!

yt/yt/server/master/chunk_server/stored_chunk_replica.h
```diff
@@ -0,0 +1,103 @@
...
+DEFINE_ENUM_WITH_UNDERLYING_TYPE(EStoredReplicaType, ui8,
+    ((ChunkLocation)   (0))
+    ((OffshoreMedia)   (1))
+);
+
+static_assert(
+    static_cast<int>(TEnumTraits<EStoredReplicaType>::GetMaxValue()) < (1LL << 3),
+    "Stored replica type must fit into 2 bits.");
+
+//! Custom class for variant chunk replica. It is similar to TAugmentedPtr, but due to variety of stored pointer has to be separate class.
```
raundn: "due to variety of stored pointer *types*"? Of "due to variability of the stored pointer". 

But let me suggest a simpler phrasing: 

```cpp
// Stores compact representation for |(variant(TChunkLocation*, TMedium*), replica_index, replica_state)|.
// Similar to TAugmentedPtr but not limited to a single fixed pointer type.
```

yt/yt/server/master/chunk_server/stored_chunk_replica.h
```diff
@@ -0,0 +1,103 @@
...
+
+#include <yt/yt/server/lib/misc/assert_sizeof.h>
+
+namespace NYT::NChunkServer {
+
+////////////////////////////////////////////////////////////////////////////////
+
+DEFINE_ENUM_WITH_UNDERLYING_TYPE(EStoredReplicaType, ui8,
+    ((ChunkLocation)   (0))
+    ((OffshoreMedia)   (1))
```
raundn, line 15: I was about to suggest turning this into a bit enum but since we're not planning on bit-packing multiple values of the enum into a single value, this isn't really the case for bit enums.

*However.* I think it's a good idea to static-assert that all members of this enum are distinct. Please take a look at `ENUM__VALIDATE_UNIQUE` for inspiration.

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
```diff
@@ -0,0 +1,220 @@
...
+namespace NYT::NChunkServer {
+
+using namespace NChunkClient;
+using namespace NObjectServer;
+using namespace NNodeTrackerClient;
+
+////////////////////////////////////////////////////////////////////////////////
+
+TStoredChunkReplicaPtrWithReplicaInfo::TStoredChunkReplicaPtrWithReplicaInfo()
+{ }
```
raundn: = defalt

yt/yt/server/master/chunk_server/stored_chunk_replica.h
```diff
@@ -0,0 +1,103 @@
...
+public:
+    TStoredChunkReplicaPtrWithReplicaInfo();
+    template <class T>
+    TStoredChunkReplicaPtrWithReplicaInfo(T* ptr, int index, EChunkReplicaState replicaState = EChunkReplicaState::Generic)
+        requires ((std::is_same_v<T, TChunkLocation> || std::is_same_v<T, TMedium>));
+
+    TStoredChunkReplicaPtrWithReplicaInfo(const TStoredChunkReplicaPtrWithReplicaInfo& other) = default;
+    TStoredChunkReplicaPtrWithReplicaInfo(TStoredChunkReplicaPtrWithReplicaInfo&& other) = default;
+
+    TStoredChunkReplicaPtrWithReplicaInfo& operator=(const TStoredChunkReplicaPtrWithReplicaInfo& other) = default;
```
raundn: I there's a move-constructor, there should be a move-assignment operator.

yt/yt/server/master/chunk_server/stored_chunk_replica.h
```diff
@@ -0,0 +1,103 @@
...
+    void Save(C& context) const;
+
+    template <class C>
+    void Load(C& context);
+
+    TStoredChunkReplicaPtrWithReplicaInfo ToGenericState() const;
+
+    EChunkReplicaState GetReplicaState() const;
+
+    // todo: -> ui8
```
raundn: - I'd argue there's no need in this.
- Why "todo", not "TODO"?

yt/yt/server/master/chunk_server/stored_chunk_replica.h
```diff
@@ -0,0 +1,103 @@
...
+    int GetReplicaIndex() const;
+    int GetEffectiveMediumIndex() const;
+    NNodeTrackerClient::TChunkLocationIndex GetChunkLocationIndex() const;
+    NChunkClient::TChunkLocationUuid GetLocationUuid() const;
+    NNodeTrackerClient::TNodeId GetNodeId() const;
```
raundn, line 59: I have an idea. These medium- and location-related methods constitute what I'd call a "union interface": essentially, we have two different types in one (a medium-replica and a location-replica) and these methods come from them both.

This makes for a class that's easy to misuse: every single call to `GetChunkLocationIndex` must be prefaced with an `if (replica.IsChunkLocationPtr())`.

You've probably guessed what I'm about to suggest. Let's split this class into three:
  - `TAugmentedChunkReplicaPtr` - similar to `TStoredChunkReplicaPtrWithReplicaInfo` but without location- and medium-specific methods; instead, it'll have two `As*` methods returning objects of the following classes;
  - `TAugmentedLocationChunkReplicaPtr` - location-specific methods (probably doesn't even need the save-loading)
  - `TAugmentedMediumChunkReplicaPtr` - medium-specific methods.

Of course, type checking will still have to be done but now you can fuse it with conversion methods (`As*` methods may return `optional<TAugment...>`, for example), and once you've done the conversion, it's literally impossible to call a wrong method - the compiler simply won't allow it.

yt/yt/server/master/chunk_server/stored_chunk_replica.h
```diff
@@ -0,0 +1,103 @@
...
+    friend void ToProto(ui64* protoValue, TStoredChunkReplicaPtrWithReplicaInfo value);
+};
+
+// Think twice before increasing this.
+YT_STATIC_ASSERT_SIZEOF_SANITY(TStoredChunkReplicaPtrWithReplicaInfo, 8);
+
+////////////////////////////////////////////////////////////////////////////////
+
+using TStoredChunkReplicaPtrWithReplicaInfoList = TCompactVector<TStoredChunkReplicaPtrWithReplicaInfo, TypicalReplicaCount>;
+using TChunkToStoredChunkReplicaPtrWithReplicaInfoList = THashMap<TChunkId, TErrorOr<TStoredChunkReplicaPtrWithReplicaInfoList>>;
```
raundn: This is a weirdly specific typedef. Is it really that widely used as to merit such wide a scope? If not, I'd suggest bringing it closer to usage points, making it's scope narrower.

yt/yt/server/master/chunk_server/stored_chunk_replica.h
```diff
@@ -0,0 +1,103 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+DEFINE_ENUM_WITH_UNDERLYING_TYPE(EStoredReplicaType, ui8,
+    ((ChunkLocation)   (0))
+    ((OffshoreMedia)   (1))
+);
+
+static_assert(
+    static_cast<int>(TEnumTraits<EStoredReplicaType>::GetMaxValue()) < (1LL << 3),
+    "Stored replica type must fit into 2 bits.");
```
raundn: Am I correct in my understanding that we could've made this 8-bit wide and are just being thrifty here?

yt/yt/server/master/chunk_server/stored_chunk_replica-inl.h
```diff
@@ -0,0 +1,80 @@
...
+{
+    YT_ASSERT((reinterpret_cast<uintptr_t>(ptr) & 0xffff000000000003LL) == 0);
+    YT_ASSERT(index >= 0 && index <= 0xff);
+    YT_ASSERT(static_cast<uintptr_t>(replicaState) <= 0x3);
+}
+
+template <class C>
+void TStoredChunkReplicaPtrWithReplicaInfo::Save(C& context) const
+{
+    using NYT::Save;
```
raundn: Empty line after this.

yt/yt/server/master/chunk_server/stored_chunk_replica-inl.h
```diff
@@ -0,0 +1,80 @@
...
+    Save(context, GetReplicaIndex());
+    Save(context, GetReplicaState());
+
+    switch (type) {
+        case EStoredReplicaType::ChunkLocation: {
+            SaveWith<NCellMaster::TRawNonversionedObjectPtrSerializer>(context, AsChunkLocationPtr());
+            break;
+        }
+        case EStoredReplicaType::OffshoreMedia: {
+            SaveWith<NCellMaster::TRawNonversionedObjectPtrSerializer>(context, AsMediumPtr());
```
raundn: Won't `Save(context, AsMediumPtr());` work here?

yt/yt/server/master/chunk_server/stored_chunk_replica-inl.h
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+}
+
+template <class C>
+void TStoredChunkReplicaPtrWithReplicaInfo::Load(C& context)
+{
+    using NYT::Load;
+
+    if (context.GetVersion() < NCellMaster::EMasterReign::RefactoringAroundChunkStoredReplicas) {
```
raundn: If `GetVersion()` returns EMasterReign, why does `context` has a templated type?

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+bool TStoredChunkReplicaPtrWithReplicaInfo::HasPtr() const
+{
+    return Value_ & 0x0000fffffffffffcLL;
```
raundn: Let's be safe and rewrite this as

```cpp
if (IsChunkLocationPtr()) {
    return AsChunkLocationPtr();
} else if (IsMediumPtr()) {
    return AsMediumPtr();
} else {
    return false;
}
```

Strictly speaking, C++ doesn't guarantee that bit representation of `nullptr` is all zeros. (It does, however, guarantee, that `nullptr` converts to false even if it isn't represented by all zeros.)

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
```diff
@@ -0,0 +1,220 @@
...
+{
+    switch (value.GetStoredReplicaType()) {
+        case EStoredReplicaType::ChunkLocation: {
+            FormatValue(builder, TChunkLocationPtrWithReplicaInfo(value.AsChunkLocationPtr(), value.GetReplicaIndex()), spec);
+            break;
+        }
+        case EStoredReplicaType::OffshoreMedia:
+        {
+            FormatValue(builder, TMediumPtrWithReplicaInfo(value.AsMediumPtr(), value.GetReplicaIndex()), spec);
+        }
```
raundn, line 221: break;

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
```diff
@@ -0,0 +1,220 @@
...
+//! Serializes node id, replica index, medium index.
+void ToProto(ui64* protoValue, TStoredChunkReplicaPtrWithReplicaInfo value)
+{
+    switch (value.GetStoredReplicaType()) {
+        case EStoredReplicaType::ChunkLocation: {
+            ToProto(protoValue, TChunkLocationPtrWithReplicaInfo(value.AsChunkLocationPtr(), value.GetReplicaIndex()));
+            break;
+        }
+        case EStoredReplicaType::OffshoreMedia: {
+            ToProto(protoValue, TMediumPtrWithReplicaInfo(value.AsMediumPtr(), value.GetReplicaIndex()));
```
raundn: break;

yt/yt/server/master/chunk_server/chunk.cpp
raundn: Oh, and this is a matter of taste but I would've written `if (replica.IsLocationPtr())` because this is what alert text below is explicitly concerned with.

yt/yt/server/master/chunk_server/chunk_placement.cpp
```diff
@@ -627,11 +632,22 @@ TChunkLocation* TChunkPlacement::GetRemovalTarget(
     TCompactFlatMap<const TDataCenter*, i8, 4> perDataCenterCounters;
 
     for (auto replica : replicas) {
-        if (replica.GetPtr()->GetEffectiveMediumIndex() != mediumIndex) {
+        if (!replica.IsChunkLocationPtr()) {
+            YT_LOG_ALERT(
+                "Non-chunk location stored replica was found during processing removal targets for chunk, ignored "
+                "(ChunkId:%v, ReplicaMediumIndex: %v, ReplicaIndex: %v)",
```
raundn: A space is missing after :

yt/yt/server/master/chunk_server/chunk_replicator.cpp
```diff
@@ -537,15 +537,30 @@ TCompactMediumMap<EChunkStatus> TChunkReplicator::ComputeChunkStatuses(
...
 {
     return ComputeChunkStatistics(chunk, replicas).Status;
 }
 
 TChunkReplicator::TChunkStatistics TChunkReplicator::ComputeChunkStatistics(
     const TChunk* chunk,
-    const TChunkLocationPtrWithReplicaInfoList& replicas)
+    const TStoredChunkReplicaPtrWithReplicaInfoList& replicas)
 {
+    auto offshoreReplicaIt = std::find_if(replicas.begin(), replicas.end(), [] (const auto& replica) { return replica.IsMediumPtr(); });
```
raundn: Let's hide this under `if (chunk->IsErasure())` to save ourselves some work. This will have the additional benefit of making to possible to traverse replicas only once.

yt/yt/server/master/chunk_server/chunk_replicator.cpp
```diff
@@ -1136,28 +1167,21 @@ TChunkReplicator::TChunkStatistics TChunkReplicator::ComputeRegularChunkStatisti
...
     for (auto entry : replication) {
         auto mediumIndex = entry.GetMediumIndex();
         auto* medium = chunkManager->FindMediumByIndex(mediumIndex);
         YT_VERIFY(IsObjectAlive(medium));
 
-        // TODO(gritukan): Check replica presence here when
-        // chunk will store offshore replicas.
-        // For now, we just ignore such media.
-        if (medium->IsOffshore()) {
-            continue;
```
raundn, line 1156: If I understand correctly, this makes it so that `ComputeRegularChunkStatisticsForMedium` is now called for offshore media, among others. And I don't think that method is ready for offshore media. Since this PR is more or less about refactoring, I recommend bringing this `continue` back and postponing properly implementing replicator support for offshore media until a later PR.

yt/yt/server/master/chunk_server/chunk_replicator.cpp
```diff
@@ -2425,6 +2452,15 @@ void TChunkReplicator::RefreshChunk(
 
     auto durabilityRequired = IsDurabilityRequired(chunk, chunkReplicas);
 
+    auto alertNonChunkLocationReplica = [&] (const auto& replica) {
+        YT_LOG_ALERT(
+            "Non-chunk location stored replica was found during refresh of chunk "
+            "(ChunkId:%v, ReplicaMediumIndex: %v, ReplicaIndex: %v)",
```
raundn: A space is missing after the first colon.

yt/yt/server/master/chunk_server/chunk_requisition.cpp
```diff
@@ -376,14 +376,23 @@ void TChunkRequisition::Load(NCellMaster::TLoadContext& context)
...
-        replicationPolicy.SetReplicationFactor(replicationFactor);
+        auto* medium = chunkManager->GetMediumByIndex(entry.MediumIndex);
+        if (isErasureChunk || medium->IsOffshore()) {
+            auto& replicationPolicy = entry.ReplicationPolicy;
+            if (!replicationPolicy) {
+                YT_LOG_ALERT(
+                    "Medium is disabled for requisition entry, ignore replication factor set (RequisitionEntry: %v)",
```
raundn: Let's log the full requisition (`*this`), otherwise it'll be hard to find the problem.

yt/yt/server/master/chunk_server/stored_chunk_replica.h
raundn: Okay, once I've read class methods, I realize now that there's some common functionality to both cases. So, a new proposal: let's make `TAugmentedChunkReplicaPtr` a base, and `TAugmentedLocationChunkReplicaPtr` and `TAugmentedMediumChunkReplicaPtr` - its subclasses.

The trick here is that we won't have to use virtual functions (because the base actually knows its real type and can implement methods that handle all variants). Which is good since we can't afford increasing the size of the class. 

Other than that, the plan still seems sound to me.

yt/yt/server/master/node_tracker_server/node_tracker.cpp
cherepasshka: Yes, of course. Protection happens in GenerateNodeId, a few rows upper: https://github.com/ytsaurus/ytsaurus/pull/1426/files#diff-82ef8535402eb4954e559d8ea44d1c8052e200b4d3cb36bc9c567b903c21bcb4R1119

yt/yt/server/master/chunk_server/stored_chunk_replica.h
cherepasshka: DEFINE_ENUM_WITH_UNDERLYING_TYPE already includes ENUM__VALIDATE_UNIQUE:
https://github.com/ytsaurus/ytsaurus/blob/main/library/cpp/yt/misc/enum.h#L112

yt/yt/server/master/chunk_server/chunk.cpp
```diff
@@ -280,11 +280,41 @@ bool TChunk::HasParents() const
...
+        if (replica.GetEffectiveMediumIndex() != medium->GetIndex()) {
+            YT_LOG_ALERT(
+                "Offshore replica medium index differs from medium index, ignored"
+                "(ChunkId: %v, OffshoreReplicaMediumIndex: %v, MediumIndex: %v)",
+                GetId(),
+                replica.GetEffectiveMediumIndex(),
+                medium->GetIndex());
+            return;
+        }
+        // TODO(cherepashka): allow once offshore media is implemented.
```
raundn: *are* implemented

yt/yt/server/master/chunk_server/stored_chunk_replica.h
```diff
@@ -0,0 +1,128 @@
...
+    TAugmentedStoredChunkReplicaPtr(TAugmentedStoredChunkReplicaPtr&& other) = default;
+
+    TAugmentedStoredChunkReplicaPtr& operator=(const TAugmentedStoredChunkReplicaPtr& other) = default;
+    TAugmentedStoredChunkReplicaPtr& operator=(TAugmentedStoredChunkReplicaPtr&& other) = default;
+
+    explicit operator bool() const;
+
+    EStoredReplicaType GetStoredReplicaType() const;
+    template <EStoredReplicaType type>
+    const typename TStoredReplicaTraits<type>::Type* As() const;
```
raundn, line 47: This can also be done with `std::conditional` but the traits approach works, too.

yt/yt/server/master/chunk_server/stored_chunk_replica-inl.h
```diff
@@ -0,0 +1,58 @@
...
+}
+
+template <EStoredReplicaType type>
+const typename TStoredReplicaTraits<type>::Type* TAugmentedStoredChunkReplicaPtr::As() const
+{
+    if (GetStoredReplicaType() != type) {
+        return nullptr;
+    }
+
+    YT_ASSERT(this);
```
raundn, line 46: Is this really necessary? How does one call a method via a null object pointer, anyway? =)

Or did you mean `YT_ASSERT(*this)`?

yt/yt/server/master/chunk_server/stored_chunk_replica-inl.h
```diff
@@ -0,0 +1,58 @@
...
+
+    if constexpr (type == EStoredReplicaType::ChunkLocation) {
+        return static_cast<const TAugmentedLocationChunkReplicaPtr*>(this);
+    }
+    else if constexpr (type == EStoredReplicaType::OffshoreMedia) {
+        return static_cast<const TAugmentedMediumChunkReplicaPtr*>(this);
+    }
+
+    Y_UNREACHABLE();
```
raundn: static_assert?

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
```diff
@@ -0,0 +1,257 @@
...
+    auto thisStoredReplicaType = GetStoredReplicaType();
+    auto otherStoredReplicaType = other.GetStoredReplicaType();
+
+    auto thisReplicaIndex = GetReplicaIndex();
+    auto otherReplicaIndex = other.GetReplicaIndex();
+
+    auto thisState = GetReplicaState();
+    auto otherState = other.GetReplicaState();
+
+    return std::tuple(thisStoredReplicaType, thisReplicaIndex, thisState, GetId()) <=> std::tuple(otherStoredReplicaType, otherReplicaIndex, otherState, other.GetId());
```
raundn: The problem with the tuple approach (as opposed to a chain of if-returns I suggested [here](https://github.com/ytsaurus/ytsaurus/pull/1426#discussion_r2424876068)) is the extra work: first you need to do (probably unnecessary) bitwise arithmetic and than construct (and later destroy) the tuples. And the resulting code isn't much shorter.

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
```diff
@@ -0,0 +1,257 @@
...
+            return As<EStoredReplicaType::ChunkLocation>()->AsChunkLocationPtr()->GetEffectiveMediumIndex();
+        case EStoredReplicaType::OffshoreMedia:
+            return As<EStoredReplicaType::OffshoreMedia>()->AsMediumPtr()->GetIndex();
+    }
+}
+
+TChunkLocationUuid TAugmentedStoredChunkReplicaPtr::GetLocationUuid() const
+{
+    switch (GetStoredReplicaType()) {
+    case EStoredReplicaType::ChunkLocation: {
```
raundn: Something's off with indentation here...

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
```diff
@@ -0,0 +1,257 @@
...
+        auto* location = As<EStoredReplicaType::ChunkLocation>()->AsChunkLocationPtr();
+        if (!IsObjectAlive(location)) {
+            return InvalidChunkLocationUuid;
+        }
+        return location->GetUuid();
+    }
+    case EStoredReplicaType::OffshoreMedia:
+        return InvalidChunkLocationUuid;
```
raundn: Hm... Does `GetLocationUuid` get used for offshore stored replicas? Because if not, the methods should be moved down to `TAugmentedLocationChunkReplicaPtr`.

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
```diff
@@ -0,0 +1,257 @@
...
+                return InvalidNodeId;
+            }
+            auto node = location->GetNode();
+            if (!IsObjectAlive(node)) {
+                return InvalidNodeId;
+            }
+            return node->GetId();
+        }
+        case EStoredReplicaType::OffshoreMedia:
+            return OffshoreNodeId;
```
raundn, line 116: Same question here.

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
```diff
@@ -0,0 +1,257 @@
...
+    }
+}
+
+NObjectClient::TObjectId TAugmentedStoredChunkReplicaPtr::GetId() const
+{
+    switch (GetStoredReplicaType()) {
+        case EStoredReplicaType::ChunkLocation: {
+            auto locationReplica = As<EStoredReplicaType::ChunkLocation>();
+            auto* location = locationReplica->AsChunkLocationPtr();
+            return IsObjectAlive(location) ? location->GetId() : NullObjectId;
```
raundn, line 175: There's a helper called `GetObjectId` for this.

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
```diff
@@ -0,0 +1,257 @@
...
+        }
+        case EStoredReplicaType::OffshoreMedia: {
+            auto* medium = As<EStoredReplicaType::OffshoreMedia>()->AsMediumPtr();
+            return IsObjectAlive(medium) ? medium->GetId() : NullObjectId;
```
raundn, line 179: Same here.

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
```diff
@@ -0,0 +1,257 @@
...
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+TChunkLocationIndex TAugmentedLocationChunkReplicaPtr::GetChunkLocationIndex() const
+{
+    YT_ASSERT(GetStoredReplicaType() == EStoredReplicaType::ChunkLocation);
+
+    auto* location = AsChunkLocationPtr();
```
raundn, line 188: `AsChunkLocationPtr` already contains the same assert as above, so the latter can be removed.

yt/yt/server/master/chunk_server/chunk_proxy.cpp
```diff
@@ -288,7 +288,8 @@ class TChunkProxy
         const TChunkLocation* location,
         int replicaIndex,
         EChunkReplicaState replicaState,
-        int mediumIndex)
+        int mediumIndex,
+        bool offshoreMedium)
```
raundn: mediumIsOffshore

yt/yt/server/master/chunk_server/chunk_replica.cpp
```diff
@@ -116,6 +117,39 @@ void ToProto(ui32* protoValue, TChunkLocationPtrWithReplicaInfo value)
...
+    THROW_ERROR_EXCEPTION_UNLESS(value.GetPtr()->IsOffshore(),
```
raundn: How do you look at adding an YT_LOG_ALERT here? First of all, to my understanding, an exception here always signifies a bug. Second of all, I'm not sure all usages (indirect ones included) of this `ToProto` are ready to it throwing. An alert, at least, will keep us notified.

yt/yt/server/master/chunk_server/chunk_replicator.cpp
```diff
@@ -537,15 +537,32 @@ TCompactMediumMap<EChunkStatus> TChunkReplicator::ComputeChunkStatuses(
...
 TChunkReplicator::TChunkStatistics TChunkReplicator::ComputeChunkStatistics(
     const TChunk* chunk,
-    const TChunkLocationPtrWithReplicaInfoList& replicas)
+    const TStoredChunkReplicaList& replicas)
 {
+    if (chunk->IsErasure()) {
+        auto offshoreReplicaIt = std::find_if(replicas.begin(), replicas.end(), [] (const auto& replica) { return replica.GetStoredReplicaType() == EStoredReplicaType::OffshoreMedia; });
+        if (offshoreReplicaIt != replicas.end()) {
+            TStoredChunkReplicaList offshoreReplicas;
+            for (const auto& replica: replicas) {
```
raundn: Why are we iterating through `replicas` twice here? (First via `find_if` and then via `for`.) Why not get rid of `find_if` and start collecting `offshoreReplicas` right away? It's a compact vector, so even in the case of there being no offshore replicas it's virtually free.

yt/yt/server/master/chunk_server/chunk_replicator.cpp
```diff
@@ -1076,9 +1106,15 @@ TChunkReplicator::TChunkStatistics TChunkReplicator::ComputeRegularChunkStatisti
     }
 
     for (auto replica : replicas) {
-        auto* chunkLocation = replica.GetPtr();
+        auto* locationReplica = replica.As<EStoredReplicaType::ChunkLocation>();
+        if (!locationReplica) {
+            auto mediumIndex = replica.GetEffectiveMediumIndex();
+            ++replicaCount[mediumIndex];
```
raundn: I'd recommend leaving this change for later, when we get to actually supporting offshore replicas in the replicator. At the moment, this is potentially dangerous: the code below may look at `replicaCount` and make certain decisions (to not replicate a chunk, for example) that the rest of the code is not prepared for.

yt/yt/server/master/chunk_server/chunk_replicator.cpp
```diff
@@ -2386,17 +2426,17 @@ void TChunkReplicator::ScheduleRepairJobs(IJobSchedulingContext* context)
...
         return;
     }
 
-    if (ephemeralChunk->IsForeign()) {
+    if (!ephemeralChunk->IsConfirmed()) {
         return;
     }
 
-    if (!IsObjectAlive(ephemeralChunk)) {
+    if (ephemeralChunk->IsForeign()) {
```
raundn: Mmm, bumping `IsObjectAlive` check up was a good idea but I think it should be followed by the `IsForeign` check. (Because calling `IsConfirmed` for a foreign chunk is semantically incorrect.)

raundn (approved):
Let's fix the remaining issues (they're minor) and merge this.

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
cherepasshka: GetObjectId doesn't check if object is alive (has positive reference count)

yt/yt/server/master/chunk_server/stored_chunk_replica.cpp
cherepasshka: Yes, for example the method is used to sort replicas by (replicaIndex, nodeId, locationUuid). It seems to me that it is better to leave this method inside the general class than to move it into a descendant and check each time at the place of use type of the replica and handle medium case with the constants (InvalidLocationUuid, OffshoreNodeId)

yt/yt/server/master/chunk_server/stored_chunk_replica-inl.h
cherepasshka: It is preventive measure, same as https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/server/master/object_server/object-inl.h#L203

cherepasshka:
Shipped!

cherepasshka:
Shipped!

cherepasshka:
Shipped!

PR #1427, opened 2025-09-08
Fix read results with yson fields by reference.

Tony-Romanov:
Type: fix
Component: query-tracker

### DIFF
--- a/yt/yql/tests/agent/test_simple.py
+++ b/yt/yql/tests/agent/test_simple.py
@@ -1000,7 +1000,7 @@ class TestQueriesYqlResultTruncation(TestQueriesYqlSimpleBase):
         rows = [{"value": ''.join(['d' for _ in range(value_size)])} for i in range(22)]
         write_table("//tmp/t", rows)
 
-        expected = [{"value": ''.join(['d' for _ in range(value_size)]), "check":""}]
+        expected = [{"value": ''.join(['d' for _ in range(value_size)]), "check": ""}]
         q = self.start_query("yql", "select value, ('[\"\"]'y).0 as `check` from `//tmp/t`")
         q.track()
         assert q.get()["result_count"] == 1


yt/yt_proto/yt/client/api/rpc_proxy/proto/api_service.proto
```diff
@@ -2851,6 +2851,7 @@ message TReqReadTable
     optional ERowsetFormat desired_rowset_format = 8 [default = RF_YT_WIRE];
     optional ERowsetFormat arrow_fallback_rowset_format = 10 [default = RF_YT_WIRE];
     optional bytes format = 9; // YSON-serialized TFormat
+    optional bool enable_any_unpacking = 11 [default = true];
```
Krock21, line 2860: It may change default behaviour of read_table in rpc proxy, am I right?

It may break users who rely on default=false

Tony-Romanov: enable_any_unpacking = true, it's default behavior.

Tony-Romanov: These changes should have been [here](https://github.com/ytsaurus/ytsaurus/pull/1390), but I lost them when creating PR.

Krock21: Indeed, https://github.com/ytsaurus/ytsaurus/blob/fbc718dcd2c8631586ebe069db5cd95dc05be4ea/yt/yt/client/table_client/config.cpp#L431

yt/yql/tests/agent/test_simple.py
```diff
@@ -987,6 +987,26 @@ def test_line_above_limit(self, query_tracker, yql_agent):
         assert "Failed to read resulting rowset. Try using INSERT INTO to save result" in str(result["error"])
         assert "full_result" not in result
 
+    @authors("a-romanov")
+    @pytest.mark.timeout(600)
```
Krock21: Why such timeout? It shouldn't take that long
I suggest leaving a default timeout or putting 180, as in the test above

It may work slower locally and adding `--build relwithdebinfo` should help (`ya make --build relwithdebinfo -A -F '*test_simple*' yt/yql/tests/agent`)

yt/yql/tests/agent/ya.make
```diff
@@ -22,6 +22,7 @@ DEPENDS(
     yql/essentials/udfs/common/file
     yql/essentials/udfs/common/python/python3_small
     yql/essentials/udfs/common/string
+    yql/essentials/udfs/common/yson2
```
Krock21, line 25: Why do we need it?
At first, I though for `select ('[\"\"]'y).0` to work, but then I saw that we already use it in existing tests

Tony-Romanov: Yes, it's for `('[\"\"]'y).0`. For `.0` to be precise.
I don't know how to write yson literal with empty string inside without this crutch.

Krisha11:
To bring this code to production you will also need to cherry-pick proxy change in stable branches 25.1, 25.2, 25.3...

Krisha11:
```yt/yql/tests/agent/test_simple.py:1003: [E231] missing whitespace after ':'```

PR #1431, opened 2025-09-09
chyt/controller: detect changes in secrets and cluster connection

koct9i:
Restart operation when secrets or cluster connection changes.

Present mechanism is unreliable and cannot track changes in secret files.

Add into oplet persistent state snapshot of controller state with
hashes for important fields of cluster connection and secrets.

### DIFF
--- a/yt/chyt/controller/internal/agent/agent.go
+++ b/yt/chyt/controller/internal/agent/agent.go
@@ -334,16 +334,8 @@ func (a *Agent) pass() {
 }
 
 func (a *Agent) updateControllerState() error {
-	changed, err := a.controller.UpdateState()
-	if err != nil {
-		return err
-	}
-	if changed {
-		for _, oplet := range a.aliasToOp {
-			oplet.SetPendingRestart("controller's state changed")
-		}
-	}
-	return nil
+	_, err := a.controller.UpdateState()
+	return err
 }
 
 func (a *Agent) initializeFromCypress() error {
--- a/yt/chyt/controller/internal/sleep/controller.go
+++ b/yt/chyt/controller/internal/sleep/controller.go
@@ -21,6 +21,10 @@ type Controller struct {
 	controllerParameter string
 }
 
+type controllerSnapshot struct {
+	ControllerParameter string `yson:"controller_parameter,omitempty"`
+}
+
 func (c Controller) Prepare(ctx context.Context, oplet *strawberry.Oplet) (
 	spec map[string]any, description map[string]any, annotations map[string]any, runAsUser bool, err error) {
 	err = nil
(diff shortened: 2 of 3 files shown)


yt/chyt/controller/internal/agent/agent.go
```diff
@@ -288,17 +288,27 @@ func (a *Agent) processOplets() {
 }
 
 func (a *Agent) pass() {
+	var err error
```
buyval01: Let's rename this variable to `passErr`
IMHO, if statements like `if err := ...` are more common, and in such situation it can lead to errors that are difficult to catch during debugging.

yt/chyt/controller/internal/strawberry/speclet.go
```diff
@@ -17,9 +17,10 @@ type RestartRequiredOptions struct {
 
 type Speclet struct {
 	RestartRequiredOptions
-	Active                 *bool   `yson:"active"`
-	Family                 *string `yson:"family"`
-	RestartOnSpecletChange *bool   `yson:"restart_on_speclet_change"`
+	Active                    *bool   `yson:"active"`
+	Family                    *string `yson:"family"`
+	RestartOnControllerChange *bool   `yson:"restart_on_controller_change"`
```
buyval01, line 22: What is the purpose of new option and storing controller snapshot in persistent state if oplets will be unconditionally restarted due to UpdateState controller method?

https://github.com/ytsaurus/ytsaurus/pull/1431/files#diff-76d0bdb68902272742265744c448afed1132a729838871cd5a05c38c1b3c587dR372

yt/chyt/controller/internal/agent/agent.go
koct9i: ok.

modern linters detect such errors with ease

yt/chyt/controller/internal/strawberry/speclet.go
koct9i: present logic with "pending restart" flag from update state is non-persistent, thus unstable and racy, and should be removed

https://github.com/ytsaurus/ytsaurus/blob/001889ee4eabca34fbcd5ce03ae76c367d27b7a8/yt/chyt/controller/internal/strawberry/oplet.go#L154

https://github.com/ytsaurus/ytsaurus/blob/001889ee4eabca34fbcd5ce03ae76c367d27b7a8/yt/chyt/controller/internal/strawberry/oplet.go#L199

koct9i: And it is nice to control this feature in per-instance manner.

koct9i: And we could remove "pending-restart" right now.
I've added patch for that.

PR #1432, opened 2025-09-10
Fix weird error message on CREATE TABLE

Tony-Romanov:
Type: fix
Component: query-tracker

And first small step to support DDL in YT.

Error before:
`tools/enum_parser/enum_serialization_runtime/enum_runtime.cpp:70: Key 'columns' not found in enum NYql::EYtSettingType. Valid options are: 'initial', 'infer_scheme', ...`
Error after:
`Error: CREATE TABLE is not supported yet.`


yt/yql/providers/yt/provider/yql_yt_op_settings.cpp
```diff
@@ -946,6 +979,12 @@ bool ValidateSettings(const TExprNode& settingsNode, EYtSettingTypes accepted, T
             }
             break;
         }
+        case EYtSettingType::Actions:
+        case EYtSettingType::PrimaryKey: {
+            ctx.AddError(TIssue(ctx.GetPosition(nameNode->Pos()), TStringBuilder()
+                << "Feature '" << nameNode->Content() << "' isn't supported."));
```
nepal, line 985: Let's make it even more clear for user - 'CREATE TABLE is not supported yet'

Tony-Romanov: Wrong place. You will not see "Feature  isn't supported." in normal case.

PR #1453, opened 2025-09-16
yt/yt/core/threading: fix signal handler stack for 64KiB page page

koct9i:
Round up stack to page size for installing guard pages.


Kontakter:
Shipped!

PR #1454, opened 2025-09-16
yt/yt/core/misc/isa_crc64/unittests: fix for arm64

koct9i:
Function crc64_yt_norm_by8() is implemented only for x86-64.


p4vook:
Shipped!

p4vook:
There is a space before EOL and tab instead of spaces, could you please fix it?

UPD: apparently that was even before the PR. I'll fix it myself and import.

PR #1458, opened 2025-09-23
yt/yt/library/s3: fix tls config

koct9i:
Pass "useTls" as separate flag. Null tls-config doesn't mean "http".
Use built-in CA bundle is config is not provided like https client.
And pass host name into dialer context for certificate verification.

Fixes: 54f81642716 ("add ssl config for s3 storage commit_hash:0aa02bf106b7efa521fe2e117f3f61f93d3c7f4f")

### DIFF
--- a/yt/yt/library/s3/http.cpp
+++ b/yt/yt/library/s3/http.cpp
@@ -220,7 +220,11 @@ namespace {
 NYT::NCrypto::TSslContextPtr CreateSslContext(const NYT::NCrypto::TSslContextConfigPtr& config, NYT::NCrypto::TCertificatePathResolver pathResolver = nullptr)
 {
     auto sslContext = New<NYT::NCrypto::TSslContext>();
-    sslContext->ApplyConfig(config, std::move(pathResolver));
+    if (config) {
+        sslContext->ApplyConfig(config, std::move(pathResolver));
+    } else {
+        sslContext->UseBuiltinOpenSslX509Store();
+    }
     sslContext->Commit();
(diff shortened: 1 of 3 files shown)


yt/yt/library/s3/http.cpp
```diff
@@ -234,12 +238,13 @@ class THttpClient
...
         NHttp::TClientConfigPtr config,
         TNetworkAddress address,
+        bool useTls,
         const NYT::NCrypto::TSslContextConfigPtr& sslContextConfig,
         IPollerPtr poller,
         IInvokerPtr invoker)
         : Config_(config)
         , Address_(std::move(address))
-        , Dialer_(sslContextConfig
+        , Dialer_(useTls
```
Kontakter, line 247: Could clarify why the separate flag needed? (instead of nullability of sslContextConfig)

Kontakter:
There are some conflicts – resolve it, please

yt/yt/library/s3/http.cpp
koct9i: NULL tls-config doesn't mean "http".
For code in https client this means use buit-in cert-storage.

I'll redesign this later. There is more than one issue.
For now it is better to keep logic in all https pieces to keep everybody who looks into this code sane.

koct9i: Also it would be nice to somehow review changes into all code related to TLS and HTTPS.

Kontakter:
Shipped!

PR #1459, opened 2025-09-24
Add S3Proxy to test S3 functionality

pavel-bash:
# Description

This PR adds a recipe to test S3-related functionality. Right now, if somebody wants to launch the existing `yt/yt/library/s3/unittests/client_ut.cpp` (and other S3 tests which will most probably be added in future), they need to launch an S3 instance next to the tests manually, like `minio` or the same `s3proxy`.

With this PR this process is automatized - `s3proxy` program is launched before the tests start, finishes when they end and it stores its logs in the output path of the tests so one can check it to debug the tests.

The `s3proxy` binary is brought into the repository for the ease of access. We think it's the best way to manage this dependency. The only prerequisite to run it is Java, which is also makes it suitable for Docker-less environments.


pavel-bash:
@Kontakter  This PR is ready to be reviewed. The tests are executed successfully.

yt/yt/tests/local_s3_recipe/ya.make
```diff
@@ -4,7 +4,13 @@ SRCDIR(
     yt/yt/tests/local_s3_recipe
 )
 
+PY_SRCS(__main__.py)
+
 PEERDIR(
+    yt/python/yt/test_helpers
+    library/python/testing/recipe
```
Kontakter, line 12: Add a newline between `yt/python/yt/test_helpers` and `library/python/testing/recipe`

yt/yt/tests/local_s3_recipe/__main__.py
```diff
@@ -0,0 +1,107 @@
...
+jclouds.filesystem.basedir={}
+jclouds.identity=
+jclouds.credential=
+"""
+
+def wait_for_s3_start(endpoint_url):
+    def check_s3_reachable():
+        s3_client = None
+        try:
+            s3_client = boto3.client('s3')
```
Kontakter: Use `"` instead of `'`

yt/yt/tests/local_s3_recipe/__main__.py
```diff
@@ -0,0 +1,107 @@
...
+
+def start(_: List[str]) -> None:
+    base_path = os.path.join(get_source_root(), "yt/yt/tests/local_s3_recipe")
+
+    # Retrieve the necessary environment variables' values.
+    endpoint_url = os.environ["AWS_ENDPOINT_URL"]
+    access_key_id = os.environ["AWS_ACCESS_KEY_ID"]
+    secret_access_key = os.environ["AWS_SECRET_ACCESS_KEY"]
+    if endpoint_url is None or access_key_id is None or secret_access_key is None:
+        raise Exception("Failed to start s3proxy: necessary environment variables are not set")
```
Kontakter: I propose to improve this code a little bit:
```python
def get_env_value(key):
    value = os.environ[key]
    if value is None:
        raise RuntimeError(f"Environment variable {key} is required to be set to start s3proxy")
    return value
...
endpoint_url = get_env_value("AWS_ENDPOINT_URL")
...
```

yt/yt/tests/local_s3_recipe/__main__.py
```diff
@@ -0,0 +1,107 @@
...
+    ]
+    s3 = subprocess.Popen(s3_args, stdout=S3_LOG_FILE, text=True, env={"LOG_LEVEL": "debug"})
+    with open(S3_PID_FILE_NAME, "w") as f:
+        print(s3.pid, file=f)
+
+    wait_for_s3_start(endpoint_url)
+
+
+def stop(_: List[str]) -> None:
+    if S3_LOG_FILE is not None:
```
Kontakter: How is it suppose to work?

As far as I know, the recipe binary isn't a long running binary, I mean that this machinery just executes `./recipe_bin start` and when tests pass executes `./recipe_bin stop`

pavel-bash: Looks like you're correct, yes. There's almost no documentation about the recipes of ya make available open-source, so I'm not sure how it works.

I think we don't need to close the file in the recipe then, as it already works correctly.

pavel-bash:
@Kontakter fixed your comments.

Kontakter:
Please, remove the file from PR [yt/yt/tests/local_s3_recipe/bin/s3proxy](https://github.com/ytsaurus/ytsaurus/pull/1459/files#diff-f21f2f655f6f23caee0d344f3a1140b0fb0ad1a0ed339d6d825ed45499d04115)

pavel-bash:
Converting this PR to draft until the solution with `s3proxy` is found.

pavel-bash:
We have decided to download the binary in the recipe itself for now. When it's added to the public S3 or `contrib`, we'll do it better.

@Kontakter ready for review again.

Kontakter:
Shipped!

PR #1461, opened 2025-09-25
Add empty target to bring contrib/python/boto3 to the project

pavel-bash:
# Description

In order to properly check #1459, we need to bring the `contrib/python/boto3` to the project. This PR adds an empty target with the specified dependency to fetch it.


Kontakter:
Shipped!

Kontakter:
Shipped!

PR #1464, opened 2025-09-25
Add resource_limits to exec_attributes of a job

Krock21:
Right now there is no simple way to understand what resources the job requested, and it gets even more complicated around memory_limit (there are at least 3 memory limits with different meanings)

This PR adds resource_limits to exec_attributes of a job

Feel free to propose a better solution on how ytsaurus users can understand what resource_limits are for a job

---
>
> * Changelog entry
>               # cpp-sdk/go-sdk/cms/excel/cron/microservices (Select one value, example: `Component: scheduler`)

* Changelog entry
Type:
Component:

TODO IF NEEDED

### DIFF
--- a/yt/yt/server/controller_agent/controllers/operation_controller_detail.cpp
+++ b/yt/yt/server/controller_agent/controllers/operation_controller_detail.cpp
@@ -10380,6 +10380,8 @@ void TOperationControllerBase::InitUserJobSpec(
         joblet->EstimatedResourceUsage.GetJobProxyMemory() * joblet->JobProxyMemoryReserveFactor.value());
 
     if (Options_->SetSlotContainerMemoryLimit) {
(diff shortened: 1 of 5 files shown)


Kontakter:
I am not happy with this solution and have following considerations:

1. There are [fields](https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/server/node/exec_node/job_info.cpp#L70-L72) that describe resources of job and that used in orchid, it is better to have some symmetry
2. I think that the resources should not be part of `exec_attributes`, because `exec_attributes` is used for describing the specific things about job execution (by specific I mean such attributes, that would be known only at startup time and may depend on the node)
3. CPU is not a static, it may change in time due to enabled cpu monitor

Krock21:
We discussed and agreed that statistics is a better place for this information, and it already contains `memory_reserve`

PR #1468, opened 2025-09-30
yt/docker: cleanup and add deadsnakes ppa repository manually again

koct9i:
Tool "add-apt-repository" buggy and breaks when IPv6 connectivity is flawed.
All that it does could be written in new lines.

Get ppa repository public key using "ADD" to simplify workflow.

Use http for "ppa.launchpadcontent.net" - "ca-certificates" is not
installed yet when "apt update" runs. Anyway https is not required
for apt repositories because chain of trust is signed with gpg key.

Install python in "base" which is only successor of "ubuntu-base".

### DIFF
--- a/yt/docker/ytsaurus/Dockerfile
+++ b/yt/docker/ytsaurus/Dockerfile
@@ -1,5 +1,9 @@
 # Arguments used in FROM statements need to be declared before the first FROM statement.
 
+ARG UBUNTU_CODENAME=noble
+ARG UBUNTU_IMAGE=ubuntu:${UBUNTU_CODENAME}
+ARG APT_MIRROR="http://archive.ubuntu.com/"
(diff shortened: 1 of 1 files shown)


koct9i:
I'm still seeing some issues (probably unrelated)

edit: yeah, that was unrelated issue

koct9i:
Issue with software-properties-common looks like this
https://gist.github.com/koct9i/fab18fff2a466dbb07daf94a68c27b38

it installs 123 packages (40 MiB in archive) to write couple lines of config
but dies with "TimeoutError: [Errno 110] Connection timed out"
because pythonsh crappy code failed to handle ipv4/ipv6 dual-stack properly

ilyaibraev:
LGTM

PR #1484, opened 2025-10-21
Add information about `preserve_expiration_time` parameter in move operation

sheinid:
Add information about `preserve_expiration_time` parameter in move operation to docs


ilyaibraev:
Hi! Could you please make an edit [here](https://github.com/ytsaurus/ytsaurus/blob/main/yt/docs/ru/_includes/api/commands.md?plain=1#L585) as well?

ilyaibraev:
Shipped!

Issue #1483, opened 2025-10-20
`yt issue-token --password` unrecognized arguments: --password password


Krock21:
According to documentation: https://ytsaurus.tech/docs/en/user-guide/storage/auth
```
yt issue-token alex --password cone
```
should work.

But it doesn't:
```
yt issue-token alex --password cone
usage: yt issue-token [-h] [--params PARAMS] user
yt issue-token: error: unrecognized arguments: --password cone
```

We need to either support that or adjust documentation

denvr:
Documentation is outdated - password is taken from stdin for security reasons
 https://github.com/ytsaurus/ytsaurus/pull/956

```
% yt issue-token --help
usage: yt issue-token [-h] [--params PARAMS] user

issues a new token for user, reading password interactively and securely from console

positional arguments:
  user             user to issue token

options:
  -h, --help       show this help message and exit
  --params PARAMS  specify additional params. structured params in yson format
```

dim-an:
Fixed in linked PR

Issue #1104, opened 2025-03-03
CLI can't set-user-password in non-interactive way


Krock21:
`yt set-user-password` command has only interactive mode and can't be used in scripts.

We need to a way to use it in scripts. For example, by allowing --password argument that disables interactive mode and uses passed password

zlobober:
That is sort of intentional, password-related arguments in CLI should be interactive and properly protected.

In scripts you can use python method yt.wrapper.set_user_password, it is non-interactive. You may invoke it via python -c, for example.

Krock21:
Yes I do this as a workaround
```
import os, yt.wrapper
ytc = yt.wrapper.YtClient(proxy=os.environ["YT_PROXY"], token=os.environ["YT_TOKEN"])
ytc.set_user_password("$user", new_password="$password")
```

By some reason, `yt.set_user_password` doesn't exist

minotru:
> That is sort of intentional, password-related arguments in CLI should be interactive and properly protected.

It would be helpful to update `--help` string of `yt set-user-password` to mention interactive/`echo <<password>> | yt set-user-password <<user>>` way.

Now help string is like:
```
$ yt set-user-password --help
usage: yt set-user-password [-h] [--params PARAMS] user

updates user password, reading it and optionally a current password
```

Docs should be updated as well: https://ytsaurus.tech/docs/en/user-guide/storage/auth Now it suggests old apporach:
```
$ yt set-user-password alex --new-password cone
```

PR #1485, opened 2025-10-22
Update documentation about authentication for secure(interactive) passwords

Krock21:
Issues: #1104 and #1483


ogorbacheva:
Shipped!

PR #1487, opened 2025-10-23
Introduce new job statistics: `job/{cpu,vcpu,memory,gpu}/cumulative_reserve`

Krock21:
- Add running-phase resource consumption in `TJob` and export aggregated CPU/vCPU/memory/GPUmetrics under `job/{cpu,vcpu,memory,gpu}/cumulative_reserve`
- Add integration tests to cover the new metrics across vanilla and merge/sort/map-reduce workloads

---

* Changelog entry
Type: feature
Component: map-reduce

Added job statistics `job/{cpu,vcpu,memory,gpu}/cumulative_reserve` for improved resource accounting.


Krock21:
Rebased to main

yt/yt/server/node/exec_node/job.cpp
```diff
@@ -4116,6 +4160,57 @@ void TJob::UpdateIOStatistics(const TStatistics& statistics)
...
+        if (newUsage) {
+            RunningPhaseResourceUsage_ = *newUsage;
+        }
+        return RunningPhaseResourceConsumption_;
+    }
+
+    auto duration = GetRunningPhaseDuration();
+    if (duration > RunningPhaseResourceConsumption_.Duration) {
+        auto durationDiff = duration - RunningPhaseResourceConsumption_.Duration;
+        auto durationMicroseconds = static_cast<i128>(durationDiff.MicroSeconds());
```
koct9i, line 4193: I don't get why you need microseocnds precision here.
What period of sampling for this tracking?
I suppose it is far bigger than 1ms.

yt/yt/server/node/exec_node/job.cpp
```diff
@@ -4116,6 +4160,57 @@ void TJob::UpdateIOStatistics(const TStatistics& statistics)
...
+        const auto usage = RunningPhaseResourceUsage_;
+
+        auto centicoresToMillicores = [] (i64 centicores) -> i128 {
+            return static_cast<i128>(centicores * 10);
+        };
+
+        RunningPhaseResourceConsumption_.Duration = duration;
+        RunningPhaseResourceConsumption_.Cpu += centicoresToMillicores(usage.Cpu.GetUnderlyingValue()) * durationMicroseconds;
```
koct9i, line 4201: This approximation is sampled both in time and usage level.
Three digits precision should be enough.

You can accumulate `centiCore*ms` and divide by 100 later
or convert into `core*ms` right here

yt/yt/server/node/exec_node/job.cpp
Krock21: > I don't get why you need microseocnds precision here.

I just need some precision that is good enough and doesn't lose anything. Microseconds is the lowest and does the work.
I can do milliseconds, it will reduce precision to 1 gibibyte-millisecond (or 1 mb-sec)

> What period of sampling for this tracking?

There is no period. It is updated whenever needed: when jobProxy reports usage to supervisorService, when job phase is changed, when statistics are calculated/changed

yt/yt/server/node/exec_node/job.cpp
Krock21: > You can accumulate core*ms usage += cpuCentoCore * durationMs / 10
or accumulate centiCore*ms and divide by ten later

I don't like centicores, this unit is not commonly used and increases risks of future bugs (it is 1/100th of a core and need to be divided by 100 to get cores). `usage += cpuCentiCore * durationMs / 100` loses 1 core*ms on **every update**, which is too much

This is why I convert to millicores and accumulate in millicores, to divide by 1000 on conversion

I can do it in centicores, but I believe this will hurt us more (with bugs) than potential benefit(?)

yt/yt/server/node/exec_node/job.cpp
```diff
@@ -4116,6 +4160,57 @@ void TJob::UpdateIOStatistics(const TStatistics& statistics)
...
+
+    auto stopTime = RunningPhaseStopTime_;
+    if (stopTime == TInstant::Zero()) {
+        stopTime = TInstant::Now();
+    }
+
+    return stopTime - RunningPhaseStartTime_;
+}
+
+TResourceConsumption TJob::UpdateConsumptionTracking(std::optional<NClusterNode::TJobResources> newUsage)
```
eshcherbin, line 4179: `const&`?

yt/yt/server/node/exec_node/job.cpp
eshcherbin: I would prefer millisecond precision as it is used everywhere else.

yt/yt/server/scheduler/helpers.cpp
```diff
@@ -339,6 +339,11 @@ const std::vector<TStatisticsDescription>& GetOperationStatisticsDescriptions()
         {"user_job/gpu/max_memory_used", "Maximum registered GPU memory usage", "bytes"},
         {"user_job/gpu/memory_total", "Total available GPU memory", "bytes"},
 
+        {"job/cpu/cumulative_reserve", "Amount of CPU reserved for job execution multiplied by time in milliseconds", "core*ms"},
```
eshcherbin, line 342: `core*ms` does not exist as far as our UI is concerned. Let's call it CPU-time and use "ms" as the unit. The same goes for vCPU and GPU.

yt/yt/server/scheduler/helpers.cpp
```diff
@@ -339,6 +339,11 @@ const std::vector<TStatisticsDescription>& GetOperationStatisticsDescriptions()
...
+        {"job/vcpu/cumulative_reserve", "Amount of vCPU reserved for job execution multiplied by time in milliseconds", "core*ms"},
+        {"job/memory/cumulative_reserve", "Amount of memory reserved for job execution (in gibibytes) multiplied by time in milliseconds", "GiB*ms"},
```
eshcherbin, line 344: I am strongly against using GiB and other binary memory units. They add unnecessary confusion and make it almost impossible to convert from bytes without using a calculator.

An alternative, which is already used, is "MB*sec". Let's use it here as well or discuss some other consistent solution.

yt/yt/server/node/exec_node/job.h
```diff
@@ -349,6 +365,11 @@ class TJob
 
     NProfiling::TBufferedProducerPtr UserJobSensorProducer_;
 
+    TResourceConsumption RunningPhaseResourceConsumption_;
+    NClusterNode::TJobResources RunningPhaseResourceUsage_ = {};
```
eshcherbin, line 369: nit: `= {}` is unnecessary

yt/yt/server/node/exec_node/job.h
```diff
@@ -64,6 +66,20 @@ struct TArtifactDescription
 
 ////////////////////////////////////////////////////////////////////////////////
 
+struct TResourceConsumption
```
eshcherbin, line 69: Maybe `TCumulativeResourceUsage`?

yt/yt/server/node/exec_node/job.h
```diff
@@ -349,6 +365,11 @@ class TJob
 
     NProfiling::TBufferedProducerPtr UserJobSensorProducer_;
 
+    TResourceConsumption RunningPhaseResourceConsumption_;
+    NClusterNode::TJobResources RunningPhaseResourceUsage_ = {};
+    TInstant RunningPhaseStartTime_ = TInstant::Zero();
+    TInstant RunningPhaseStopTime_ = TInstant::Zero();
```
eshcherbin, line 371: Let's move these lower closer to other `XxxStartTime_`/`XxxFinishTime_`. 
Also, make it optional and name `FinishTime` for consistency.

yt/yt/server/node/exec_node/job.cpp
```diff
@@ -4116,6 +4160,57 @@ void TJob::UpdateIOStatistics(const TStatistics& statistics)
...
+    if (RunningPhaseStartTime_ == TInstant::Zero()) {
+        return TDuration::Zero();
+    }
+
+    auto stopTime = RunningPhaseStopTime_;
+    if (stopTime == TInstant::Zero()) {
+        stopTime = TInstant::Now();
+    }
+
+    return stopTime - RunningPhaseStartTime_;
```
eshcherbin, line 4176: nit: just `return stopTime.value_or(TInstant::Now()) - RunningPhaseStartTime_`?
Assuming optional from other comment.

yt/yt/server/node/exec_node/job.cpp
```diff
@@ -2076,11 +2082,23 @@ void TJob::SetJobPhase(EJobPhase phase)
...
     YT_LOG_DEBUG(
         "Setting new job phase (Previous: %v, New: %v)",
-        JobPhase_,
+        previousPhase,
         phase);
 
+    if (phase == EJobPhase::Running && previousPhase != EJobPhase::Running) {
+        RunningPhaseStartTime_ = TInstant::Now();
+        RunningPhaseStopTime_ = TInstant::Zero();
+        RunningPhaseResourceConsumption_ = {};
```
eshcherbin, line 2095: nit: `RunningPhaseStopTime_` and `RunningPhaseResourceConsumption_` are already default-initialized.

yt/yt/server/node/exec_node/job.cpp
```diff
@@ -4116,6 +4160,57 @@ void TJob::UpdateIOStatistics(const TStatistics& statistics)
...
+    return stopTime - RunningPhaseStartTime_;
+}
+
+TResourceConsumption TJob::UpdateConsumptionTracking(std::optional<NClusterNode::TJobResources> newUsage)
+{
+    YT_ASSERT_THREAD_AFFINITY(JobThread);
+
+    if (RunningPhaseStartTime_ == TInstant::Zero()) {
+        if (newUsage) {
+            RunningPhaseResourceUsage_ = *newUsage;
```
eshcherbin, line 4185: Why do you want to store this resource usage separately? Don't we already store current resource usage somewhere else?

yt/yt/server/node/exec_node/job.cpp
```diff
@@ -4116,6 +4160,57 @@ void TJob::UpdateIOStatistics(const TStatistics& statistics)
...
+        }
+        return RunningPhaseResourceConsumption_;
+    }
+
+    auto duration = GetRunningPhaseDuration();
+    if (duration > RunningPhaseResourceConsumption_.Duration) {
```
eshcherbin, line 4191: I'd suggest simplifying a bit by removing this condition and computing `auto durationDelta = max(duration - RunningPhaseResourceConsumption_.Duration, 0)`.

yt/yt/server/node/exec_node/job.cpp
```diff
@@ -119,6 +119,8 @@
 
 #include <library/cpp/yt/system/handle_eintr.h>
 
+#include <limits>
```
eshcherbin, line 122: What do we need this include for?

yt/yt/server/scheduler/helpers.cpp
Krock21: `MB*sec` exists, `bytes * sec` exists, `ratio * ms` exists, `sec * MB`, `ms * power`, `ms * frequency`

I think just `ms` will add confusion and will break UI (possible support for converting and showing CPU*hr / CPU*s)

Same for vCPU (I can remove it btw) and GPU

yt/yt/server/scheduler/helpers.cpp
Krock21: It was mb*sec before, @koct9i asked for GiB to maintain consistency by time (ms everywhere)

> other binary memory units

Do you suggest using MB as in 1000000 bytes? It is 1MiB now, and MiB and GiB is used everywhere in YT

https://github.com/ytsaurus/ytsaurus/blob/86cfce251c1a328f150531206b33857b482cca55/yt/yt/server/job_proxy/memory_tracker.cpp#L210

https://github.com/ytsaurus/ytsaurus/blob/86cfce251c1a328f150531206b33857b482cca55/util/generic/size_literals_ut.cpp#L9

So current solution is consistent by time (ms) and binary units (GiB as 1024^3 bytes)

yt/yt/server/node/exec_node/job.cpp
Krock21: We don't store it, we update `ResourceHolder_->SetBaseResourceUsage(newUsage);`, and this `ResourceHolder` is from allocation and doesn't have "GetBaseResourceUsage" (only total with some strange Additional, which seems unused)

yt/yt/server/node/exec_node/job.cpp
Krock21: For explicit clamping with `std::numeric_limits<i64>::max()`
statistics have to be i64, but bytes*ms for memory is overflowing it quite fast

yt/yt/server/node/exec_node/job.cpp
eshcherbin: > `usage += cpuCentiCore * durationMs / 100` loses 1 core*ms on **every update**, which is too much

How often do you expect the updates to happen? I don't see how 1 CPU-millisecond error can be too much.

yt/yt/server/scheduler/helpers.cpp
eshcherbin: `ms` is currently used for CPU-time statistics: https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/server/scheduler/helpers.cpp#L293

yt/yt/server/scheduler/helpers.cpp
eshcherbin: I see, you are right. In my opinion, using binary units for `CumulativeMemoryUsageMBSec_` was a sad oversight on my part, but that seems to be unfixable now. Yes, I believe the best way is to use decimal units everywhere. The fact that unit suffixes in `size_literals.h` are called `_MB` instead of `_MiB` is also something I do not agree with.

Given all that, I would prefer to at least not introduce any more new units without preliminary discussion.

yt/yt/server/node/exec_node/job.cpp
eshcherbin: I believe getting resource usage from `ResourceHolder` is the right way to go.

eshcherbin (changes requested):
(see previous comments)

PR #1488, opened 2025-10-23
yt/core/bus: handle SSL transport errors

koct9i:
- **yt/core/bus: report SSL syscall error as transport error**
- **yt/core/bus: test ssl connection transport errors**
- **yt/core/bus: handle SslError as retriable channel failure error**
- **yt/core/rpc: test RPC over BUS with TLS**

---

* Changelog entry
Type: fix
Component: misc-server

Network errors behind SSL connection should be reported as transport error
to avoid confusing with actual SSL error and do proper retrying.


yalekseev:
Shipped!

PR #1489, opened 2025-10-27
yt/ytlib: add region into cluster connection

koct9i:
In cloud hierarchy "region" represent geographical location which
may contain several "Availability Zones" (AZ),
which in other hand may contain several "Data Centers" (DC).

Region in cluster connection allows to project it to cloud and network.

Each cluster node also could announce which Data Center it belongs.

---

* Changelog entry
Type: feature
Component: misc-server

Add field "region" into cluster connection configuration.


maxim-babenko:
Shipped!

PR #1491, opened 2025-10-28
fix S3 client hanging on HEAD errors

faucct:
`NYT::NS3::TClient::ErrorFromResponse` hangs till a Read timeout for HEAD request responses:
https://github.com/ytsaurus/ytsaurus/blob/317c99834316ac46f05429494d0b3291ffbeb8c8/yt/yt/library/s3/client.cpp#L459

### DIFF
--- a/yt/yt/core/http/stream.cpp
+++ b/yt/yt/core/http/stream.cpp
@@ -233,7 +233,7 @@ int THttpParser::OnHeadersComplete(http_parser* parser)
      * HEAD request which may contain 'Content-Length' or 'Transfer-Encoding:
      * chunked' headers that indicate the presence of a body.
      */
-    return that->ParserType_ == HTTP_RESPONSE && that->RequestMethod_ == EMethod::Head ? 1 : 0;
+    return (that->ParserType_ == HTTP_RESPONSE && that->RequestMethod_ == EMethod::Head) ? 1 : 0;
 }
 
 int THttpParser::OnBody(http_parser* parser, const char* at, size_t length)
--- a/yt/yt/core/http/stream.h
+++ b/yt/yt/core/http/stream.h
@@ -31,7 +31,7 @@ DEFINE_ENUM(EParserState,
 class THttpParser
 {
 public:
-    explicit THttpParser(http_parser_type parserType, std::optional<EMethod> requestMethod);
+    THttpParser(http_parser_type parserType, std::optional<EMethod> requestMethod);
 
     static http_parser_settings GetParserSettings();
 


savnadya:
It’s probably already fixed [here](https://github.com/ytsaurus/ytsaurus/commit/1d20e5c0f4d82effd931e4be5550e1423a693ee0).

faucct:
It looks like it might be, as this was the only HEAD request, and I will check tomorrow, but the http client is still not following the spec, so I feel like this change should also be merged.

yt/yt/core/http/stream.cpp
```diff
@@ -253,15 +255,15 @@ THttpInput::THttpInput(
     IConnectionPtr connection,
     const TNetworkAddress& remoteAddress,
     IInvokerPtr readInvoker,
-    EMessageType messageType,
```
achulkov2, line 256: This transition from `messageType -> requestMethod` is kind of non-obvious, they represent different things. For the callee, it seems non-obvious that EMessageType::Request = leave requestMethod as std::nullopt — one might think that maybe we were supposed to provide the method this is a response to, and still provide it. (it is especially confusing because we provide **request**Method only for EMessageType::**Response**)

Let's keep the message type parameter, adding `requestMethod` as an additional one? This way the intent is clearer, code is easier to understand (and we are less likely to make a mistake). For example, right now it is harder to verify that we are not breaking any current logic.

yt/yt/core/http/stream.cpp
```diff
@@ -58,11 +58,12 @@ http_parser_settings THttpParser::GetParserSettings()
 
 const http_parser_settings ParserSettings = THttpParser::GetParserSettings();
 
-THttpParser::THttpParser(http_parser_type parserType)
-    : ParserType_(parserType)
+THttpParser::THttpParser(std::optional<EMethod> requestMethod)
+    : ParserType_(requestMethod ? HTTP_RESPONSE : HTTP_REQUEST)
     , Headers_(New<THeaders>())
+    , IsHeadResponse_(requestMethod == EMethod::Head)
```
achulkov2: Let's just persist `requestMethod` as a whole then?

yt/yt/core/http/stream.cpp
```diff
@@ -222,7 +223,8 @@ int THttpParser::OnHeadersComplete(http_parser* parser)
 
     that->State_ = EParserState::HeadersFinished;
 
-    return 0;
+    // https://datatracker.ietf.org/doc/html/rfc2616#page-33
+    return that->IsHeadResponse_ ? 1 : 0;
```
achulkov2: Let's add a comment about the proper return value for OnHeadersComplete from the library we are using.

```
/* Callbacks should return non-zero to indicate an error. The parser will
 * then halt execution.
 *
 * The one exception is on_headers_complete. In a HTTP_RESPONSE parser
 * returning '1' from on_headers_complete will tell the parser that it
 * should not expect a body. This is used when receiving a response to a
 * HEAD request which may contain 'Content-Length' or 'Transfer-Encoding:
 * chunked' headers that indicate the presence of a body.
 ```
 
This section really helps understand what is going on here and it is beneficial for the reader not to have to go looking for it. Let's write a few sentences.

File: yt/yt/core/http/stream.cpp [hunk 4]
```diff
@@ -222,7 +223,17 @@ int THttpParser::OnHeadersComplete(http_parser* parser)
 
     that->State_ = EParserState::HeadersFinished;
 
-    return 0;
+    // https://datatracker.ietf.org/doc/html/rfc2616#page-33
+    /* Callbacks should return non-zero to indicate an error. The parser will
+     * then halt execution.
+     *
+     * The one exception is on_headers_complete. In a HTTP_RESPONSE parser
```
achulkov2, line 230: Judging by the comment text, we should also check that this is a `HTTP_RESPONSE` parser? Would be the safest bet.

yt/yt/core/http/stream.cpp
```diff
@@ -254,18 +265,22 @@ THttpInput::THttpInput(
...
     , ReadInvoker_(std::move(readInvoker))
     , InputBuffer_(TSharedMutableRef::Allocate<THttpParserTag>(Config_->ReadBufferSize, {.InitializeStorage = false}))
-    , Parser_(messageType == EMessageType::Request ? HTTP_REQUEST : HTTP_RESPONSE)
+    , Parser_(messageType == EMessageType::Request ? HTTP_REQUEST : HTTP_RESPONSE, requestMethod)
     , StartByteCount_(Connection_->GetReadByteCount())
     , StartStatistics_(Connection_->GetReadStatistics())
     , LastProgressLogTime_(TInstant::Now())
-{ }
+{
+    YT_VERIFY((MessageType_ == EMessageType::Response) == requestMethod.has_value());
```
achulkov2: I wouldn't say the requirement is needed both ways? Let's just verify that if `MessageType_ == EMessageType::Response`, then `requestMethod` is non null, since we need to know it. If someone decided to provide the method for `EMessageType::Request`, why not. (combined with the fact that we will check for the parser type in the parser, as mentioned in the other comment)

yt/yt/core/http/stream.cpp
```diff
@@ -254,18 +265,22 @@ THttpInput::THttpInput(
...
     THttpIOConfigPtr config)
     : Connection_(std::move(connection))
     , RemoteAddress_(remoteAddress)
     , MessageType_(messageType)
+    , RequestMethod_(requestMethod)
     , Config_(std::move(config))
     , ReadInvoker_(std::move(readInvoker))
     , InputBuffer_(TSharedMutableRef::Allocate<THttpParserTag>(Config_->ReadBufferSize, {.InitializeStorage = false}))
-    , Parser_(messageType == EMessageType::Request ? HTTP_REQUEST : HTTP_RESPONSE)
+    , Parser_(messageType == EMessageType::Request ? HTTP_REQUEST : HTTP_RESPONSE, requestMethod)
```
achulkov2: Let's place these params on separate lines, otherwise ternary operator is hard to read.

yt/yt/core/http/stream.cpp
```diff
@@ -222,7 +223,17 @@ int THttpParser::OnHeadersComplete(http_parser* parser)
...
+    /* Callbacks should return non-zero to indicate an error. The parser will
+     * then halt execution.
+     *
+     * The one exception is on_headers_complete. In a HTTP_RESPONSE parser
+     * returning '1' from on_headers_complete will tell the parser that it
+     * should not expect a body. This is used when receiving a response to a
+     * HEAD request which may contain 'Content-Length' or 'Transfer-Encoding:
+     * chunked' headers that indicate the presence of a body.
+     */
+    return that->ParserType_ == HTTP_RESPONSE && that->RequestMethod_ == EMethod::Head ? 1 : 0;
```
Kontakter: Let's take the condition in braces `(xxx && yyy) ? 1 : 0`

yt/yt/core/http/stream.h
```diff
@@ -31,7 +31,7 @@ DEFINE_ENUM(EParserState,
 class THttpParser
 {
 public:
-    explicit THttpParser(http_parser_type parserType);
+    explicit THttpParser(http_parser_type parserType, std::optional<EMethod> requestMethod);
```
Kontakter: `explicit` is not needed anymore

savnadya:
Shipped!

maxim-babenko:
Please capitalize the commit message

yt/yt/core/http/stream.h
```diff
@@ -60,6 +60,7 @@ class THttpParser
     TStringBuilder NextValue_;
 
     THeadersPtr Headers_;
+    std::optional<EMethod> RequestMethod_;
```
maxim-babenko, line 63: Let's make it const and also rearrange const and move above to place it near the other const members (respecting the order)

yt/yt/core/http/stream.cpp
```diff
@@ -254,18 +265,26 @@ THttpInput::THttpInput(
...
-    , Parser_(messageType == EMessageType::Request ? HTTP_REQUEST : HTTP_RESPONSE)
+    , Parser_(
+        messageType == EMessageType::Request ? HTTP_REQUEST : HTTP_RESPONSE,
+        requestMethod)
     , StartByteCount_(Connection_->GetReadByteCount())
     , StartStatistics_(Connection_->GetReadStatistics())
     , LastProgressLogTime_(TInstant::Now())
-{ }
+{
+    if (MessageType_ == EMessageType::Response) {
```
maxim-babenko, line 284: How about expressing this as a single `YT_VERIFY`?

PR #1501, opened 2025-11-06
Update descriptions for operation statistics

koct9i:
- **yt/yt/server/scheduler: update descriptions for operation statistics**
- **yt/tests: check operations statistics descriptions**

### DIFF
--- a/yt/yt/tests/integration/node/test_user_job.py
+++ b/yt/yt/tests/integration/node/test_user_job.py
@@ -20,6 +20,7 @@ from yt_commands import (
     sync_create_cells, get_singular_chunk_id,
     update_nodes_dynamic_config, set_node_banned, check_all_stderrs,
     assert_statistics, assert_statistics_v2, extract_statistic_v2,
+    OperationStatisticsViewer,
     heal_exec_node,
     make_random_string, raises_yt_error, update_controller_agent_config, update_scheduler_config,
     get_supported_erasure_codecs)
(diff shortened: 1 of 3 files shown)


yt/yt/server/scheduler/helpers.cpp
```diff
@@ -339,14 +385,29 @@ const std::vector<TStatisticsDescription>& GetOperationStatisticsDescriptions()
         {"user_job/gpu/max_memory_used", "Maximum registered GPU memory usage", "bytes"},
         {"user_job/gpu/memory_total", "Total available GPU memory", "bytes"},
 
+        {"user_job/finalization_time", "Finalization time of the process after closing stderr and stdout pipes", "ms"},
+
+        {"user_job/profiler_failure_count", "Number of job profiler failures", "pieces"},
+
+        // Custom Statistics.
+        {"custom/**", "Custom statistics reported by user job", ""},
```
eshcherbin: I am not sure if such a description is needed and if it will be processed correctly by the UI. The main purpose of this list is for UI to be able to display pop-up helpers and format values with given units. Custom statistics should be addressed in the documentation but not here, in my opinion.

koct9i: Ok.

eshcherbin:
Shipped!

yt/yt/tests/integration/node/test_user_job.py
```diff
@@ -4193,6 +4194,9 @@ def test_do_not_account_prepare_time(self):
             job_type="task",
         ))
 
+        undescribed = OperationStatisticsViewer().get_undescribed(op.get_statistics())
+        assert not undescribed, f"Undescribed operation statistics: {undescribed}"
```
eshcherbin, line 4198: This test does not pass because of several undescribed GPU statistics. I would rather you added them, however, some of them are purely internal, so you might have some difficulty describing them. Alternatively, I can do it myself when I have some free time this week.

eshcherbin:
I have added descriptions for all statistics and refactored the code slightly.

PR #1514, opened 2025-11-20
[stable/25.3] cherry-pick TLS commits

koct9i:
- **Make HTTPS client secure by default**
- **yt/queue_agent: use native connection bus channel factory for orchid redirects**
- **mTLS: job proxy connectivity**
- **yt/environment: enable mutual TLS for cluster connection**
- **Add config for job proxy environment variables**
- **yt/core/bus: handle SSL transport errors**
- **Explicitly add generate.sh to test_framework library commit_hash:faf4d0b7c14016a8f3499ad65fc049e4fe72d2fa**

### DIFF
--- a/yt/yt/core/test_framework/CMakeLists.darwin-arm64.txt
+++ b/yt/yt/core/test_framework/CMakeLists.darwin-arm64.txt
@@ -77,12 +77,12 @@ target_link_libraries(yt-core-test_framework.global PUBLIC
 
 target_sources(yt-core-test_framework.global PRIVATE
   ${PROJECT_SOURCE_DIR}/yt/yt/core/test_framework/framework.cpp
-  ${PROJECT_BINARY_DIR}/yt/yt/core/test_framework/cb76b690f4e496634c2dc6f5c8361292.cpp
+  ${PROJECT_BINARY_DIR}/yt/yt/core/test_framework/bfbc2f48095040480eacdd42dba4fd39.cpp
 )
 
 
 resources(yt-core-test_framework.global
-  ${PROJECT_BINARY_DIR}/yt/yt/core/test_framework/cb76b690f4e496634c2dc6f5c8361292.cpp
+  ${PROJECT_BINARY_DIR}/yt/yt/core/test_framework/bfbc2f48095040480eacdd42dba4fd39.cpp
   INPUTS
   ${PROJECT_SOURCE_DIR}/yt/yt/core/test_framework/testdata/ca.pem
   ${PROJECT_SOURCE_DIR}/yt/yt/core/test_framework/testdata/ca_ec.pem
@@ -98,6 +98,7 @@ resources(yt-core-test_framework.global
   ${PROJECT_SOURCE_DIR}/yt/yt/core/test_framework/testdata/rpc_server_key.pem
   ${PROJECT_SOURCE_DIR}/yt/yt/core/test_framework/testdata/rpc_client_cert.pem
   ${PROJECT_SOURCE_DIR}/yt/yt/core/test_framework/testdata/rpc_client_key.pem
+  ${PROJECT_SOURCE_DIR}/yt/yt/core/test_framework/testdata/generate.sh
   KEYS
   /testdata/ca.pem
   /testdata/ca_ec.pem
@@ -113,5 +114,6 @@ resources(yt-core-test_framework.global
   /testdata/rpc_server_key.pem
   /testdata/rpc_client_cert.pem
   /testdata/rpc_client_key.pem
+  /testdata/generate.sh
   OPTS
 )
--- a/yt/yt/core/test_framework/CMakeLists.darwin-x86_64.txt
+++ b/yt/yt/core/test_framework/CMakeLists.darwin-x86_64.txt
@@ -77,12 +77,12 @@ target_link_libraries(yt-core-test_framework.global PUBLIC
 
 target_sources(yt-core-test_framework.global PRIVATE
   ${PROJECT_SOURCE_DIR}/yt/yt/core/test_framework/framework.cpp
-  ${PROJECT_BINARY_DIR}/yt/yt/core/test_framework/cb76b690f4e496634c2dc6f5c8361292.cpp
+  ${PROJECT_BINARY_DIR}/yt/yt/core/test_framework/bfbc2f48095040480eacdd42dba4fd39.cpp
 )
 
 
 resources(yt-core-test_framework.global
-  ${PROJECT_BINARY_DIR}/yt/yt/core/test_framework/cb76b690f4e496634c2dc6f5c8361292.cpp
+  ${PROJECT_BINARY_DIR}/yt/yt/core/test_framework/bfbc2f48095040480eacdd42dba4fd39.cpp
   INPUTS
   ${PROJECT_SOURCE_DIR}/yt/yt/core/test_framework/testdata/ca.pem
   ${PROJECT_SOURCE_DIR}/yt/yt/core/test_framework/testdata/ca_ec.pem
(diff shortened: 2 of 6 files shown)


yt/yt/server/node/exec_node/bootstrap.cpp
```diff
@@ -520,8 +524,12 @@ class TBootstrap
 
         newJobProxyConfigTemplate->AuthenticationManager = GetConfig()->ExecNode->JobProxy->JobProxyAuthenticationManager;
 
-        newJobProxyConfigTemplate->SupervisorConnection = New<NYT::NBus::TBusClientConfig>();
-        newJobProxyConfigTemplate->SupervisorConnection->Address = localAddress;
+        if (const auto& supervisorConnection = GetConfig()->ExecNode->JobProxy->SupervisorConnection) {
+            newJobProxyConfigTemplate->SupervisorConnection = CloneYsonStruct(supervisorConnection);
```
yalekseev, line 528: It might be useful to add either

YT_VERIFY(newJobProxyConfigTemplate->SupervisorConnection->Address);

or

if (!newJobProxyConfigTemplate->SupervisorConnection->Address) {  
    newJobProxyConfigTemplate->SupervisorConnection->Address = localAddress;  
}

just to make sure that address is always set.

yt/yt/core/crypto/config.h
```diff
@@ -30,9 +30,6 @@ DEFINE_REFCOUNTED_TYPE(TPemBlobConfig)
 
 ////////////////////////////////////////////////////////////////////////////////
 
-//! FIXME: Enabled during migration, because this code has always been broken.
-constexpr bool DefaultInsecureSkipVerify = true;
```
yalekseev, line 34: Does this change now require servers to verify client certificates?

yt/yt/core/rpc/helpers.cpp
```diff
@@ -66,6 +66,7 @@ bool IsChannelFailureError(const TError& error)
     }
     return
         code == NRpc::EErrorCode::TransportError ||
+        code == NRpc::EErrorCode::SslError ||
```
yalekseev, line 69: It looks like all SSL certificate verification errors are becoming retry-able now.  Would it be all right?

yt/yt/core/crypto/config.cpp
```diff
@@ -44,7 +44,7 @@ TString TPemBlobConfig::LoadBlob(TCertificatePathResolver pathResolver) const
         auto filePath = pathResolver ? pathResolver(*FileName) : *FileName;
         return TFileInput(filePath).ReadAll();
     }
-    THROW_ERROR_EXCEPTION("Neither \"environment_variable\" nor \"file_name\" nor \"value\" is given");
+    THROW_ERROR_EXCEPTION("Cannot load environment variable %Qv", EnvironmentVariable);
```
yalekseev, line 47: We are loading a blob here from environment, config value or a file. Probably a more generic error message would be better:

```
THROW_ERROR_EXCEPTION("Cannot load PEM blob: no valid source specified (environment_variable=%v, file_name=%v, value=%v)",  
    EnvironmentVariable, FileName, Value ? "<set>" : "<unset>");
```

yt/yt/server/job_proxy/user_job.cpp
```diff
@@ -1390,6 +1391,16 @@ class TUserJob
             }
         }
 
+        for (const auto& variable : Config_->EnvironmentVariables) {
```
koct9i, line 1394: Look upper. There is global forward all switch

yalekseev, line 1394: It is difficult to understand what this code snippet is intended to do. Is it

```
for (const auto& variable : Config_->EnvironmentVariables) {  
    bool shouldForward = variable->ForwardToUserJob.value_or(Config_->ForwardAllEnvironmentVariables);  
    
    if (shouldForward && !Config_->ForwardAllEnvironmentVariables) {  
        // Explicitly forward this variable  
        SetEnvironment(Format("%v=%v", variable->Name, GetEnv(variable->Name)));  
    } else if (!shouldForward && Config_->ForwardAllEnvironmentVariables) {  
        // Explicitly block this variable from being forwarded  
        SetEnvironment(variable->Name);  
    }  
}
```

?

yt/yt/server/lib/job_proxy/config.cpp
```diff
@@ -107,6 +109,46 @@ void TBindConfig::Register(TRegistrar registrar)
...
+            return *std::move(value);
+        }
+    }
+    if (Value) {
+        return *Value;
+    }
+    if (FileName) {
+        return TFileInput(*FileName).ReadAll();
+    }
+    THROW_ERROR_EXCEPTION("Cannot load environment variable %Qv", EnvironmentVariable);
```
yalekseev, line 147: We are loading a blob here from environment, config value or a file. Probably a more generic error message would be better:

```
THROW_ERROR_EXCEPTION("Cannot load PEM blob: no valid source specified (environment_variable=%v, file_name=%v, value=%v)",  
    EnvironmentVariable, FileName, Value ? "<set>" : "<unset>");
```

koct9i: This is cherry-pick from main branch. Could we improve errors there first and cherry pick into stable later?

koct9i: If file name is defined then we'll never get this exception.

koct9i: Same is true for "value".

yt/yt/server/job_proxy/user_job.cpp
koct9i:
```
        if (Config_->ForwardAllEnvironmentVariables) {
            auto env = GetEnviron();
            for (const auto& variable : env) {
                SetEnvironment(variable);
            }
        }

        for (const auto& variable : Config_->EnvironmentVariables) {
            if (variable->ForwardToUserJob.value_or(false) && !Config_->ForwardAllEnvironmentVariables) {
                // Set environment variable if it is not forwarded yet.
                SetEnvironment(Format("%v=%v", variable->Name, GetEnv(variable->Name)));
            } else if (!variable->ForwardToUserJob.value_or(true) && Config_->ForwardAllEnvironmentVariables) {
                // Unset environment variable if it should not be forwarded with all variables.
                SetEnvironment(variable->Name);
            }
        }
```

yt/python/yt/environment/api/__init__.py
```diff
@@ -81,6 +81,8 @@ class LocalYtConfig(object):
     public_ca_cert_key = attr.ib(None)
     rpc_cert = attr.ib(None)
     rpc_cert_key = attr.ib(None)
+    rpc_client_cert = attr.ib(None)
+    rpc_client_cert_key = attr.ib(None)
```
yalekseev, line 85: A little bit below piblic_rpc_cert_key -> public_rpc_cert_key.

yt/python/yt/environment/configs_provider.py
```diff
@@ -1657,6 +1686,34 @@ def _build_rpc_proxy_configs(multidaemon_config_output,
...
+    if not yt_config.enable_tls:
+        return {
+            "encryption_mode": "disabled",
+            "verification_mode": "none",
+        }
+
+    config = {
+        "encryption_mode": "required",
+        "verification_mode": "full",
+        "peer_alternative_host_name": yt_config.cluster_name,
```
yalekseev, line 1699: when is_server=True do we want to set peer_alternative_host_name?

koct9i: in current scheme both sides verifies peer certificate agains "alt-name"

yalekseev:
Shipped!

yt/yt/server/job_proxy/user_job.cpp
```diff
@@ -1390,6 +1391,16 @@ class TUserJob
...
+            if (variable->ForwardToUserJob.value_or(false) && !Config_->ForwardAllEnvironmentVariables) {
```
yalekseev, line 1395: What if TLS is disabled, are we going to forward certificates to user job any way? Do we need to?

yt/yt/core/rpc/helpers.cpp
koct9i: We have several issues when something went wrong because request were not retries because network connection issues after wrapping connection with tls were handled differently. We don't have points against not making it retriable yet. Too many things in YT works only just because requests are retried. I don't have much time to invest into this matter.

yt/yt/server/job_proxy/user_job.cpp
koct9i: This option acts likes config said. It knows nothing about TLS.

PR #1516, opened 2025-11-24
Use get_config_from_env for YtClient in init scripts

koct9i:
- **init_operations_archive: use get_config_from_env**
- **init_queue_agent_state: use get_config_from_env**
- **init_query_tracker_state: use get_config_from_env**

---

Init scripts should respect configuration files and environment variables.
For example this is required for passing CA-bundle for HTTPS-only cluster.

There is simple way to fetch all options from environment and config files.
Unfortunately this is not default behaviour and contains some boilerplate.

update_config_from_env() configs applying order:
- config argument
- apply env YT_CONFIG_PATCHES
- apply config file YT_CONFIG_PATH (~/.yt/config)
- apply env YT_*

YtClient configs applying order:
- get_default_config()
- apply config argument
- apply proxy and token argument
- apply remote patch

### DIFF
--- a/yt/python/yt/environment/init_query_tracker_state.py
+++ b/yt/python/yt/environment/init_query_tracker_state.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python3
 
-from yt.wrapper import YtClient, config
+from yt.wrapper import YtClient
+from yt.wrapper.default_config import update_config_from_env
 
 from yt.environment.migrationlib import TableInfo, Migration, Conversion
 
@@ -1940,7 +1941,7 @@ def build_arguments_parser():
     parser.add_argument("--force", action="store_true", default=False)
     parser.add_argument("--state-path", type=str, default=DEFAULT_STATE_PATH)
     parser.add_argument("--shard-count", type=int, default=DEFAULT_SHARD_COUNT)
-    parser.add_argument("--proxy", type=str, default=config["proxy"]["url"])
+    parser.add_argument("--proxy", type=str, default=None)
 
     group = parser.add_mutually_exclusive_group(required=True)
     group.add_argument("--target-version", type=int)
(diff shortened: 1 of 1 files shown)


yt/python/yt/environment/init_query_tracker_state.py
```diff
@@ -1954,12 +1955,11 @@ def main():
     args = build_arguments_parser().parse_args()
     client = YtClient(
         proxy=args.proxy,
-        token=config["token"],
-        config={
+        config=update_config_from_env({
```
denvr, line 1958: Common way is to use config from `yt.wrapper.config` (as accumulated point of settings from env and other configuration methods) instead direct env access (get_config_from_env, update_config_from_env).

denvr:
What is the purpose of this PR that changes behaviour only part of init scripts?

yt/python/yt/environment/init_query_tracker_state.py
koct9i: Global config has awkward "load on import/first usage" semantics.
Explicit function call is much better.

It'a hard to tell how global config supports config files and you definitely cannot change config profile later.

And it does not support injection user code default options between global defaults and user config file/environment options options like here.

sane applying order is
- library config defaults
- user code config defaults
- user config file
- environment variables
- cli arguments
- constructor arguments
- remote config patch (this one could be weaker, but should be fine apply it last)

koct9i:
> What is the purpose of this PR that changes behaviour only part of init scripts?

For HTTPS-only cluster init scripts must be able to pick non-trivial config.
Like TLS CA bundle or IP options.
Right now they ignore any options besides YT_PROXY and YT_TOKEN

yt/python/yt/environment/init_query_tracker_state.py
denvr: In common - all Clients should be configured in one way (defaults, patch sequence, any config logic) and global config serve it well for almost all cases.
But yes, there lack of good functional method for it (`update_config_from_env` expect "full config" for first parameter, not only patch)

denvr:
LGTM

PR #1525, opened 2025-11-28
Consider SSL_CERT_FILE and SSL_CERT_DIR environment variables in python binaries

koct9i:
There are several ways how builtin CA bundle reaches SSLContext:
- SSLContext.load_default_certs()
- certify.where() -> (callable) -> patched SSLContext.load_verify_locations()
- yt python-requests certify.where() -> (blob) -> SSLContext

Lets use here path from environment variable SSL_CERT_FILE when defined.
This aligns with most modern design and golang behaviour.

Environment variable SSL_CERT_DIR provides directories with certificates
while will be used in addition to main CA bundle.

### DIFF
--- a/contrib/tools/python3/Lib/ssl.py
+++ b/contrib/tools/python3/Lib/ssl.py
@@ -543,7 +543,10 @@ class SSLContext(_SSLContext):
         if not isinstance(purpose, _ASN1Object):
             raise TypeError(purpose)
 
-        self.load_verify_locations(cadata=builtin_cadata())
+        cafile, capath, cadata = os.environ.get("SSL_CERT_FILE"), os.environ.get("SSL_CERT_DIR"), None
+        if cafile is None:
+            cadata=builtin_cadata()
+        self.load_verify_locations(cafile=cafile, capath=capath, cadata=cadata)
 
         if sys.platform == "win32":
             for storename in self._windows_cert_stores:
--- a/yt/python/contrib/python-requests/requests/certs.py
+++ b/yt/python/contrib/python-requests/requests/certs.py
@@ -17,6 +17,10 @@ import os
 import sys
 
 def where():
+    cafile = os.environ.get("SSL_CERT_FILE")
+    if cafile is not None:
+        return cafile
+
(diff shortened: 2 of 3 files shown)


koct9i:
I've posted link in yatool channel as well.

yalekseev:
Shipped!

yt/python/contrib/python-requests/requests/certs.py
```diff
@@ -17,6 +17,10 @@
 import sys
 
 def where():
+    cafile = os.environ.get("SSL_CERT_FILE")
```
denvr: Is this behaviour compatible with `requests`?

contrib/tools/python3/Lib/ssl.py
```diff
@@ -543,7 +543,10 @@ def load_default_certs(self, purpose=Purpose.SERVER_AUTH):
         if not isinstance(purpose, _ASN1Object):
             raise TypeError(purpose)
 
-        self.load_verify_locations(cadata=builtin_cadata())
+        cafile, capath, cadata = os.environ.get("SSL_CERT_FILE"), os.environ.get("SSL_CERT_DIR"), None
+        if cafile is None:
+            cadata=builtin_cadata()
+        self.load_verify_locations(cafile=cafile, capath=capath, cadata=cadata)
```
shadchin, line 549: This will break the old behavior:
* We always loaded builtin_cadata
* Right now, if there is cafile, then we don't upload it.

I suggest so:
```suggestion
        cafile, capath, cadata = os.environ.get("SSL_CERT_FILE"), os.environ.get("SSL_CERT_DIR"), builtin_cadata()
        self.load_verify_locations(cafile=cafile, capath=capath, cadata=cadata)
```
* Always loaded builtin_cadata
* If there is cafile, it will also be additionally uploaded.

koct9i: > This will break the old behavior:

This is intentional.

https://t.me/yatool_ru/293

I want to get rid of built-in ca bundle and provide complete  control for source of trust.

PR #1526, opened 2025-11-28
S3 regular chunk reader & writer

cherepasshka:
* Changelog entry
Type: feature
Component: s3

Added regular chunk reader and writer for S3

### DIFF
--- a/yt/yt/server/lib/s3/unittests/upload_session_ut.cpp
+++ b/yt/yt/server/lib/s3/unittests/upload_session_ut.cpp
@@ -92,7 +92,7 @@ protected:
     void CleanBuckets()
     {
         YT_VERIFY(S3Client_);
-    
+
         // Clean all the objects and buckets up.
         auto listBucketsRsp = WaitFor(S3Client_->ListBuckets({}))
             .ValueOrThrow();
--- a/yt/yt/server/lib/s3/unittests/ya.make
+++ b/yt/yt/server/lib/s3/unittests/ya.make
@@ -4,8 +4,6 @@ INCLUDE(${ARCADIA_ROOT}/yt/ya_cpp.make.inc)
 
 SUBSCRIBER(g:yt)
 
-ALLOCATOR(TCMALLOC)
-
 SRCS(
     chunk_reader_writer_ut.cpp
     upload_session_ut.cpp
@@ -15,9 +13,7 @@ INCLUDE(${ARCADIA_ROOT}/yt/opensource.inc)
 
 PEERDIR(
     yt/yt/server/lib/s3
-
     yt/yt/ytlib
-
     library/cpp/digest/md5
 )
 


yt/yt/server/lib/s3/chunk_reader.cpp
```diff
@@ -0,0 +1,288 @@
...
+        return ReadBlockRanges(options, GetBlockRanges(firstBlockIndex, blockCount));
+    }
+
+    TFuture<TRefCountedChunkMetaPtr> GetMeta(
+        const TGetMetaOptions& options,
+        const std::optional<TPartitionTags>& partitionTags = std::nullopt,
+        const std::optional<std::vector<int>>& extensionTags = {}) override
+    {
+        if (partitionTags) {
+            // YT_LOG_ALERT("Get meta request for S3 media was formed with partition tags (PartitionTags: %v)", partitionTag);
```
AleksandraZh: м?

yt/yt/server/lib/s3/chunk_reader.cpp
```diff
@@ -0,0 +1,288 @@
...
+            // THROW_ERROR_EXCEPTION("Partiton tags are not supported in get meta request")
+            //     << TErrorAttribute("partition_tags", partitionTags);
+        }
+        if (extensionTags) {
```
AleksandraZh: пустую строчку

yt/yt/server/lib/s3/chunk_reader.cpp
```diff
@@ -0,0 +1,288 @@
...
+            THROW_ERROR_EXCEPTION("Extension tags are not supported in get meta request")
+                << TErrorAttribute("extension_tags", extensionTags);
+        }
+
+        return DoGetMeta(options.ClientOptions);
+    }
+
+    TChunkId GetChunkId() const override
+    {
+        return ChunkId_;
```
AleksandraZh, line 127: chunk id только ради этой функции нужен?

yt/yt/server/lib/s3/chunk_reader.cpp
```diff
@@ -0,0 +1,288 @@
...
+        request.Key = ChunkMetaPlacement_.Key;
+        return Client_->GetObject(request)
+            .Apply(BIND([options, this, this_ = MakeStrong(this)] (const TGetObjectResponse& response) {
+                auto metaWithChunkId = PhysicalChunkLayoutReader_->DeserializeMeta(response.Data, options.ChunkReaderStatistics);
+
+                auto guard = WriterGuard(MetaLock_);
+
+                if (!ChunkMeta_) {
+                    ChunkMeta_ = metaWithChunkId.ChunkMeta;
+                    BlocksExt_ = New<NIO::TBlocksExt>(GetProtoExtension<NChunkClient::NProto::TBlocksExt>(ChunkMeta_->extensions()));
```
AleksandraZh: why are we storing it in a separate field?

yt/yt/server/lib/s3/chunk_reader.cpp
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+
+    TFuture<NIO::TBlocksExtPtr> GetBlocksExt(
+        const TClientChunkReadOptions& options,
+        IInvokerPtr invoker = nullptr)
+    {
+        return DoGetMeta(options, invoker)
+            .AsVoid()
+            .Apply(BIND([this, this_ = MakeStrong(this)] () {
```
AleksandraZh: () are extra

yt/yt/server/lib/s3/chunk_reader.cpp
```diff
@@ -0,0 +1,288 @@
...
+                auto guard = ReaderGuard(MetaLock_);
+                return BlocksExt_;
+            }));
+    }
+
+
+    void DumpBrokenMeta(TRef /*block*/) const
+    {
```
AleksandraZh: { }

yt/yt/server/lib/s3/chunk_reader.cpp
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    void DumpBrokenBlock(
+        int /*blockIndex*/,
+        const NIO::TBlockInfo& /*blockInfo*/,
+        TRef /*block*/) const
+    {
```
AleksandraZh: and here

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,825 @@
...
+public:
+    TS3UploadSessionBase(
+        IClientPtr client,
+        TS3MediumDescriptor::TS3ObjectPlacement objectPlacement,
+        const NLogging::TLogger& logger)
+        : Client_(std::move(client))
+        , ObjectPlacement_(std::move(objectPlacement))
+        , Logger(logger.WithTag("Bucket: %v, Key: %v", ObjectPlacement_.Bucket, ObjectPlacement_.Key))
+    {
+        // Only errors are expected to be set to this future, it should never complete successfully.
```
AleksandraZh: 😱

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,825 @@
...
+
+        // If we return false, a call to GetReadyEvent will follow which will return the error to the caller.
+        if (IsAborted()) {
+            return false;
+        }
+
+        auto guard = Guard(SpinLock_);
+
+        // You should not add data to a session that has not been started (or that is already completing).
+        YT_VERIFY(State_ == ES3UploadSessionState::Started);
```
AleksandraZh: exception maybe?

yt/yt/server/lib/io/chunk_file_reader.cpp
```diff
@@ -284,43 +435,18 @@ std::vector<TBlock> TChunkFileReader::OnBlocksRead(
...
-                    checksum)
-                    << TErrorAttribute("first_block_index", firstBlockIndex)
-                    << TErrorAttribute("block_count", blockCount);
-            }
-        }
-        blocks.push_back(TBlock(block, blockInfo.Checksum));
-    }
-
-    return blocks;
+    return PhysicalChunkLayoutReader_->DeserializeBlocks(
```
AleksandraZh: ну да, это все как свободные функции выглядят

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,825 @@
...
+            };
+        }
+
+        // There could have been context switches between the last check and this point,
+        // so it does not hurt to check again.
+        if (IsAborted()) {
+            return;
+        }
+
+        // TODO(achulkov2): [PLater] Cancel this future if the session is aborted.
```
achulkov2: Can you drop my `[PSmth]` parts from TODO's? I was triaging todo's, but I think at this point it is rather useless, all of them are to be done later (unless you find some that you think you should do now).

yt/yt/ytlib/chunk_client/medium_descriptor.cpp
```diff
@@ -0,0 +1,177 @@
...
+TS3MediumDescriptor::TS3MediumDescriptor(std::string name, int index, int priority, TS3MediumConfigPtr config)
+    : TMediumDescriptor(std::move(name), index, priority)
+    , Config_(std::move(config))
+{ }
+
+TS3MediumDescriptor::TS3ObjectPlacement TS3MediumDescriptor::GetChunkPlacement(TChunkId chunkId) const
+{
+    return {
+        .Bucket = Config_->Bucket,
+        .Key = Format("chunk-data/%v", chunkId)
```
achulkov2: In our own code, we have already improved this to fit better with original designs, now it looks like this:
```cpp
Format("chunk-data/%02x/%02x/%v", chunkId.ReversedParts8[1], chunkId.ReversedParts8[0], chunkId);
```

Let's do this right away in upstream, it is better in every sense.

achulkov2 (changes requested):
Let's improve chunk storage layout, otherwise overall approach — LGTM. Don't see a point in reviewing in detail, as a lot of the code was written by me originally, so I am biased.

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,832 @@
...
+    //! to be done, like for multi-part uploads.
+    //! NB: When implementing this method, keep in mind that upload may
+    //! have completed since checked before invocation.
+    virtual TFuture<void> AbortIncompleteUpload()
+    {
+        return VoidFuture;
+    }
+
+    //! Abort upload that has successfully uploaded the object.
+    virtual TFuture<void> AbortCompletedUpload()
```
AleksandraZh: I am not sure the user of whis writer should distinguish betweem complete and inclompete upload, especially since the upload could become completed after abort is called

I think we should just have Abort and the writer should take care of the rest

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,832 @@
...
+        return CompletePromise_;
+    }
+
+    //! Returns the current offset within the complete uploaded object.
+    //! Used for sanity-checks.
+    i64 GetCurrentOffset() const
+    {
+        YT_ASSERT_THREAD_AFFINITY_ANY();
+
+        auto guard = Guard(SpinLock_);
```
AleksandraZh: rwlock/guard?

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,832 @@
...
+    TFuture<void> GetReadyEvent() override
+    {
+        return ChunkUploadSession_->GetReadyEvent();
+    }
+
+    TFuture<void> Close(
+        const IChunkWriter::TWriteBlocksOptions& /*options*/,
+        const TWorkloadDescriptor& /*workloadDescriptor*/,
+        const TDeferredChunkMetaPtr& chunkMeta,
+        std::optional<int> /*truncateBlockCount*/) override
```
AleksandraZh: TruncateBlocks?

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,832 @@
...
+        return NErasure::ECodec::None;
+    }
+
+    bool IsCloseDemanded() const override
+    {
+        return false;
+    }
+
+    // TODO(achulkov2): [PLater] Cancellation was written by me before I realized that it should be implemented much later.
+    // So, for now, it is left in a conserved state, and will be finished up at a later time.
```
AleksandraZh: maybe write smth like THROW_UNIMPLEMENTED then? I don't like having half implemented methods

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,832 @@
...
+        // TODO(achulkov2): [PForReview] What should we return here, if anything?
+        YT_UNIMPLEMENTED();
+    }
+
+    TWrittenChunkReplicasInfo GetWrittenChunkReplicasInfo() const override
+    {
+        YT_ASSERT_THREAD_AFFINITY_ANY();
+
+        // This method may only be called if the chunk was closed successfully,
+        // so we can assume that the upload session was completed.
```
AleksandraZh, line 122: let's check that here

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,832 @@
...
+    }
+
+    TFuture<void> GetReadyEvent()
+    {
+        YT_ASSERT_THREAD_AFFINITY_ANY();
+
+        auto promise = NewPromise<void>();
+        promise.TrySetFrom(GetStateFuture());
+        promise.TrySetFrom(UploadWindowSemaphore_->GetReadyEvent());
+        // Both futures we are setting from are uncancelable, but it is wise to show our intent anyway.
```
AleksandraZh: something is wrong with the first part of this sentence

yt/yt/server/lib/s3/chunk_reader.cpp
```diff
@@ -0,0 +1,287 @@
...
+                    std::move(response.Data),
+                    ChunkMetaPlacement_.Key,
+                    ChunkId_,
+                    options.ChunkReaderStatistics,
+                    /*dumpBrokenMeta*/ {});
+
+                auto guard = WriterGuard(MetaLock_);
+
+                if (!ChunkMeta_) {
+                    ChunkMeta_ = metaWithChunkId.ChunkMeta;
```
AleksandraZh: do we use chunkId from metaWithChunkId anywhere? If not, let's change return type to just meta

yt/yt/server/lib/s3/chunk_reader.cpp
```diff
@@ -0,0 +1,287 @@
...
+
+    int localIndex = 0;
+    while (localIndex < std::ssize(blockIndexes)) {
+        auto endLocalIndex = localIndex;
+
+        while (endLocalIndex < std::ssize(blockIndexes) && blockIndexes[endLocalIndex] == blockIndexes[localIndex] + (endLocalIndex - localIndex)) {
+            ++endLocalIndex;
+        }
+
+        blockRanges.push_back({blockIndexes[localIndex], blockIndexes[localIndex] + (endLocalIndex - localIndex)});
```
AleksandraZh: emplace_back

yt/yt/server/lib/s3/chunk_reader.cpp
cherepasshka: no, we don't use it, will remove chunk id

yt/yt/server/lib/s3/chunk_writer.cpp
cherepasshka: We have pretty common practice to check writer's/reader's state with verifies. For example in file writer: https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/server/lib/io/chunk_file_writer.cpp#L65
in replication writer https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/ytlib/chunk_client/replication_writer.cpp#L416

yt/yt/server/lib/s3/chunk_reader.cpp
cherepasshka: нет, он еще в DeserializeMeta в DoGetMeta используется

yt/yt/server/lib/s3/unittests/chunk_reader_writer_ut.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+    void TearDown() override
+    {
+        CleanBuckets();
+    }
+
+    void CleanBuckets()
+    {
+        if (!S3Client_) {
+            // It means that we have skipped this test suite.
```
raundn: This is weird, could you explain when this may happen? There was no call to `SetUp` but there was to `TearDown`?

yt/yt/server/lib/s3/unittests/chunk_reader_writer_ut.cpp
```diff
@@ -0,0 +1,337 @@
...
+        }))
+        .Apply(BIND([&] {
+            auto deferredMeta = New<TDeferredChunkMeta>();
+            deferredMeta->set_type(0);
+            deferredMeta->set_format(0);
+            *deferredMeta->mutable_extensions() = {};
+            return S3ChunkWriter_->Close({}, {}, deferredMeta);
+        }))
+        .Wait(TDuration::Seconds(120));
+
```
raundn, line 229: I think it's a good idea to `EXPECT_TRUE` that the future is set here (i.e. that the `Wait` was enough). This will simplify investigating test failures.

yt/yt/server/lib/s3/unittests/chunk_reader_writer_ut.cpp
```diff
@@ -0,0 +1,337 @@
...
+            return S3ChunkWriter_->GetReadyEvent();
+        }))
+        .Apply(BIND([&] {
+            auto deferredMeta = New<TDeferredChunkMeta>();
+            deferredMeta->set_type(0);
+            deferredMeta->set_format(0);
+            *deferredMeta->mutable_extensions() = {};
+            return S3ChunkWriter_->Close({}, {}, deferredMeta);
+        }))
+        .Wait(TDuration::Seconds(120));
```
raundn: Same here.

yt/yt/server/lib/s3/unittests/chunk_reader_writer_ut.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+    auto pool = NConcurrency::CreateThreadPool(16, "Worker");
+    auto invoker = pool->GetInvoker();
+    auto testCase = GetParam();
+
+    IChunkReader::TReadBlocksOptions readBlockOptions;
+    for (int batchIndex = 0; batchIndex < testCase.BatchCount; ++batchIndex) {
+        std::vector<TFuture<std::vector<TBlock>>> readFutures;
+
+        int requestSize = std::max(std::max(testCase.BlockCountInChunk / testCase.RequestCountInReadBatch, 1), testCase.BlockInRequest);
```
raundn: `std::max` is actually variadic - it may take any number of arguments. So there's no need to nest `std::max` calls.

yt/yt/server/lib/s3/chunk_reader.cpp
```diff
@@ -0,0 +1,269 @@
...
+using namespace NThreading;
+using namespace NIO;
+
+////////////////////////////////////////////////////////////////////////////
+
+namespace {
+
+std::vector<TBlockRange> GetBlockRanges(const std::vector<int>& blockIndexes)
+{
+    std::vector<TBlockRange> blockRanges;
```
raundn, line 31: `YT_VERIFY(!blockIndexes.empty())`? Or return empty result in that case.

yt/yt/server/lib/s3/chunk_reader.cpp
```diff
@@ -0,0 +1,269 @@
...
+{
+public:
+    TS3Reader(
+        IClientPtr client,
+        const TS3MediumDescriptorPtr& mediumDescriptor,
+        TS3ReaderConfigPtr config,
+        TChunkId chunkId)
+        : Client_(std::move(client))
+        , Config_(std::move(config))
+        , ChunkId_(std::move(chunkId))
```
raundn: The move is not necessary.

yt/yt/server/lib/s3/chunk_reader.cpp
```diff
@@ -0,0 +1,269 @@
...
+        , ChunkPlacement_(mediumDescriptor->GetChunkPlacement(ChunkId_))
+        , ChunkMetaPlacement_(mediumDescriptor->GetChunkMetaPlacement(ChunkId_))
```
raundn: Just in case, I want to reiterate: please let's come up with a different term for "placement".

yt/yt/server/lib/s3/chunk_reader.cpp
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    TFuture<TRefCountedChunkMetaPtr> GetMeta(
+        const TGetMetaOptions& /*options*/,
+        const std::optional<TPartitionTags>& /*partitionTags*/,
+        const std::optional<std::vector<int>>& extensionTags) override
+    {
+        if (extensionTags) {
+            YT_LOG_ALERT("Get meta request for S3 media was formed with extension tags (ExtensionTags: %v)", extensionTags);
+            THROW_ERROR_EXCEPTION("Extension tags are not supported in get meta request")
```
raundn: S3 should be mentioned somewhere in the error message.

yt/yt/server/lib/s3/chunk_reader.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+        return DoGetMeta();
+    }
+
+    TChunkId GetChunkId() const override
+    {
+        return ChunkId_;
+    }
+
+    //! This is needed for erasure repair, so no implementation necessary.
```
raundn, line 130: Does this mean we won't be able to store erasure chunks in S3? Or what exactly is the scenario that would lead to calling this method?

yt/yt/server/lib/s3/chunk_reader.cpp
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    TFuture<std::vector<TBlock>> ReadBlockRanges(
+        const TReadBlocksOptions& options,
+        const std::vector<TBlockRange>& blockRanges,
+        const TBlocksExtPtr& blocksExt = nullptr)
+    {
+        if (!blocksExt) {
+            return GetBlocksExt(GetSessionInvoker(options))
+                .Apply(BIND(&TS3Reader::ReadBlockRanges, MakeStrong(this), options, blockRanges).AsyncVia(GetSessionInvoker(options)));
```
raundn: `.AsyncVia` to separate line

yt/yt/server/lib/s3/chunk_reader.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+        for (const auto& blockRange : blockRanges) {
+            futures.emplace_back(ReadBlockRange(options, blockRange, blocksExt));
+        }
+
+        return AllSet(std::move(futures))
+            .Apply(BIND([] (const std::vector<TErrorOr<std::vector<TBlock>>>& results) {
+                std::vector<TBlock> blocks;
+                for (const auto& result : results) {
+                    const auto& rangeBlocks = result.ValueOrThrow();
```
raundn: Since we throw on all errors anyway, why not use `AllSucceeded` instead of `AllSet`?  No need to wait for all blocks to be read only to throw them away.

pavel-bash:
When integrating my replication reader functionality on top of this PR, I have noticed a bug which crashes any `ServerProgram`: please take a look at [this](https://github.com/ytsaurus/ytsaurus/pull/1553/files#diff-68fca3e462fe794b034e47c6b11708bddd6fe3d0a49ea91355544f7414f7a9ffR497) code snippet. It contains a dirty fix to this problem which I propose to solve nicely in this PR as it directly concerns the chunk reader part.

Long story short - we are using a synchronous `Apply` on top of the result of `MakeRequest`. Then we call the `response->GetStatusCode()` which internally does a `WaitFor()` which is forbidden to use in the server program code by the [guard](https://github.com/ytsaurus/ytsaurus/commit/31dbc689962962ed8e2e93a67ec1801eaa8019e5#diff-d36fe2e8e645147ec06bc75198a0a81a99ca5cae0116735879705a5ec2ae3c45R747). I think it's almost guaranteed to happen when running chunk reader from this PR.

I am proposing two options to solve it: the first option I have implemented in my PR [here](https://github.com/ytsaurus/ytsaurus/pull/1553/files#diff-68fca3e462fe794b034e47c6b11708bddd6fe3d0a49ea91355544f7414f7a9ffR498). In the second approach there's less boilerplate code (look [here](https://gist.github.com/pavel-bash/1667e65ef52e13019f1e57a198b544d4)), but another `Apply(..)` on top. I have no strong preference, so please decide which one you like more :)

@cherepasshka

yt/yt/server/lib/s3/unittests/chunk_reader_writer_ut.cpp
```diff
@@ -0,0 +1,337 @@
...
+namespace {
+
+////////////////////////////////////////////////////////////////////////////////
+
+struct TS3TestCase
+{
+    int BatchCount = 1;
+    int BlockCountInChunk = 10;
+    int RequestCountInReadBatch = 10;
+    int BlockInRequest = 2;
```
raundn: BlocksInRequest?

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,810 @@
...
+        TS3MediumDescriptor::TS3ObjectPlacement objectPlacement,
+        const NLogging::TLogger& logger)
+        : Client_(std::move(client))
+        , ObjectPlacement_(std::move(objectPlacement))
+        , Logger(logger.WithTag("Bucket: %v, Key: %v", ObjectPlacement_.Bucket, ObjectPlacement_.Key))
+    { }
+
+    //! Aborts the upload session with the given error.
+    //! This will cancel intermediate upload requests in a best-effort manner.
+    void SetStateError(const TError& error)
```
raundn: SetStateError, StateError_, GetStateFuture and IsAborted are closely tied but their naming fails to convey this. For example, it's not obvious from its name alone that GetStateFuture returns a future that's set if and only if the session is aborted. One may reasonably expect that GetStateFuture() should also be set on session success.
Let me suggest some naming changes:

SetStateError(error) -> SetAbortError
StateError_ -> Abort[Promise|Error]
GetStateFuture -> GetAbortFuture|GetAbortedEvent
IsAborted is fine as it is.

Alternatively, it may be a good idea to change the promise from "abort promise" to "finish promise" and set it on session success, too. That is, to merge GetStateError() and GetCompletionFuture() into one.

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,810 @@
...
+    TFuture<void> GetStateFuture()
+    {
+        return StateError_.ToFuture().ToUncancelable();
+    }
+
+    //! This future should be set when the upload session is completed.
+    virtual TFuture<void> GetCompletionFuture() const = 0;
+
+    //! Abort upload that is not completed, in there is something
+    //! to be done, like for multi-part uploads.
```
raundn: Probably "in there is something" should be "if there is something" but I still fail to grasp the meaning of this commentary.

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,810 @@
...
+    //! NB: When implementing this method, keep in mind that upload may
```
raundn: The method is already implemented. "When overriding..."?

yt/yt/server/lib/s3/chunk_writer.cpp
raundn: I'm with Aleksandra on this. Frankly, a cursory glance at `AbortIncompleteUpload` and `AbortCompletedUpload` leaves me confused. Without getting bogged down in details, it's hard to understand when exactly is one supposed to be called as opposed to the other. I mean, isn't uploading an asynchronous activity? What if I call `AbortIncompleteUpload` but then upload finishes immediately?

I think separating public methods from private (or protected) methods would help here.

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,810 @@
...
+        return Client_->DeleteObjects(TDeleteObjectsRequest{
+            .Bucket = ObjectPlacement_.Bucket,
+            .Objects = {ObjectPlacement_.Key},
+        })
+            .AsVoid();
+    }
+
+    //! Cancel the upload session.
+    //! This method is intended to be best-effort, it is not possible to achieve race-free cancellation.
+    //! Some stray parts/objects may remain in S3 and should be removed separately.
```
raundn: I agree with the first sentence and disagree with the second one :-) I won't urge you to fix this, but may be a TODO is in order here. Because right now "should be removed separately" sounds like "you're on your own, pal" which is akin to admitting defeat =)

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,810 @@
...
+        , Invoker_(std::move(invoker))
+        , UploadWindowSemaphore_(New<TAsyncSemaphore>(Options_.UploadWindowSize))
+    {
+        StartPromise_.TrySetFrom(GetStateFuture());
+        CompletePromise_.TrySetFrom(GetStateFuture());
+        // This seems safer than using individual callbacks for each part upload future.
+        // NB: It is important to specify an invoker here, so that the callback is not
+        // executed synchronously in a thread holding our lock.
+        GetStateFuture()
+            .Subscribe(BIND(&TS3MultiPartUploadSession::CancelPendingUploads, MakeWeak(this)).Via(Invoker_));
```
raundn: `.Via` should be on the next line

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,810 @@
...
+            .Objects = {ObjectPlacement_.Key},
+        })
+            .AsVoid();
+    }
+
+    //! Cancel the upload session.
+    //! This method is intended to be best-effort, it is not possible to achieve race-free cancellation.
+    //! Some stray parts/objects may remain in S3 and should be removed separately.
+    //! Do not call any other methods after this one.
+    TFuture<void> Cancel()
```
raundn: Shouldn't this be called `Abort`? (Or maybe everything that mentions "abort" should instead use "cancel"?)

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,810 @@
...
+    }
+
+    void DoStart()
+    {
+        YT_ASSERT_THREAD_AFFINITY_ANY();
+
+        YT_LOG_DEBUG("Starting multi-part upload to S3");
+
+        if (IsAborted()) {
+            return;
```
raundn: This approach is fragile and error-prone. Strictly speaking, we must check `IsAborted()` every step of  the way - which is not reasonable. For example, just a few lines below we call `WaitFor`. What if someone calls `Cancel`/`SetStateError` while we are waiting? Should we check `IsAborted()` again immediately after waking up?

What I think this code needs is a `TCancelableContext`. First of all, we can use it directly to propagate session abort everywhere in a non-racy way (as much as doing so is at all possible). No more calling TrySetFrom time and time again. Second of all, we can use cancelable context to wrap the invoker into a cancelable invoker. This will solve the `WaitFor` problem.

yt/yt/server/lib/s3/chunk_reader.cpp
cherepasshka: Via this reader we will not be able to store erasure chunks in S3, this reader is strictly for regular chunks (I named creation function CreateS3RegularChunkReader). For erasure chunks we will need another reader, but it is better to do in another PR - I think  first, it's worth finishing the entire pipeline for regular chunks

yt/yt/server/lib/s3/upload_session.h
```diff
@@ -0,0 +1,167 @@
+#pragma once
```
cherepasshka, line 1: I decided to move all upload related classes into separate file to be able to add tests for them

yt/yt/ytlib/chunk_client/config.h
```diff
@@ -257,7 +257,7 @@ struct TS3MediumConfig
     : public NS3::TS3ConnectionConfig
 {
     //! Name of the bucket to use.
-    std::string Bucket;
+    TString Bucket;
```
cherepasshka: NS3::IClient hasn't migrated from TString to std::string yet, so to avoid compile errors in S3 reader and writer I replace here std::string into TString

yt/yt/server/lib/s3/chunk_reader.cpp
```diff
@@ -0,0 +1,273 @@
...
+        return ReadBlockRanges(options, GetBlockRanges(firstBlockIndex, blockCount));
+    }
+
+    TFuture<TRefCountedChunkMetaPtr> GetMeta(
+        const TGetMetaOptions& /*options*/,
+        const std::optional<TPartitionTags>& /*partitionTags*/,
+        const std::optional<std::vector<int>>& extensionTags) override
+    {
+        if (extensionTags) {
+            YT_LOG_ALERT("Get meta request for S3 media was formed with extension tags (ExtensionTags: %v)", extensionTags);
```
AleksandraZh: YT_LOG_ALERT_AND_THROW

yt/yt/server/lib/s3/chunk_reader.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+                return ChunkMeta_;
+            }).AsyncVia(invoker ? invoker : GetCurrentInvoker()));
+    }
+
+    TFuture<TBlocksExtPtr> GetBlocksExt(
+        IInvokerPtr invoker = nullptr)
+    {
+        return DoGetMeta(invoker)
+            .AsVoid()
```
AleksandraZh: maybe remove AsVoid and return extensions from DoGetMeta future? You also would not need a guard in that case

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,200 @@
...
+        , ChunkMetaUploadSession_(New<TS3SimpleUploadSession>(
+            Client_,
+            mediumDescriptor->GetS3ObjectPlacementForChunkMeta(SessionId_.ChunkId),
+            TDispatcher::Get()->GetWriterInvoker(),
+            Logger))
+    { }
+
+    TFuture<void> Open() override
+    {
+        YT_LOG_INFO("Offshore S3 writer opened");
```
AleksandraZh: DEBUG

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,200 @@
...
+        return ChunkUploadSession_->GetReadyEvent();
+    }
+
+    TFuture<void> Close(
+        const IChunkWriter::TWriteBlocksOptions& /*options*/,
+        const TWorkloadDescriptor& /*workloadDescriptor*/,
+        const TDeferredChunkMetaPtr& chunkMeta,
+        std::optional<int> truncateBlockCount) override
+    {
+        // Journal chunks are not supported.
```
AleksandraZh, line 88: What happens if I try to create one living in S3? I think there should be a clear error somewhere

yt/yt/server/lib/s3/chunk_writer.cpp
```diff
@@ -0,0 +1,200 @@
...
+    const IClientPtr Client_;
+    const TSessionId SessionId_;
+
+    const NLogging::TLogger Logger;
+
+    const TS3MultiPartUploadSessionPtr ChunkUploadSession_;
+    const TS3SimpleUploadSessionPtr ChunkMetaUploadSession_;
+
+    const NChunkClient::TRefCountedChunkMetaPtr ChunkMeta_ = New<NChunkClient::TRefCountedChunkMeta>();
+    NChunkClient::NProto::TChunkInfo ChunkInfo_;
```
AleksandraZh, line 170: empty line between const and nonconst

yt/yt/server/lib/s3/upload_session.cpp
```diff
@@ -0,0 +1,465 @@
...
+    YT_LOG_DEBUG("Aborting S3 upload session (State: %v)", completed);
+    return completed
+        ? AbortCompletedUpload()
+        : AbortIncompleteUpload();
+}
+
+bool TS3UploadSessionBase::IsUploadCompleted() const
+{
+    auto completionFuture = GetCompletionFuture();
+    return completionFuture.IsSet() && completionFuture.BlockingGet().IsOK();
```
AleksandraZh: BlockingGet -> GetOrCrash

yt/yt/server/lib/s3/upload_session.cpp
```diff
@@ -0,0 +1,465 @@
...
+}
+
+TFuture<void> TS3UploadSessionBase::AbortIncompleteUpload()
```
AleksandraZh: there should be a todo or something

yt/yt/server/lib/s3/upload_session.cpp
```diff
@@ -0,0 +1,465 @@
...
+
+TFuture<void> TS3MultiPartUploadSession::Start()
+{
+    YT_ASSERT_THREAD_AFFINITY_ANY();
+
+    UploadSessionCancelableContext_ = New<TCancelableContext>();
+    CancelableInvoker_ = UploadSessionCancelableContext_->CreateInvoker(UnderlyingInvoker_);
+
+    // There is no need to schedule this callback twice.
+    if (TryExchangeState(ES3UploadSessionState::Created, ES3UploadSessionState::Starting)) {
```
AleksandraZh, line 105: what happens if the callback is already scheduled, but we recreate context and invoker in lines above?

yt/yt/server/lib/s3/upload_session.cpp
```diff
@@ -0,0 +1,465 @@
...
+    return StartPromise_;
+}
+
+bool TS3MultiPartUploadSession::Add(std::vector<TSharedRef> data)
+{
+    YT_ASSERT_THREAD_AFFINITY_ANY();
+
+    auto guard = WriterGuard(SpinLock_);
+
+    // You should not add data to a session that has not been started (or that is already completing).
```
AleksandraZh: We should not crash if it happens though? Especially since state change can happen from any thread

yt/yt/server/lib/s3/upload_session.cpp
```diff
@@ -0,0 +1,465 @@
...
+
+    return UploadWindowSemaphore_->IsReady();
+}
+
+TFuture<void> TS3MultiPartUploadSession::Complete()
+{
+    YT_ASSERT_THREAD_AFFINITY_ANY();
+
+    // You cannot complete a session that has not been started.
+    YT_VERIFY(GetState() != ES3UploadSessionState::Created);
```
AleksandraZh: let's at least throw with a better error

yt/yt/server/lib/s3/upload_session.cpp
```diff
@@ -0,0 +1,465 @@
...
+    }
+
+    return CompletionPromise_;
+}
+
+TFuture<void> TS3MultiPartUploadSession::GetReadyEvent()
+{
+    YT_ASSERT_THREAD_AFFINITY_ANY();
+
+    auto promise = NewPromise<void>();
```
AleksandraZh: let's save it as a class field, so that we don't create a promise on each GetReadyEvent call (I don't know if there are many)

yt/yt/server/lib/s3/upload_session.cpp
```diff
@@ -0,0 +1,465 @@
...
+        return true;
+    }
+    return false;
+}
+
+void TS3MultiPartUploadSession::SchedulePartUploadIfNeeded()
+{
+    YT_ASSERT_THREAD_AFFINITY_ANY();
+
+    auto guard = ReaderGuard(SpinLock_);
```
AleksandraZh: why Reader?

yt/yt/server/lib/s3/upload_session.cpp
```diff
@@ -0,0 +1,465 @@
...
+
+    if (!BufferedData_.empty() && (BufferedDataSize_ >= Options_.PartSize || State_ == ES3UploadSessionState::Completing)) {
+        GuardedSchedulePartUpload();
+    }
+}
+
+void TS3MultiPartUploadSession::GuardedSchedulePartUpload()
+{
+    YT_ASSERT_THREAD_AFFINITY_ANY();
+    YT_ASSERT_SPINLOCK_AFFINITY(SpinLock_);
```
AleksandraZh: YT_ASSERT_READER/WRITER_SPINLOCK_AFFINITY 
(it looks like this should be a writer lock)

yt/yt/server/lib/s3/upload_session.cpp
```diff
@@ -0,0 +1,465 @@
...
+
+    YT_LOG_DEBUG(
+        "Scheduling part upload (UploadId: %v, PartIndex: %v, ObjectOffset: %v, Size: %v, Md5: %v)",
+        UploadId_,
+        partIndex,
+        CurrentObjectOffset_,
+        partSize,
+        md5);
+
+    auto uploadFuture = Client_->UploadPart(TUploadPartRequest{
```
AleksandraZh, line 249: no need for this variable, you can do Client->Upload().Apply() right away

yt/yt/server/lib/s3/upload_session.cpp
```diff
@@ -0,0 +1,465 @@
...
+    });
+
+    auto uploadFutureWithHandler = uploadFuture
+        .Apply(BIND([weakThis = MakeWeak(this), partIndex, partSize] (const TErrorOr<TUploadPartResponse>& response) {
+            if (auto strongThis = weakThis.Lock()) {
+                return strongThis->OnPartUploadCompleted(response, partIndex, partSize);
+            }
+
+            // Not sure this can actually happen, but let's be safe.
+            THROW_ERROR_EXCEPTION("Multi-part upload session destroyed");
```
AleksandraZh, line 262: why is it an exception though? What exactly are we afraid of, of some garbage getting stuck is S3?

yt/yt/server/lib/s3/upload_session.cpp
```diff
@@ -0,0 +1,465 @@
...
+{
+    CancelableInvoker_ = UploadSessionCancelableContext_->CreateInvoker(UnderlyingInvoker_);
+}
+
+TFuture<void> TS3UploadSessionBase::GetCompletionFuture() const
+{
+    return CompletionPromise_.ToFuture().ToUncancelable();
+}
+
+TFuture<void> TS3UploadSessionBase::Abort(TError error)
```
AleksandraZh, line 38: I think abort should change state to something like Abortred

yt/yt/server/lib/s3/upload_session.cpp
```diff
@@ -0,0 +1,465 @@
...
+        .UploadId = UploadId_,
+        .Parts = std::move(uploadedParts),
+    }));
+
+    if (!multiPartUploadOrError.IsOK()) {
+        Y_UNUSED(Abort(multiPartUploadOrError));
+        return;
+    }
+
+    YT_VERIFY(TryExchangeState(ES3UploadSessionState::Completing, ES3UploadSessionState::Completed));
```
AleksandraZh: It is weird that there is a verify here, but TrySet in a line below

yt/yt/server/lib/s3/upload_session.cpp
```diff
@@ -0,0 +1,465 @@
...
+{
+    {
+        auto guard = ReaderGuard(SpinLock_);
+        // There is no point in aborting if the upload was not started or is already completed.
+        // The latter is unlikely because we check for it before running this method.
+        if (State_ == ES3UploadSessionState::Created || State_ == ES3UploadSessionState::Starting || State_ == ES3UploadSessionState::Completed) {
+            return;
+        }
+    }
+
```
AleksandraZh, line 409: you need to change state here so no more upload is happening while you do that

yt/yt/server/lib/s3/upload_session.cpp
```diff
@@ -0,0 +1,465 @@
...
+    // Accessing UploadId_ is fine, since the value is read-only after the upload is started.
+    YT_VERIFY(!UploadId_.empty());
+
+    YT_LOG_DEBUG("Aborting incomplete multi-part upload to S3 (UploadId: %v)", UploadId_);
+    WaitFor(Client_->AbortMultipartUpload(TAbortMultipartUploadRequest{
```
AleksandraZh, line 414: for each pending upload session you should also delete stuff if it was successfull, you can do Apply to futures for now or something

yt/yt/server/lib/s3/upload_session.cpp
AleksandraZh: okay, I see, there is an override. Shouldn't it be pure virtual then? You can override it in TS3SimpleUploadSession if you want it doing nothing for tests

AleksandraZh: or maybe override Abort in general

yt/yt/server/lib/s3/upload_session.cpp
```diff
@@ -0,0 +1,465 @@
...
+    }
+
+    YT_LOG_DEBUG("Object upload completed (ETag: %v)", putObjectResponse.Value().ETag);
+
+    CompletionPromise_.TrySet();
+}
+
+////////////////////////////////////////////////////////////////////////////
+
+} // namespace NYT::NS3
```
AleksandraZh: newline

yt/yt/server/lib/s3/upload_session.cpp
cherepasshka: For more readability I have saved this variable, but removed uploadFutureWithHandler

yt/yt/server/lib/s3/chunk_writer.cpp
cherepasshka: We will fail on writer/reader creation stage at line `YT_VERIFY(IsRegularChunkId(sessionId.ChunkId));`
Also for readability name of the writer and the reader has "RegularChunk" substring, so this assert will not be unexpected

yt/yt/server/lib/s3/upload_session.cpp
cherepasshka: Fixed

yt/yt/server/lib/s3/upload_session.cpp
cherepasshka: There is no need to do so: AbortMultipartUpload does all that
Also I have unit test for that case (AbortCompletedMultiPartUpload)

yt/yt/server/lib/s3/upload_session.cpp
cherepasshka: Added change of state in TS3MultiPartUploadSession

yt/yt/server/lib/s3/upload_session.cpp
cherepasshka: We are not afraid of anything here, just need to return something since type of the whole future is TFuture < TUploadPartResponse >

yt/yt/server/lib/s3/upload_session.cpp
cherepasshka: Have added override of Abort for TS3MultiPartUploadSession with change of State_

yt/yt/server/lib/s3/upload_session.cpp
cherepasshka: We will lose scheduled DoStart, will move UploadSessionCancelableContext_ and CancelableInvoker_ initialization inside this if, it will fix it

yt/yt/server/lib/s3/upload_session.cpp
```diff
@@ -0,0 +1,475 @@
...
+    return CompletionPromise_;
+}
+
+TFuture<void> TS3MultiPartUploadSession::Abort(TError error)
+{
+    {
+        auto guard = WriterGuard(SpinLock_);
+        State_ = ES3UploadSessionState::Aborting;
+    }
+    return TS3UploadSessionBase::Abort(error)
```
AleksandraZh, line 167: this can easily be scheduled multiple times and will crash

yt/yt/server/lib/s3/upload_session.cpp
```diff
@@ -0,0 +1,475 @@
...
+    auto guard = WriterGuard(SpinLock_);
+    GuardedSchedulePartUploadIfNeeded();
+}
+
+void TS3MultiPartUploadSession::GuardedSchedulePartUploadIfNeeded()
+{
+    YT_ASSERT_THREAD_AFFINITY_ANY();
+    YT_ASSERT_WRITER_SPINLOCK_AFFINITY(SpinLock_);
+
+    YT_VERIFY(State_ == ES3UploadSessionState::Started || State_ == ES3UploadSessionState::Completing);
```
AleksandraZh: SchedulePartUploadIfNeeded from DoComplete is callend async through CancelableInvoker_, which means that when we are here state can already be changed 

also, I do not understand why GuardedSchedulePartUploadIfNeeded is called through CancelableInvoker_ in DoComplete, but not in Add, there should be symmetry I think

 everything doing something with s3 itself (like UploadPart) should deffinetely be done in CancelableInvoker_, otherwise I don't understand why we need it

yt/yt/server/lib/s3/upload_session.cpp
```diff
@@ -0,0 +1,475 @@
...
+
+    if (!BufferedData_.empty() && (BufferedDataSize_ >= Options_.PartSize || State_ == ES3UploadSessionState::Completing)) {
+        GuardedSchedulePartUpload();
+    }
+}
+
+void TS3MultiPartUploadSession::GuardedSchedulePartUpload()
+{
+    YT_ASSERT_THREAD_AFFINITY_ANY();
```
AleksandraZh: let's make an affinily for CancelableInvoker_

yt/yt/server/lib/s3/upload_session.h
```diff
@@ -0,0 +1,174 @@
...
+        IClientPtr client,
+        NChunkClient::TS3MediumDescriptor::TS3ObjectPlacement objectPlacement,
+        IInvokerPtr invoker,
+        const NLogging::TLogger& logger);
+
+    //! Returns future that should become set when the upload session is completed.
+    TFuture<void> GetCompletionFuture() const;
+
+    //! Aborts the upload session.
+    //! This method is intended to be best-effort, it is not possible to achieve race-free cancellation.
```
AleksandraZh, line 29: I think it should not be best-effort in production ready code

It is ok for it to behave like this now, but then it should be a TODO, but I actually think it should almost works already

cherepasshka: Added todo

yt/yt/server/lib/s3/upload_session.cpp
cherepasshka: Fixed

cherepasshka:
Ship!

cherepasshka:
Shipped!

cherepasshka:
Shipped!

cherepasshka:
Shipped!

PR #1529, opened 2025-12-01
Fix helm-chart installation command

Dakyz:


ogorbacheva:
Shipped!

ogorbacheva:
Already fixed in another PR

PR #1535, opened 2025-12-11
Remove extra slash from include

mikhainin:
---
>
> * Changelog entry
>               # cpp-sdk/go-sdk/cms/excel/cron/microservices (Select one value, example: `Component: scheduler`)

* Changelog entry
Type: fix
Component: logging

`clang` gets confused with the extra slash and keeps failing

```
In file included from /workspace/third_party/ytsaurus-cpp-sdk/library/cpp/yt/logging/plain_text_formatter/formatter.h:5:
/workspace/third_party/ytsaurus-cpp-sdk/library/cpp/yt/logging//logger.h:81:1: error: redefinition of 'GetEnumTraitsImpl'
...
/workspace/third_party/ytsaurus-cpp-sdk/library/cpp/yt/logging/logger.h:81:1: note: previous definition is here
   81 | DEFINE_ENUM(ELogMessageKind,
```

### DIFF
--- a/library/cpp/yt/logging/plain_text_formatter/formatter.h
+++ b/library/cpp/yt/logging/plain_text_formatter/formatter.h
@@ -2,7 +2,7 @@
 
 #include <library/cpp/yt/string/raw_formatter.h>
 
-#include <library/cpp/yt/logging//logger.h>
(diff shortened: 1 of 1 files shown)


ilyaibraev:
LGTM

PR #1536, opened 2025-12-11
yt/core/crypto: use CA certificates from environment variable SSL_CERT_FILE

koct9i:
- **yt/yt/core/crypto: use built-in CA when config is not specified**
- **yt/yt/core/crypto: use CA certificates from environment variable SSL_CERT_FILE**

---

* Changelog entry
Type: feature
Component: cpp-sdk

Use CA certificates set by user in SSL_CERT_FILE instead of default built-in.
This aligns with most modern design, golang and curl behaviour.


koct9i:
The same for python - #1525

PR #1537, opened 2025-12-11
yt/yt/core/https: handle plain http connections too

koct9i:
Now https client is able to handle "http://..." URLs too.
Off by default, requires setting AllowHTTP = true in config.

---

* Changelog entry
Type: feature
Component: cpp-sdk

Add option to handle both "http://" and "https://" URLs by single client.

### DIFF
--- a/yt/yt/core/net/dialer.h
+++ b/yt/yt/core/net/dialer.h
@@ -20,6 +20,8 @@ struct TDialerContext final
 {
     //! Host is used for TlsDialer.
     std::optional<TString> Host;
+    //! BypassTLS is used by HTTPS client for plain HTTP connections.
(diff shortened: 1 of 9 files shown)


yt/yt/core/net/dialer.h
```diff
@@ -20,6 +20,8 @@ struct TDialerContext final
 {
     //! Host is used for TlsDialer.
     std::optional<TString> Host;
+    //! BypassTLS is used by HTTPS client for plain HTTP connections.
+    bool BypassTLS = false;
```
yalekseev, line 24: May be TryBypassTLS? It does not necessary bypasses TLS depending on dialer config.

yalekseev:
Shipped!

PR #1544, opened 2025-12-17
Support DROP ... IF EXISTS & CREATE ... IF NOT EXISTS features.

Tony-Romanov:
Type: feature
Component: query-tracker


yt/yql/providers/yt/expr_nodes/yql_yt_expr_nodes.json
```diff
@@ -396,12 +396,18 @@
         {
             "Name": "TYtDropTable",
             "Base": "TYtIsolatedOpBase",
-            "Match": {"Type": "Callable", "Name": "YtDropTable!"}
+            "Match": {"Type": "Callable", "Name": "YtDropTable!"},
```
nepal, line 399: Let's make settings non-optional and add them in type annotation in case of callable with missing settings

Tony-Romanov: Done.

PR #1546, opened 2025-12-18
Physical chunk reader writer refactoring

cherepasshka:
* Changelog entry
Type: refactoring
Component: server/lib

Refactored physical chunk reader and writer and moved part responsible for physical chunk layout

### DIFF
--- a/yt/yt/server/lib/io/unittests/physical_chunk_layout_ut.cpp
+++ b/yt/yt/server/lib/io/unittests/physical_chunk_layout_ut.cpp
@@ -92,7 +92,7 @@ TEST(TPhysicalChunkLayout, SerializeAndDeserializeMeta)
         chunkId,
         /*dumpBrokenMeta*/ {});
 
-    EXPECT_TRUE(google::protobuf::util::MessageDifferencer::Equals(*deserializedMeta.ChunkMeta, *finalizedMeta));
+    EXPECT_TRUE(google::protobuf::util::MessageDifferencer::Equals(*deserializedMeta, *finalizedMeta));
 }
 
 ////////////////////////////////////////////////////////////////////////////////


yt/yt/server/lib/io/chunk_file_reader.cpp
```diff
@@ -54,17 +51,172 @@ void ReadHeader(
...
+TPhysicalChunkLayoutReader::TPhysicalChunkLayoutReader(
+    TChunkId chunkId,
+    TString chunkFileName,
+    TString chunkMetaFileName,
+    const TOptions& options,
+    NLogging::TLogger logger,
+    TDumpBrokenBlockCallback dumpBrokenBlockCallback,
+    TDumpBrokenMetaCallback dumpBrokenMetaCallback)
+    : ChunkFileName_(std::move(chunkFileName))
+    , ChunkMetaFileName_(std::move(chunkMetaFileName))
```
AleksandraZh: а зачем мы имена файлов в этот класс сложили? Как будто мы их только логируем и наружу показвыаем

yt/yt/server/lib/io/chunk_file_reader.cpp
```diff
@@ -54,17 +51,172 @@ void ReadHeader(
...
+    , Options_(options)
+    , Logger(logger.WithTag("ChunkId: %v", chunkId))
+    , ChunkId_(chunkId)
```
AleksandraZh: про chunk id тот же вопрос

yt/yt/server/lib/io/chunk_file_reader.h
```diff
@@ -56,6 +56,74 @@ DEFINE_ENUM(EDirectIOFlag,
...
+        TBlockRange blockRange,
+        const TBlocksExtPtr& blocksExt,
+        NChunkClient::TChunkReaderStatisticsPtr chunkReaderStatistics);
+
+    const TString& GetChunkFileName() const;
+    const TString& GetChunkMetaFileName() const;
+    NChunkClient::TChunkId GetChunkId() const;
+
+private:
+    const TString ChunkFileName_;
```
AleksandraZh: вообще довольно странный класс -- тут как будто нет внутреннего состояния (а то, которое есть, мы никак не используем и просто возвращаем наружу), а тут скорее набор "свободных функций"

yt/yt/server/lib/io/chunk_file_reader.h
```diff
@@ -56,6 +56,43 @@ DEFINE_ENUM(EDirectIOFlag,
 
 ////////////////////////////////////////////////////////////////////////////////
 
+using TDumpBrokenMetaCallback = TCallback<void(TRef /*block*/)>;
```
AleksandraZh, line 60: in this pr these can stay here, but we should probably move them to a separate file in s3 readers/writes pr

yt/yt/server/lib/io/chunk_file_writer.h
```diff
@@ -20,6 +20,23 @@ namespace NYT::NIO {
...
+    std::vector<TSharedRef> Buffers;
+};
+
+i64 TruncateBlocks(NChunkClient::NProto::TBlocksExt& blocksExt, int truncateBlockCount, i64 oldDataSize);
+
+TWriteBlocksRequest MakeWriteRequest(i64 startOffset, const std::vector<NChunkClient::TBlock>& blocks, NChunkClient::NProto::TBlocksExt& blocksExt);
+
+NChunkClient::TRefCountedChunkMetaPtr FinalizeChunkMeta(NChunkClient::TDeferredChunkMetaPtr chunkMeta, const NChunkClient::NProto::TBlocksExt& blocksExt);
+
+TSharedMutableRef PrepareChunkMetaBlob(NChunkClient::TChunkId chunkId, const NChunkClient::TRefCountedChunkMetaPtr& chunkMeta);
```
AleksandraZh: Maybe PrepareChunkMetaBlob -> SerializeChunkMeta 
MakeWriteRequest -> SerializeBlocks
to preserve some symetry with reader

yt/yt/server/lib/io/chunk_file_writer.h
```diff
@@ -20,6 +20,23 @@ namespace NYT::NIO {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+struct TWriteBlocksRequest
```
AleksandraZh: TSerializedBlocks/Data/Smth/?

yt/yt/server/lib/io/chunk_file_reader.cpp
```diff
@@ -54,6 +55,131 @@ void ReadHeader(
...
+                    "Incorrect checksum of block %v in chunk data file %v: expected %v, actual %v",
+                    blockIndex,
+                    chunkFileName,
+                    blockInfo.Checksum,
+                    checksum)
+                    << TErrorAttribute("first_block_index", blockRange.StartBlockIndex)
+                    << TErrorAttribute("block_count", blockRange.EndBlockIndex - blockRange.StartBlockIndex);
+            }
+        }
+        blocks.push_back(TBlock(block, blockInfo.Checksum));
```
AleksandraZh: emplace_back?

yt/yt/server/lib/io/chunk_file_reader.cpp
```diff
@@ -54,6 +55,131 @@ void ReadHeader(
...
+std::vector<TBlock> DeserializeBlocks(
+    TSharedRef blocksBlob,
+    TBlockRange blockRange,
+    bool validateBlockChecksums,
+    const TString& chunkFileName,
+    const TBlocksExtPtr& blocksExt,
+    NChunkClient::TChunkReaderStatisticsPtr chunkReaderStatistics,
+    TDumpBrokenBlockCallback dumpBrokenBlockCallback)
+{
+    chunkReaderStatistics->DataBytesReadFromDisk.fetch_add(
```
AleksandraZh: I think these should be in chunk file writer, not sure DataBytesReadFromDisk is applicable to S3

yt/yt/server/lib/io/unittests/physical_chunk_layout_ut.cpp
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    return blocks;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+TEST(TPhysicalChunkLayout, SerializeAndDeserializeBlocks)
+{
+    constexpr int BlocksCount = 100;
```
raundn, line 51: BlockCount

yt/yt/server/lib/io/unittests/physical_chunk_layout_ut.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+    auto chunkId = MakeRandomId(NCypressClient::EObjectType::Chunk, NObjectClient::TCellTag(0xf003));
+    auto generator = TRandomGenerator(42);
```
raundn, line 54: I know using `42` here is kind of a pattern in our unit tests by now, but I'd argue it's not a good pattern. Unless determinism is crucial, it's better to use something like `TInstant::Now().GetValue()` to seed the RNG and thus provide testing data with some variety. It's also a good idea to log the seed in order to be able to reproduce the test exactly in case the RNG turns out important and the test crashes in some unlikely scenario.

yt/yt/server/lib/io/chunk_file_reader.h
```diff
@@ -56,6 +56,36 @@ DEFINE_ENUM(EDirectIOFlag,
 
 ////////////////////////////////////////////////////////////////////////////////
 
+// TODO(cherepashka): move these usings below somewhere more suitable for S3 chunk readers & writers.
+using TDumpBrokenMetaCallback = TCallback<void(TRef /*block*/)>;
+using TDumpBrokenBlockCallback = TCallback<void(int /*blockIndex*/, const NIO::TBlockInfo& /*blockInfo*/, TRef /*block*/)>;
+
+//! Deserializes chunk meta from blob with format validation.
+//! For chunk meta version 2+, the local chunk id stored in this class is validated against the one
```
raundn, line 64: > the local chunk id stored in this class

In which class? :-) The commentary should be updated, it would seem.

yt/yt/server/lib/io/chunk_file_reader.h
```diff
@@ -56,6 +56,36 @@ DEFINE_ENUM(EDirectIOFlag,
...
+//! stored in the meta file. Passing NullChunkId to this class suppresses this check.
```
raundn, line 65: > Passing NullChunkId to this class

... to this function?

yt/yt/server/lib/io/unittests/physical_chunk_layout_ut.cpp
```diff
@@ -0,0 +1,101 @@
...
+TEST(TPhysicalChunkLayout, SerializeAndDeserializeMeta)
+{
+    auto chunkId = MakeRandomId(NCypressClient::EObjectType::Chunk, NObjectClient::TCellTag(0xf003));
+
+    NChunkClient::NProto::TBlocksExt protoBlocksExt;
+
+    auto finalizedMeta = FinalizeChunkMeta(New<NChunkClient::TDeferredChunkMeta>(), protoBlocksExt);
+    auto metaBlob = SerializeChunkMeta(chunkId, finalizedMeta);
+
+    auto deserializedMeta = DeserializeMeta(
```
raundn, line 89: There's a certain asymmetry in naming between `SerializeChunkMeta` and `{S,Des}erializeMeta`. I haven't read the code ahead yet, but should the latter be named `{S,Des}erializeChunkMeta`?

yt/yt/server/lib/io/chunk_file_reader.h
```diff
@@ -56,6 +56,36 @@ DEFINE_ENUM(EDirectIOFlag,
 
 ////////////////////////////////////////////////////////////////////////////////
 
+// TODO(cherepashka): move these usings below somewhere more suitable for S3 chunk readers & writers.
+using TDumpBrokenMetaCallback = TCallback<void(TRef /*block*/)>;
+using TDumpBrokenBlockCallback = TCallback<void(int /*blockIndex*/, const NIO::TBlockInfo& /*blockInfo*/, TRef /*block*/)>;
```
raundn, line 61: I dont' think these should be called "\*Dump\*" - the dumping is not part of the contract.

For `DeserializeMeta` and `DeserializeBlocks`, these callbacks are just that - opaque callbacks. They may dump the corrupted data, they may just log the fact the corruption was encountered or they may ignore the call altogether. I'd call them `TBrokenBlockCallback` or `TOnBrokenBlock` or `TBeforeIncorrectChecksumExceptionThrown` - you get the idea.

yt/yt/server/lib/io/chunk_file_reader.h
```diff
@@ -56,6 +56,36 @@ DEFINE_ENUM(EDirectIOFlag,
...
+
+//! Deserializes chunk meta from blob with format validation.
+//! For chunk meta version 2+, the local chunk id stored in this class is validated against the one
+//! stored in the meta file. Passing NullChunkId to this class suppresses this check.
+NChunkClient::TRefCountedChunkMetaPtr DeserializeMeta(
+    TSharedRef metaFileBlob,
+    const std::string& chunkMetaFilename,
+    NChunkClient::TChunkId chunkId,
+    TDumpBrokenMetaCallback dumpBrokenMeta);
```
raundn, line 70: onBrokenMeta

yt/yt/server/lib/io/chunk_file_reader.h
```diff
@@ -56,6 +56,36 @@ DEFINE_ENUM(EDirectIOFlag,
...
+};
+
+//! Deserializes chunk blocks from blob with optional checksum validation.
+std::vector<NChunkClient::TBlock> DeserializeBlocks(
+    TSharedRef blocksBlob,
+    TBlockRange blockRange,
+    bool validateBlockChecksums,
+    const std::string& chunkFileName,
+    const TBlocksExtPtr& blocksExt,
+    TDumpBrokenBlockCallback dumpBrokenBlockCallback);
```
raundn, line 85: onBrokenBlock

yt/yt/server/lib/io/chunk_file_reader.cpp
```diff
@@ -54,6 +55,117 @@ void ReadHeader(
...
+    for (int blockIndex = blockRange.StartBlockIndex; blockIndex < blockRange.EndBlockIndex; ++blockIndex) {
+        const auto& blockInfo = blocksExt->Blocks[blockIndex];
+        auto block = blocksBlob.Slice(
+            blockInfo.Offset - firstBlockInfo.Offset,
+            blockInfo.Offset - firstBlockInfo.Offset + blockInfo.Size);
+        if (validateBlockChecksums) {
+            auto checksum = GetChecksum(block);
+            if (checksum != blockInfo.Checksum) {
+                if (dumpBrokenBlockCallback) {
+                    dumpBrokenBlockCallback(blockIndex, blockInfo, block);
```
raundn, line 148: Is `dumpBrokenBlockCallback` allowed to throw? If not, it's a good idea to try-catch and log here.

yt/yt/server/lib/io/chunk_file_reader.h
```diff
@@ -56,6 +56,36 @@ DEFINE_ENUM(EDirectIOFlag,
...
+    NChunkClient::TChunkId chunkId,
+    TDumpBrokenMetaCallback dumpBrokenMeta);
+
+struct TBlockRange
+{
+    i64 StartBlockIndex = 0;
+    i64 EndBlockIndex = 0;
+};
+
+//! Deserializes chunk blocks from blob with optional checksum validation.
```
raundn, line 78: It's not obvious that `DeserializeBlocks` will throw if it encounters a broken block - even when non-null `dumpBrokenBlockCallback` is provided. In other words, it's not obvious that `dumpBrokenBlockCallback` is not supposed to throw anything and, instead, the error will be thrown unconditionally. 

At the least, this should be mentioned in the commentary. At the most, we should consider merging `dumpBrokenBlockCallback` and the code throwing the error. I'm not sure this is good idea - just something to consider. (I see a few benefits of doing it this way. First, `chunkFileName` will become unnecessary as it will get hidden inside the callback. Second, this will make `DeserializeBlocks` more flexible - by providing a non-throwing callback, one will be able do a best-effort deserialization - simply skipping all broken blocks and deserializing everything possible. Again, this is just a suggestion - I'm not sure it's worth it.)

yt/yt/server/lib/io/chunk_file_reader.cpp
```diff
@@ -445,70 +537,20 @@ TRefCountedChunkMetaPtr TChunkFileReader::OnMetaRead(
...
+    auto meta = DeserializeMeta(
+        metaFileBlob,
+        metaFileName,
+        ChunkId_,
+        BIND(&TChunkFileReader::DumpBrokenMeta, MakeWeak(this)));
+    
     chunkReaderStatistics->MetaBytesReadFromDisk.fetch_add(
         metaFileBlob.Size(),
         std::memory_order::relaxed);
     chunkReaderStatistics->MetaIORequests.fetch_add(
```
raundn, line 549: I suspect it's better to update these statistics *before* calling `DeserializeMeta` because if it throws (and there're many reasons it may throw) we'll never update the statistics despite having actually consumed some IO.

yt/yt/server/lib/io/chunk_file_reader.h
raundn: UPDATE: having now read both `DeserializeBlocks` and `DeserializeMeta`, I see that a single "handle broken data (likely by throwing)" callback won't be enough because there're several different cases `DeserializeMeta` has to throw in. Doing what I've suggested above is still possible by replacing the callback with and interface (`class IOnBrokenCallback { void OnIncorrectHeaderSignature(); void OnIncorrectChecksum(); ... };`) but the change really starts to look like it's not worth it now.

PR #1552, opened 2025-12-22
Arrow string types should be represented as YT utf8 and arrow binary types as YT string

achulkov2:
* Changelog entry
Type: fix
Component: misc-server

Arrow string types should be represented as YT utf8 and arrow binary types as YT string.


savnadya:
Shipped!

PR #1553, opened 2025-12-23
Introduce the OffshoreDataGateway component

pavel-bash:
# Description

This PR adds the `OffshoreDataGateway` component which is used when reading data from S3. This component is also integrated into the replication reader to actually perform the reads.

One of the most important topics to note in this work is the introduction of the new proto message `TChunkReplicaSpec` and its usage in the `OffshoreDataGateway`'s RPC requests. This structure allows us to pass the medium index which is required to find the specific offshore medium where the chunk is located. Maybe there are other solutions, I'll be happy to discuss that.

Another topic is the approach to handling different media in the replication reader. Right now the replication reader uses only addresses to uniquely identify different peers to read from. This works because we read from data nodes only, and we do not care about the medium there. Now the reads may also happen from `OffshoreDataGateway`-s, and the medium starts to matter.

Imagine a scenario when a chunk has two offshore replicas on two different mediums, one pointing to Google S3 storage, another to AWS S3. We must have a way to differentiate those two replicas, even though the address is the same - the sentinel `OffshoreNodeAddress`. This is why I introduce a structure called `TPeerId` which includes both the address and the medium index, and now replication reader works with it to differentiate different replicas.

Last topic is the testing. It's impossible to implement an "honest" integration test at the moment as writes to offshore replicas are not implemented, and also masters know nothing of them. I have implemented a C++ unit test to check the behaviour of replication reader only - `test_s3_data.cpp`. However, it's not finished due to another issue - the mediums loose their "domestic VS offshore" differentiation when being serialized, and it makes it impossible to use S3 medium in OffshoreDataGateway. This should probably be fixed in the PR which introduces the mediums.

### DIFF
--- a/yt/yt/library/s3/client.cpp
+++ b/yt/yt/library/s3/client.cpp
@@ -494,32 +494,25 @@ private:
                 &req,
                 CredentialProvider_);
 
-            // TODO(PR): this should probably be a separate/another PR fix and look much cleaner. But it works now.
             return Client_->MakeRequest(std::move(req))
-                .AsUnique()
-                .Apply(BIND([this, this_ = MakeStrong(this), method = req.Method] (TErrorOr<NHttp::IResponsePtr>&& responseOrError) -> TFuture<TCommandResponse> {
-                    auto promise = NewPromise<TCommandResponse>();
+                .AsUnique().Apply(BIND([method = req.Method] (TErrorOr<NHttp::IResponsePtr>&& responseOrError) -> TFuture<TCommandResponse> {
+                    if (!responseOrError.IsOK()) {
+                        return MakeFuture<TCommandResponse>(
+                            TError("HTTP request failed") << std::move(responseOrError));
+                    }
+                    auto& response = responseOrError.Value();
 
-                    ExecutionInvoker_->Invoke(BIND([promise, method, responseOrError = std::move(responseOrError)] () {
-                        if (!responseOrError.IsOK()) {
-                            promise.Set(TError("HTTP request failed") << std::move(responseOrError));
-                            return;
-                        }
+                    // 3xx are not really errors but we don't handle redirects anyways
+                    if (ToUnderlying(response->GetStatusCode()) >= 300) {
+                        return MakeFuture<TCommandResponse>(
+                            ErrorFromResponse(method, response));
+                    }
 
-                        auto& response = responseOrError.Value();
-                        // 3xx are not really errors but we don't handle redirects anyways
-                        if (ToUnderlying(response->GetStatusCode()) >= 300) {
-                            promise.Set(ErrorFromResponse(method, response));
-                            return;
-                        }
-
-                        TCommandResponse rsp;
-                        rsp.Deserialize(response);
-                        promise.Set(std::move(rsp));
-                    }));
-
-                    return promise.ToFuture();
-                }));
+                    TCommandResponse rsp;
+                    rsp.Deserialize(response);
+                    return MakeFuture<TCommandResponse>(std::move(rsp));
+                })
+                    .AsyncVia(ExecutionInvoker_));
         })
             .AsyncVia(ExecutionInvoker_)
             .Run(std::move(req));


yt/yt/client/node_tracker_client/node_directory.cpp
```diff
@@ -726,6 +736,49 @@ const TAddressMap& GetAddressesOrThrow(const TNodeAddressMap& nodeAddresses, EAd
 
 ////////////////////////////////////////////////////////////////////////////////
 
+const TNodeDescriptor* FindPotentiallyOffshoreNodeDescriptor(const TNodeDirectoryPtr& nodeDirectory, TNodeId id)
+{
+    if (id == OffshoreNodeId) {
+        return &OffshoreNodeDescriptor();
```
cherepasshka: I think it is dangerous: you return pointer to an object that will be destroyed after FindPotentiallyOffshoreNodeDescriptor finishes. Better to create a global static variable / or local to this function static variable and to return pointer to it

yt/yt/client/node_tracker_client/node_directory.cpp
```diff
@@ -726,6 +736,49 @@ const TAddressMap& GetAddressesOrThrow(const TNodeAddressMap& nodeAddresses, EAd
...
+
+const TNodeDescriptor& GetPotentiallyOffshoreNodeDescriptor(const TNodeDirectoryPtr& nodeDirectory, TNodeId id)
+{
+    if (id == OffshoreNodeId) {
+        return OffshoreNodeDescriptor();
+    }
+    return nodeDirectory->GetDescriptor(id);
+}
+
+const TNodeDescriptor* FindPotentiallyOffshoreNodeDescriptor(const TNodeDirectoryPtr& nodeDirectory, const std::string& address)
```
cherepasshka: Is this function and GetPotentiallyOffshoreNodeDescriptor used anywhere? I see the usage only of function that accepts node id instead of address

yt/python/yt/environment/yt_env.py
```diff
@@ -680,6 +685,9 @@ def start(self, on_masters_started_func=None):
             if self.yt_config.replicated_table_tracker_count > 0:
                 self.start_replicated_table_trackers(sync=False)
 
+            if self.yt_config.offshore_data_gateway_count > 0:
+                self.start_offshore_data_gateways(sync=False)
```
cherepasshka, line 689: [optional]
I think it will be useful to add TODO for future to add `modify_offshore_data_gateway_config`, somewhere [here](https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/tests/library/yt_env_setup.py#L446) and [here](https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/tests/library/yt_env_setup.py#L1215) or add it in this PR

yt/yt/server/lib/nbd/random_access_file_reader.cpp
```diff
@@ -428,7 +428,7 @@ class TRandomAccessFileReader
                 New<TRemoteReaderOptions>(),
                 ChunkReaderHost_,
                 FromProto<TChunkId>(chunkSpec.chunk_id()),
-                /*seedReplicas*/ {});
+                /*seedReplicas*/ TChunkReplicaWithMediumList{});
```
cherepasshka: Is this change really necessary, can you please explain why?

yt/yt/server/offshore_data_gateway/bootstrap.cpp
```diff
@@ -0,0 +1,316 @@
...
+using namespace NYson;
+using namespace NFusion;
+using namespace NNodeTrackerClient;
+using namespace NOrchid;
+using namespace NProfiling;
+using namespace NAlertManager;
+
+////////////////////////////////////////////////////////////////////////////////
+
+static constexpr auto& Logger = OffshoreDataGatewayLogger;
```
cherepasshka: We use `constinit const auto Logger` more often

yt/yt/server/offshore_data_gateway/bootstrap.cpp
```diff
@@ -0,0 +1,316 @@
...
+            /*threadNamePrefix*/ "Storage"))
+    {
+        if (Config_->AbortOnUnrecognizedOptions) {
+            AbortOnUnrecognizedOptions(Logger(), Config_);
+        } else {
+            WarnForUnrecognizedOptions(Logger(), Config_);
+        }
+    }
+
+    TFuture<void> Run() final
```
cherepasshka: Probably `override` instead of `final`?

yt/yt/server/offshore_data_gateway/bootstrap.cpp
```diff
@@ -0,0 +1,316 @@
...
+private:
+    const TOffshoreDataGatewayBootstrapConfigPtr Config_;
+    const INodePtr ConfigNode_;
+    const IServiceLocatorPtr ServiceLocator_;
+
+    const NConcurrency::TActionQueuePtr ControlQueue_;
+    const IInvokerPtr ControlInvoker_;
+    const TOffshoreDataGatewayDynamicConfigPtr DynamicConfig_;
+
+    const NConcurrency::IThreadPoolPtr StorageThreadPool_;
```
cherepasshka: You have used `NConcurrency` namespace, so you can omit it

yt/yt/server/offshore_data_gateway/bootstrap.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+    TString InstanceId_;
+
+    NMonitoring::IMonitoringManagerPtr MonitoringManager_;
+    NYT::NBus::IBusServerPtr BusServer_;
```
cherepasshka: We almost never use `NYT::` prefix

yt/yt/server/offshore_data_gateway/bootstrap.cpp
```diff
@@ -0,0 +1,316 @@
...
+    const INodePtr ConfigNode_;
+    const IServiceLocatorPtr ServiceLocator_;
+
+    const NConcurrency::TActionQueuePtr ControlQueue_;
+    const IInvokerPtr ControlInvoker_;
+    const TOffshoreDataGatewayDynamicConfigPtr DynamicConfig_;
+
+    const NConcurrency::IThreadPoolPtr StorageThreadPool_;
+
+    TString InstanceId_;
```
cherepasshka: We are gradually migrating to std::string, and as I see [BuildServiceAddress returns](https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/core/net/address.h#L28) std::string, so I think it is better to make InstanceId_'s type std::string

yt/yt/server/offshore_data_gateway/bootstrap.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+    void DoRun()
+    {
+        DoInitialize();
+        DoStart();
+    }
+
+    void DoInitialize()
+    {
+        YT_LOG_INFO("Starting offshore data gateway process");
```
cherepasshka: You can move this log below and add InstanceId_ into log message

yt/yt/server/offshore_data_gateway/bootstrap.cpp
```diff
@@ -0,0 +1,316 @@
...
+            Config_->ClusterConnection,
+            std::move(connectionOptions));
+
+        SetupClusterConnectionDynamicConfigUpdate(
+            NativeConnection_,
+            Config_->ClusterConnectionDynamicConfigPolicy,
+            ConfigNode_->AsMap()->GetChildOrThrow("cluster_connection"),
+            Logger());
+
+        NativeConnection_->GetClusterDirectorySynchronizer()->Start();
```
cherepasshka: Since you have separated initialization part from start part, these starts and start of DynamicConfigManager_ will be better to move into `DoStart` method

yt/yt/server/offshore_data_gateway/bootstrap.cpp
```diff
@@ -0,0 +1,316 @@
...
+            ControlInvoker_,
+            StorageThreadPool_->GetInvoker(),
+            NYT::NBus::TTcpDispatcher::Get()->GetXferPoller(),
+            NativeAuthenticator_,
+            NativeConnection_->GetMediumDirectory(),
+            NativeConnection_->GetMediumDirectorySynchronizer()));
+    }
+
+    void DoStart()
+    {
```
cherepasshka, line 241: I think it will be good to add YT_ASSERT_INVOKER_AFFINITY for control invoker here and for DoRun and DoInitialize()

yt/yt/server/offshore_data_gateway/bootstrap.cpp
```diff
@@ -0,0 +1,316 @@
...
+        const TOffshoreDataGatewayDynamicConfigPtr& oldConfig,
+        const TOffshoreDataGatewayDynamicConfigPtr& newConfig)
+    {
+        TSingletonManager::Reconfigure(newConfig);
+
+        StorageThreadPool_->SetThreadCount(newConfig->StorageThreadCount);
+
+        YT_LOG_DEBUG(
+            "Updated offshore data gateway dynamic config (OldConfig: %v, NewConfig: %v)",
+            ConvertToYsonString(oldConfig, EYsonFormat::Text),
```
cherepasshka, line 301: Probably it will be too fat to convert configs into yson here

yt/yt/server/offshore_data_gateway/bootstrap.cpp
```diff
@@ -0,0 +1,316 @@
...
+            } else {
+                YT_LOG_DEBUG(error, "Error updating Cypress node");
+            }
+        }
+    }
+
+    void OnDynamicConfigChanged(
+        const TOffshoreDataGatewayDynamicConfigPtr& oldConfig,
+        const TOffshoreDataGatewayDynamicConfigPtr& newConfig)
+    {
```
cherepasshka, line 292: I think it will be good to mark this method thread affinity as well

yt/yt/server/offshore_data_gateway/config.h
```diff
@@ -0,0 +1,77 @@
...
+class TOffshoreDataGatewayBootstrapConfig
+    : public NServer::TNativeServerBootstrapConfig
+{
+public:
+    bool AbortOnUnrecognizedOptions;
+
+    NYTree::IMapNodePtr CypressAnnotations;
+
+    NDynamicConfig::TDynamicConfigManagerConfigPtr DynamicConfigManager;
+    TString DynamicConfigPath;
```
cherepasshka: TDynamicConfigManagerOptions::ConfigPath is `NYPath::TYPath`, so I think it is better to make DynamicConfigPath's type `NYPath::TYPath` as well

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
```diff
@@ -0,0 +1,245 @@
...
+        , S3Poller_(std::move(s3Poller))
+        , MediumDirectory_(std::move(mediumDirectory))
+        , MediumDirectorySynchronizer_(std::move(mediumDirectorySynchronizer))
+    {
+        RegisterMethod(RPC_SERVICE_METHOD_DESC(GetBlockSet));
+        RegisterMethod(RPC_SERVICE_METHOD_DESC(GetChunkMeta));
+    }
+
+private:
+    IInvokerPtr StorageInvoker_;
```
cherepasshka: Since these fields are never reassigned, I think it is better to mark them as const

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
```diff
@@ -0,0 +1,245 @@
...
+        const RequestType& request,
+        TS3ReaderConfigPtr s3ReaderConfig,
+        TChunkId chunkId)
+    {
+        // TODO(discuss in PR): this is an extremely questionable approach, which we'll need to
+        // discuss and fix. The problem appeared in tests, but I see it may appear in a real
+        // environment as well: if this code is reached a short time after medium creation, this
+        // new medium will not appear in the directory until it's synced, leading to an error.
+        // As a better approach we can invoke the NextSync after an unsuccessful attempt to get
+        // a medium by index, for instance.
```
cherepasshka: I think to sync medium directory after unsuccessful attempt is better than to do sync uncnditional

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
```diff
@@ -0,0 +1,245 @@
...
+            S3Poller_->GetInvoker());
+
+        return CreateS3RegularChunkReader(
+            std::move(s3Client),
+            mediumDescriptor,
+            s3ReaderConfig,
+            chunkId);
+    }
+
+    template <typename RequestsType>
```
cherepasshka: We start typenames in templates with T as well (**T**RequestsType)

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
```diff
@@ -0,0 +1,245 @@
...
+        THashMap<TChunkId, IChunkReaderPtr> readers;
+        for (const auto& request: requests) {
+            auto chunkId = FromProto<TChunkId>(request.chunk_id());
+            if (readers.contains(chunkId)) {
+                continue;
+            }
+
+            readers[chunkId] = CreateS3ReaderForRequest(
+                request,
+                s3ReaderConfig,
```
cherepasshka: To move

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+        return readers;
+    }
+
+    template <class RequestType>
+    TFuture<std::vector<TErrorOr<TRefCountedChunkMetaPtr>>> GetChunkMetasForRequests(
+        const RepeatedPtrField<RequestType>& requests,
+        THashMap<TChunkId, IChunkReaderPtr>& s3Readers)
+    {
+        std::vector<TFuture<TRefCountedChunkMetaPtr>> chunkMetaFutures;
```
cherepasshka, line 153: I think could be useful to reserve space

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
```diff
@@ -0,0 +1,245 @@
...
+        for (const auto& request : requests) {
+            auto chunkId = FromProto<TChunkId>(request.chunk_id());
+            chunkMetaFutures.push_back(s3Readers[chunkId]->GetMeta({}));
+        }
+        return AllSet(chunkMetaFutures);
```
cherepasshka: To move

yt/yt/tests/cpp/test_s3_data/test_s3_data.cpp
```diff
@@ -0,0 +1,290 @@
...
+using namespace NYTree;
+using namespace NNodeTrackerClient;
+using namespace NYson;
+
+using ::testing::HasSubstr;
+using ::testing::SizeIs;
+
+////////////////////////////////////////////////////////////////////////////////
+
+// const auto Logger = CppTestsLogger;
```
cherepasshka: Not used

yt/yt/tests/cpp/test_s3_data/test_s3_data.cpp
```diff
@@ -0,0 +1,290 @@
...
+    std::vector<TBlock> GeneratedBlocks_;
+
+    NS3::IClientPtr S3Client_;
+
+    IChunkReaderPtr S3ChunkReader_;
+    IChunkWriterPtr S3ChunkWriter_;
+    TChunkId ChunkId_;
+    TS3MediumDescriptorPtr MediumDescriptor_;
+
+    NNative::IClientPtr NativeClient_ = DynamicPointerCast<NNative::IClient>(Client_);
```
cherepasshka: I think Client_ is not initialized with something reasonable at this point

yt/yt/ytlib/chunk_client/replication_reader.cpp
```diff
@@ -93,6 +93,78 @@ static const double MaxBackoffMultiplier = 1000.0;
 
 ////////////////////////////////////////////////////////////////////////////////
 
+struct TPeerId
```
cherepasshka: Better to make it a class, since in has some non-trivial methods and logic behind

yt/yt_proto/yt/client/chunk_client/proto/chunk_spec.proto
```diff
@@ -8,6 +8,13 @@ option go_package = "go.ytsaurus.tech/yt/go/proto/client/chunk_client";
 
 ////////////////////////////////////////////////////////////////////////////////
 
+// Describes the replica of the chunk.
+message TChunkReplicaSpec
+{
+    // Encoded TChunkReplicaWithMedium.
+    required fixed64 value = 1;
```
cherepasshka: Maybe to rename it into encoded_chunk_replica_with_medium?

yt/yt/ytlib/chunk_client/replication_reader.cpp
```diff
@@ -93,6 +93,78 @@ static const double MaxBackoffMultiplier = 1000.0;
...
+{
+    TPeerId() = default;
+
+    explicit TPeerId(const std::string& address)
```
cherepasshka: Maybe to make only one constructor with 2 parameters: address and mediumIndex and to set default value `GenericMediumIndex` or the second?

yt/yt/ytlib/chunk_client/replication_reader.cpp
```diff
@@ -93,6 +93,78 @@ static const double MaxBackoffMultiplier = 1000.0;
...
+    {
+        YT_VERIFY(!IsOffshore() || MediumIndex != GenericMediumIndex);
+    }
+
+    bool IsOffshore() const
+    {
+        return IsAddressOffshore(Address);
+    }
+
+    bool operator == (const TPeerId& other) const noexcept
```
cherepasshka: `bool operator == (const TPeerId& other) const noexcept = default`

yt/yt/ytlib/chunk_client/replication_reader.cpp
```diff
@@ -915,13 +1013,23 @@ class TSessionBase
...
+                // Do not mix domestic and offshore peers in the result of this function.
+                if (!topPeerId.IsOffshore()) {
+                    // If the peer isn't offshore, we can just add it.
+                    candidates.push_back(top.Peer);
+                } else if (candidates.empty()) {
+                    // If the peer is offshore and it's the first peer we encounter, just return it.
+                    candidates.push_back(top.Peer);
+                    PeerQueue_.pop();
+                    return {candidates, false};
+                }
```
cherepasshka, line 1046: Maybe to alert or just log that we received request with offshore and domestic peers?

yt/yt/ytlib/chunk_client/replication_reader.cpp
```diff
@@ -1194,6 +1302,11 @@ class TSessionBase
             return {candidates.begin(), candidates.end()};
         }
 
+        // Offshore node does not participate in probing and selecting process with others.
+        YT_VERIFY(std::all_of(candidates.begin(), candidates.end(), [] (const auto& peer) {
```
cherepasshka: I think it will be better to alert offshore nodes and filter them out than to crash service

yt/yt/ytlib/chunk_client/replication_reader.h
```diff
@@ -23,6 +23,17 @@ IChunkReaderAllowingRepairPtr CreateReplicationReader(
     TChunkId chunkId,
     TChunkReplicaList seedReplicas);
 
+// TODO(discuss in PR): we need the last argument to be of type TChunkReplicaWithMedium,
+// as we need the medium index to read offshore data. 8 months ago there was a commit which stops
```
cherepasshka, line 27: Could you please attach link to this commit?

yt/yt/server/offshore_data_gateway/bootstrap.cpp
achulkov2: I didn't regret this decision in QA components (they all log config updates), it is useful to see what changed and when. Let's keep it up to the point where it gets too big to be readable (so far it is ~empty).

yt/yt/ytlib/chunk_client/replication_reader.h
pavel-bash: This is the commit: https://github.com/ytsaurus/ytsaurus/commit/b474d77. `TChunkReplicaWithMediumList` was switched to `TChunkReplicaList` in many places including the replication reader.

yt/yt/ytlib/chunk_client/replication_reader.cpp
```diff
@@ -93,6 +93,78 @@ static const double MaxBackoffMultiplier = 1000.0;
...
+    }
+
+    TPeerId(const std::string& address, int mediumIndex)
+        : Address(address)
+        , MediumIndex(mediumIndex)
+    {
+        YT_VERIFY(!IsOffshore() || MediumIndex != GenericMediumIndex);
+    }
+
+    bool IsOffshore() const
```
achulkov2: Can we make an attempt at actually getting rid of this? I.e. make offshoreness a boolean-ish flag and set it when constructing PeerId, and make our new node directory return these peer ids instead of addresses.
And/or we can do inheritance to retrieve channel based on peer id.

Might be too cumbersome to implement, but we should try. Right now `TPeerId` isolates the "kostyl", but it is still there.

yt/yt/client/node_tracker_client/node_directory.cpp
pavel-bash: Well, the only actually used function across those 4 is the `FindPotentiallyOffshoreNodeDescriptor` with `TNodeId` argument. The other 3 I have added for consistency with the `NodeDirectory` and potential future uses. Do you think it's a bad idea to keep them?

yt/python/yt/environment/yt_env.py
pavel-bash: Added TODO-s.

yt/yt/server/lib/nbd/random_access_file_reader.cpp
pavel-bash: You're right, it should have been the `TChunkReplicaList{}`. The explicit creation is required now as otherwise it leads to a compilation error - it needs to know which overload to use.

yt/yt/tests/cpp/test_s3_data/test_s3_data.cpp
pavel-bash: It seems like it actually is initialized: GTest runs `SetUpTestCase` functions before it constructs the test objects themselves, and the initialization of `Client_` is inside the `TApiTestBase::SetUpTestCase`. However, I will put the initialization of `NativeClient_` into `SetUpReplicationReader()` to avoid confusion.

yt/yt/ytlib/chunk_client/replication_reader.cpp
pavel-bash: If I understand correctly, you're talking about the situation when a chunk has both domestic and offshore replicas? I think it's totally acceptable and even expected in future - I can see users who may want to have replicas on cluster and on their S3 storage, for example.

The main idea for us here is not to mix together domestic and offshore replicas when we perform the actual read request, as otherwise we could send a domestic replica to offshore data gateway and vice versa. But if one replica is unavailable, it should be supported to read from another.

yt/yt/ytlib/chunk_client/replication_reader.cpp
pavel-bash: Hm, I think I've added these lines by mistake - it's acceptable for offshore peers to participate in this process, there's even a stub for probing an offshore peer. I'll remove the verify here.

yt/yt/ytlib/chunk_client/replication_reader.cpp
pavel-bash: I have removed the kostyl, now there's no `OffshoreNodeAddress`. However, I feel like I went a bit backwards with this - now replication reader has more code around the offshore-ness of peers; for instance, in `PrepareNextPass` we again do not retrieve the offshore node from the directory, but branch right away. Though I agree it's probably less kostyl than it was before. Let me know what you think of the new approach.

pavel-bash:
@cherepasshka fixed your comments, replied to a couple of them as well.

yt/yt/server/offshore_data_gateway/bootstrap.cpp
AleksandraZh, line 183: а почему это fatal? как будто подписаться можно просто

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
AleksandraZh, line 54: if it is a subset, we can probably make a separate TBlaBlaProxy and use it both in data node and here (maybe inherit TDataNodeServiceProxy from TBlaBlaProxy)

AleksandraZh: auto

yt/yt/ytlib/chunk_client/replication_reader.cpp
AleksandraZh, line 101: looks like variant
or at least maybe we can do 2 constructors (one with nodeId and address and another one with medium index)

AleksandraZh: надо просто "реплики" банить, а не ноды
но вообще мб это все надо уносить отсюда

yt/yt/server/job_proxy/job_proxy.cpp
AleksandraZh: memory

AleksandraZh, line 962: а зачем это в джоб проксе появилось? Выглядит немного как смерть если честно

yt/yt/server/lib/job_proxy/config.h
AleksandraZh, line 563: а какую проблему мы решаем? Это надо обсудить, мне кажется так нельзя

yt/yt/server/master/cell_master/world_initializer.cpp
AleksandraZh, line 530: конфиг обычно лежит атрибутом, а не документом рядом
то есть 
//sys/offshore_data_gateways
и
//sys/offshore_data_gateways/@config

yt/yt/server/offshore_data_gateway/bootstrap.cpp
AleksandraZh, line 278: мне не очень нравится while true, но вроде такое уже есть, можно думаю потом отдельно что-нибудь сделать

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
AleksandraZh: maybe just CreateS3Reader

AleksandraZh, line 84: why "generic"?

AleksandraZh: Emplace/InsertOrCrash

AleksandraZh, line 187: maybe s3 reader should fill those?

yt/yt/ytlib/chunk_client/replication_reader.cpp
AleksandraZh, line 196: so medium is now present in both TPeerId and TChunkReplicaWithMedium? Why?

AleksandraZh: maybe banning offshore replica forever in a replication reader is ok, they don't live that long

AleksandraZh, line 1037: why?

AleksandraZh: this should be a TODO or smth

AleksandraZh: why are we looking at GetMediumIndex from suggestor here?

AleksandraZh: I think there should be a clearer indication that this doesn't work for offshore for now

AleksandraZh, line 999: there should be an explaination on why do we need that bool at all

yt/yt/server/offshore_data_gateway/bootstrap.cpp
pavel-bash: Well, it's the same for other bootstaps, so I've done it here as well (`cluster_node/bootstrap.cpp`, `query_tracker/bootstrap.cpp`, ...).

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
pavel-bash: There's a very specific reason why we've done it like that, implementing a subset of methods instead of having a separate service/proxy or inheriting from a common one. Take a look at `replication_reader.cpp`, the `DoRequestMeta`, for instance.

There we are able to instantiate the instance of the proxy just as `TDataNodeServiceProxy proxy(channel);`, and this channel may be a channel to a DataNode or to ODG. That's extremely convenient and clean, in my opinion. If the code is extended, and ODG will be invoked with non-implemented method of DataNode, it will just return the corresponding error, and we will know we need to also implement it there.

On the other hand, if we use two different services, we would need to either instantiate both proxies in that code or at least have knowledge of the underlying type (if we decide to instantiate the ancestor). Thus, I really believe the current approach is cleaner and easier.

yt/yt/ytlib/chunk_client/replication_reader.cpp
pavel-bash: I have tried an approach with two constructors which led to such a code:
```
    //! This constructor must be invoked for non-offshore peers; they require a
    //! valid address.
    TPeerId(TNodeId nodeId, std::string address)
        : IsOffshore(nodeId == OffshoreNodeId)
        , MediumIndex(GenericMediumIndex)
        , Address(std::move(address))
    {
        YT_VERIFY(!IsOffshore);
        YT_VERIFY(!Address->empty());
    }

    //! This constructor must be invoked for offshore peers; they require a
    //! valid medium index.
    explicit TPeerId(int mediumIndex)
        : IsOffshore(true)
        , MediumIndex(mediumIndex)
    {
        YT_VERIFY(MediumIndex != GenericMediumIndex);
    }
```

Even though it may look prettier than a single ctor, it makes the surrounding code a little bit messier. For instance, I as a caller must always know which peer I'm creating (domestic or offshore) to select the correct ctor. However, it's not always immediately obvious.

Consider the `AddPeer` function. Currently there we construct a `TPeerId` right away without thinking which type of peer it is, and it makes the whole code around type-agnostic. If we were to proceed with 2 ctors, we would need to compare the `replica.GetNodeId()` and select the correct ctor, which is much less clean in my opinion. Same would happen in other parts of the replication reader.

That's why at the end I liked the single ctor approach more.

yt/yt/server/job_proxy/job_proxy.cpp
pavel-bash: I agree it's a kostyl, but we found no other way to handle this scenario (for now). If we do not perform the synchronization right away before the job starts execution, we risk that the medium directory will not be updated by the time the code reading from it is executed.

If we leave this parameter as `false`, such a situation happens on our clusters 75% of the time: a job starts, requires to read something offshore, goes to ODG, ODG performs `MediumDirectory_->GetByIndexOrThrow` to get the S3 medium and miserably fails even though such medium exists, but the directory wasn't synchronized yet.

As Andrei suggested, the non-kostyl way to fix it is probably to pass the medium directory to job proxy in job spec (written in TODO in `job_proxy/config.h`), but I'm not sure we want to do right now or even if we want to do it.

___

I wrote this and understood that in this upstream version I've added the `NextSync` call into `CreateS3Reader` function in `offshore_data_gateway_service.cpp` to solve the problem with mediums which were just created. However, it also solves this problem as well. So we can consider to remove the config parameter and everything around it.

yt/yt/server/lib/job_proxy/config.h
pavel-bash: Responded to you in https://github.com/ytsaurus/ytsaurus/pull/1553#discussion_r2781807729, we can discuss there.

yt/yt/server/offshore_data_gateway/bootstrap.cpp
pavel-bash: Yeah, it happens in most other bootstraps as well.

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
pavel-bash: Github did not mark which lines you're referencing here, but I assume it's:
```
auto genericMediumDescriptor = MediumDirectory_->FindByIndex(replicaWithMedium.GetMediumIndex());
```

I called it generic as it's not yet establishes as S3 medium which we need and cast to later. I can call this one `mediumDescriptor` and the S3 one a `mediumDescriptorS3`, but I don't see much difference.

pavel-bash: Github did not mark which lines you're referencing here, and I can't guess in this case.

yt/yt/ytlib/chunk_client/replication_reader.cpp
pavel-bash: Well, the medium index which is present in `TPeerId` is part of the unique identifier of the replica on a given peer (together with node ID and address), while `TChunkReplicaWithMedium` is part of the peer's definition and contains more than just the medium index.

So, I'd say they play different roles and should remain separated to make the intentions clear.

pavel-bash: Github did not mark which lines you're referencing here, and I can't guess in this case.

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
pavel-bash: For reference: this comment is about lines starting with:
```
response->set_has_complete_chunk(true);
```

I don't think S3 reader should fill those. The `response` is a very specific RPC object `TRspGetBlockSet`, and I don't think it's a good idea to mix or somehow pass it to the more generic S3 reader which operates with blocks.

yt/yt/ytlib/chunk_client/replication_reader.cpp
pavel-bash: For reference, the line commented is:
```
// Do not mix domestic and offshore peers in the result of this function.
```

Multiple peers as a result of this function are used to create hedging channels. It's not possible currently to create a hedging channel between a domestic and an offshore replicas, I think we discussed it during the meeting a couple of weeks ago. It was also suggested that the hedging can be improved later to include such a possibility, but not at the moment.

pavel-bash:
Closing this PR in favor to https://github.com/ytsaurus/ytsaurus/pull/1688.

PR #1554, opened 2025-12-24
libevent: avoid redefining arc4random_buf on glibc

gearonixx:
Tested on Ubuntu 18.04 and 20.04: build OK.

On Ubuntu 24.04 (glibc 2.39) and newer Ubuntu releases: the build fails because arc4random_buf is defined both by glibc and libevent/arc4random.c. This patch guards the local arc4random_buf() so it is only compiled on glibc < 2.36

### DIFF
--- a/contrib/libs/libevent/arc4random.c
+++ b/contrib/libs/libevent/arc4random.c
@@ -484,6 +484,7 @@ arc4random(void)
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


ilyaibraev:
@gearonixx, hello!

Thanks for the suggested solution. It could potentially cause problems on other systems.

Is this problem urgent? My colleagues are trying to come up with a more correct, but time-consuming, solution.

ilyaibraev:
Fixed in aafa9df21e59c7832948312ff3b26463f43e5c9b.

Issue #1527, opened 2025-12-01
BETWEEN operator in PG syntax requires an exact match of argument types.


Tony-Romanov:
This query: 
```
--!syntax_pg

select * from dbgen_version
where "date" between '2025-11-20'::date and '2025-11-10'::date + '10 days'::interval;
--This filter works fine:
--where "date" >= '2025-11-20'::date and  "date" <= '2025-11-10'::date + '10 days'::interval;

-- The same proplem with integer literals.
with x as (select count(*) cnt from dbgen_version)
select * from x
where cnt between 0 and 6;
```
contains errors:
- Mismatch of type of between elements: date and timestamp
- Mismatch of type of between elements: int8 and int4

But in original postgres this "between" works fine.

<img width="1602" height="1433" alt="Image" src="https://github.com/user-attachments/assets/bffc39af-8da2-4567-a03a-cd08e6755d6b" />

PR #1557, opened 2025-12-30
Fix type annotation of PgBetween callable.

Tony-Romanov:
The fix of [issue](https://github.com/ytsaurus/ytsaurus/issues/1527).

Type: fix
Component: query-tracker


yql/essentials/core/type_ann/type_ann_pg.cpp
```diff
@@ -6068,45 +6068,55 @@ IGraphTransformer::TStatus PgBetweenWrapper(const TExprNode::TPtr& input, TExprN
...
-        bool isUniversal;
-        if (!ExtractPgType(type, argType, convertToPg, input->Child(i)->Pos(), ctx.Expr, isUniversal)) {
+    for (ui32 i = 0; i < children.size(); ++i) {
+        bool convertToPg = false;
+        bool isUniversal = false;
+        if (!ExtractPgType(children[i]->GetTypeAnn(), argTypes[i], convertToPg, input->Child(i)->Pos(), ctx.Expr, isUniversal)) {
             return IGraphTransformer::TStatus::Error;
         }
 
-        if (isUniversal) {
```
vitstn, line 1964: lost code

anyway, I can't see any test for this change
and have no idea if any breaking changes were introduced (should be tied to langver in this case + a pragma for check for this incompatibility should be developed)

Tony-Romanov: Was lost on update and merge. Fixed.

About test: all YQL+YT test has deleted from github.
I asked @nepal about reason and why on our last meeting.

nepal (approved):

psushin:
We have discussed inside - to merge this PR we would like to see at least some tests. Frankly speaking we would like to ensure that behaviour of the BETWEEN is fully compatible with the behaviour of actual PostrgeSQL, in terms of supported data types.

Tony-Romanov:
> We have discussed inside - to merge this PR we would like to see at least some tests. Frankly speaking we would like to ensure that behaviour of the BETWEEN is fully compatible with the behaviour of actual PostrgeSQL, in terms of supported data types.

I'd gladly add more tests if the YQL tests were available on GitHub.
That's exactly why I asked for them to be published.

When?

I'd also like to add that because of this bug, I had to further modify ~70% of the original queries from TPC-DS in the PG syntax.

For now I manually checked the PG compliance.

PR #1562, opened 2026-01-09
Fix TRetryingPeriodicExecutor logic without period or with backoff limit

koct9i:
- fix GetBackoffInterval() - CachedBackoffMultiplier were wrong
- handle retries for out of band invocations for executor without period
- exit backoff mode at reaching invocation count limit
- set DefaultInvocationCount to int max for backward compatibility

This way retrying periodic executor without period could be used for
exponential backoff retrying with limited retry count. Retrying stop
at first success or reaching limit.

If period is set then after leaving backoff mode execution returns back
to normal periodic execution. Because this never happened before default
InvocationCount is changed to int max.

---

* Changelog entry
Type: fix
Component: misc-server

Fix logic of exponential backoff in retrying periodic executor.

### DIFF
--- a/yt/yt/core/concurrency/unittests/periodic_ut.cpp
+++ b/yt/yt/core/concurrency/unittests/periodic_ut.cpp
@@ -6,12 +6,15 @@
 #include <yt/yt/core/concurrency/action_queue.h>
 #include <yt/yt/core/concurrency/delayed_executor.h>
 #include <yt/yt/core/concurrency/periodic_executor.h>
+#include <yt/yt/core/concurrency/retrying_periodic_executor.h>
 #include <yt/yt/core/concurrency/scheduler.h>
 #include <yt/yt/core/concurrency/thread_pool.h>
 
 #include <yt/yt/core/misc/lazy_ptr.h>
 
 #include <yt/yt/core/logging/log.h>
+
+#include <util/generic/xrange.h>
 
 #include <exception>
 #include <atomic>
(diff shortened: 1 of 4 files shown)


yt/yt/core/concurrency/unittests/periodic_ut.cpp
```diff
@@ -370,6 +373,52 @@ TEST_W(TPeriodicTest, Stop)
...
+            .MinBackoff = TDuration::MilliSeconds(10),
+            .MaxBackoff = TDuration::MilliSeconds(1000),
+            .BackoffMultiplier = 2,
+            .BackoffJitter = 0,
+        });
+
+    executor->Start();
+    auto prev = TInstant::Now();
+    executor->ScheduleOutOfBand();
+    for (auto i: xrange(4)) {
```
k-pogorelov: Missing space before ":"

yt/yt/core/concurrency/unittests/periodic_ut.cpp
```diff
@@ -370,6 +373,52 @@ TEST_W(TPeriodicTest, Stop)
     EXPECT_TRUE(startFuture.Get().IsOK());
 }
```
k-pogorelov, line 375: Could you make a test with making period 0?

k-pogorelov (changes requested):

yt/yt/core/concurrency/unittests/periodic_ut.cpp
koct9i: I don't understand.

period is `optional<TDuration>`
period 0 means call in loop without any delay (and switch to backoff after failure)

k-pogorelov: I meant std::nullopt

Issue #1563, opened 2026-01-12
Authorization token from request may be logged in response when using python sdk


Krock21:
How to reproduce the issue:

Run this python script for any operation:
```
#!/usr/bin/env python3
import sys

from yt.wrapper.job_commands import list_jobs

op_id = sys.argv[1]
for _ in range(101):
    try:
        list_jobs(op_id)
    except Exception:
        pass
```

While it is running, after some successful calls, make so that an error is returned (if http proxies are behind balancer, then just killing proxies should be enough)
See token in logs:
```
Message: 'Perform HTTP %s request %s (%s)'
Arguments: ('GET', 'http://cluster/api/v4/list_jobs', 'headers: {\'User-Agent\': \'Python wrapper 0.13.45\', \'Accept-Encoding\': \'gzip, identity\', \'X-Started-By\': b\'{"pid"=62467;"user"="krock21";}\', \'X-YT-Header-Format\': \'<format=text>yson\', \'X-YT-Parameters\': b\'{"suppress_transaction_coordinator_sync"=%false;"operation_id"="2d9a6b7-188ee0d4-510e03e8-d706788e";"output_format"=<"ignore_inner_attributes"=%false;"always_create_attributes"=%false;"format"="binary";"lazy"=%false;"sort_keys"=%false;>"yson";}\', \'X-YT-Accept-Framing\': \'1\', \'X-YT-Correlation-Id\': \'7f336c80-77df1d4-b16bb87f-5a04ebd1\'}, request_id: 4783e03b')
2026-01-09 12:00:52,684	ERROR	Bad response:
	StatusCode: 503
	Date: Fri, 09 Jan 2026 11:00:54 GMT
	Content-Type: text/html
	Content-Length: 190
	Connection: keep-alive
	X-Request-ID: 086ab75ecc8b2911864174efe05ad587
	Authorization: OAuth ytct-1234-12345                                -------         HERE

	<html>
	<head><title>503 Service Temporarily Unavailable</title></head>
	<body>
	<center><h1>503 Service Temporarily Unavailable</h1></center>
	<hr><center>nginx</center>
	</body>
	</html>
[...]

PR #1564, opened 2026-01-12
Remove misleading code from python sdk that adds Authorization headers to response, which may be logged unobfuscated

Krock21:
Remove misleading code from python sdk that adds Authorization headers to response, which may be logged unobfuscated

Issue: https://github.com/ytsaurus/ytsaurus/issues/1563

---

* Changelog entry
Type: fix
Component: python-sdk

Fixed Authorization headers appearing in response when it is logged

### DIFF
--- a/yt/python/yt/wrapper/http_driver.py
+++ b/yt/python/yt/wrapper/http_driver.py
@@ -160,13 +160,8 @@ class TokenAuth(AuthBase):
         if self.token is not None:
             request.headers["Authorization"] = "OAuth " + self.token
 
-    def handle_redirect(self, request, **kwargs):
-        self.set_token(request)
-        return request
-
     def __call__(self, request):
         self.set_token(request)
-        request.register_hook("response", self.handle_redirect)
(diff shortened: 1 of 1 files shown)


yt/python/yt/wrapper/http_driver.py
```diff
@@ -160,13 +160,8 @@ def set_token(self, request):
         if self.token is not None:
             request.headers["Authorization"] = "OAuth " + self.token
 
-    def handle_redirect(self, request, **kwargs):
```
Krock21, line 163: Don't be confused by naming here. `handle_redirect` doesn't mean it. It means `handle_every_response`

`request` doesn't mean it. It means `response`

denvr: This legacy feature used for authorisation on redirect (`requests` remover auth headers on redirect)
But now it works w/o it (requests make call `auth` for each request (even redirects))

denvr:
LGTM

PR #1571, opened 2026-01-19
Improve memory consumption for Top/TopSort comp nodes

UgnineSirdis:
Improve memory consumption for the cases when the result is much less than max count parameter.
This improvement fixes the bug: https://github.com/ydb-platform/ydb/issues/31905


Artemonchik:
👍

Issue #1406, opened 2025-08-15
RPC Proxy Multilookup behaviour when one of the tables is unmounted


mtvch:
As for now, when performing multilookup and one of the tables does not exist, I receive following error:

`call [UUID] failed: error getting mount info for [TABLE_PATH]: error getting attributes of table [TABLE_PATH]: error communicating with master: error resolving path [TABLE_PATH]/@: node [DIR_PATH] has no child with key “[TABLE_NAME]”
`

When table is unmounted I receive a similar error that tells that it `cannot read from tablet [UID] of table [TABLE_PATH] while it is in "unmounted" state`

In my use case, I need to receive data from as many tables as possible, and if one of them is unmounted, it's ok if I receive no data from it. I think that there needs to be an option in multilookup method that enables such behaviour. If you approve, I can try to create a PR for that.

mtvch:
@savnadya Hi! Any updates here? Can I implement it or are you planning to implement it or do you reject this feature?

PR #1577, opened 2026-01-28
Added allow_failure and subrequest_timeout options for multilookups

mtvch:
Feature for https://github.com/ytsaurus/ytsaurus/issues/1406

Hello! I’ve added new `allow_failure` and `subrequest_timeout` options for `multilookup` requests in the RPC proxy.

Both options are needed for cases when there is something wrong with one of the tables in subrequest and you want to still get results from other tables.

`allow_failure` option allows you to receive data even if some of the tables in the subrequests are unmounted or do not exist.

`subrequest_timeout` option helps you handle cases when one of the tables started responding slowly and you want to get data from other tables.

Without these options if you query 10 tables and even one of them is unmounted or missing or is slow to respond, the entire request fails.
* Changelog entry
Type: feature
Component: proxy

Added new `allow_failure` and `subrequest_timeout`  options for `multilookup` requests.

### DIFF
--- a/yt/yt/client/api/dynamic_table_client.h
+++ b/yt/yt/client/api/dynamic_table_client.h
@@ -6,6 +6,8 @@
 
 #include <yt/yt/client/query_client/query_statistics.h>
 
+#include <yt/yt/core/misc/error.h>
+
 namespace NYT::NApi {
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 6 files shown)


mtvch:
I removed changes in Go SDK from this PR and added tests using cpp client

mtvch:
I added `subrequest_timeout` option to this PR, I think it makes sense to review them together

PR #1580, opened 2026-02-02
yt/yt/core/crypto: use built-in CA when config is not specified

koct9i:
Just consolidation of logic in one place.
No changes in behavior for known users.


koct9i:
part of #1536
which most likely will be implemented on lover levels inside openssl
worth to apply regardless of approach for handling SSL_CERT_FILE

PR #1582, opened 2026-02-03
Fix use-after-free in TError during program shutdown

perst20:
---

Type: fix
Component: library/cpp/yt/error

Problem: TError::Enricher_ and TError::FromExceptionEnricher_ are static class members whose destructors are registered when error.cpp is loaded. However, if any code calls Singleton<>() during static initialization before error.cpp loads, the OnExit handler gets registered in standard atexit() first. Due to LIFO ordering, at program exit the enrichers are destroyed before OnExit() runs, but OnExit() then destroys Singletons whose destructors may create TError objects (e.g., to cancel futures), which invokes Enrich() on the already-destroyed std::function, causing use-after-free. This can lead to intermittent segfaults depending on the link order of translation units.

Solution: Store enrichers in a LeakySingleton<TEnricherStorage> so they are never destroyed, as TError can be created anywhere including during program shutdown.

### DIFF
--- a/library/cpp/yt/error/error.cpp
+++ b/library/cpp/yt/error/error.cpp
@@ -244,19 +244,17 @@ std::vector<TError>& ApplyWhitelist(std::vector<TError>& errors, const THashSet<
 
 ////////////////////////////////////////////////////////////////////////////////
 
-TError::TErrorOr() = default;
+void TError::TImplDeleter::operator()(TImpl* impl) const
+{
+    delete impl;
+}
+
+////////////////////////////////////////////////////////////////////////////////
 
 TError::~TErrorOr() = default;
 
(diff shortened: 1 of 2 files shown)


perst20:
@maxim-babenko hello, could you please take a look?

maxim-babenko:
Ship it!

library/cpp/yt/error/error.cpp
maxim-babenko: This brace must go to the next line

maxim-babenko: Make is struct and drop `public`

maxim-babenko (changes requested):
Some minor cosmetics, otherwise LGTM

maxim-babenko:
Shipped!

PR #1583, opened 2026-02-04
Support master --compatibility-info without passing the config

temich-nosov:
Previously the `--compatibility-info` flag required passing the master config:
```
root@b87863f0a363:/# /bin/ytserver-master --compatibility-info
(NYT::TErrorException) Error parsing config file 
    origin          (unknown) (pid 18, thread ytserver-master, fid 0)
    datetime        2026-02-02T18:11:24.149029Z
  (Error 2: No such file or directory) util/system/file.cpp:936: can't open "" with mode RdOnly|Seq (0x00000028)
      origin          (unknown) (pid 18, thread ytserver-master, fid 0)
      datetime        2026-02-02T18:11:24.149258Z
      exception_type  TFileError
```
That is not needed to retrieve the reign.

Moved flag handling to `TProgram` class to exit before config loading.

By default `--compatibility-info` outputs a plain text now, if `--yson` flag is not passed:
```
$ ./ytserver-all ytserver-master --compatibility-info
Current Reign: 3204
$ ./ytserver-all ytserver-master --compatibility-info --yson
{
    "current_reign" = 3204;
}
```

---
* Changelog entry
Type: fix
Component: master

Allow --compatibility-info to work without providing a config file.

### DIFF
--- a/yt/yt/library/program/program.cpp
+++ b/yt/yt/library/program/program.cpp
@@ -88,6 +88,9 @@ TProgram::TProgram()
     Opts_.AddLongOption("build", "Prints build information")
         .NoArgument()
         .StoreValue(&PrintBuild_, true);
+    Opts_.AddLongOption("compatibility-info", "Prints compatibility info (e.g. current reign)")
+        .NoArgument()
+        .StoreValue(&PrintCompatibilityInfo_, true);
     Opts_.SetFreeArgsNum(0);
 }
 
@@ -98,7 +101,7 @@ void TProgram::SetCrashOnError()
     CrashOnError_ = true;
 }
 
-void TProgram::HandleVersionAndBuild()
+void TProgram::HandleProgramInfo()
 {
     if (PrintVersion_) {
         PrintVersionAndExit();
@@ -109,6 +112,15 @@ void TProgram::HandleVersionAndBuild()
     if (PrintBuild_) {
         PrintBuildAndExit();
     }
+    if (PrintCompatibilityInfo_) {
+        DoPrintCompatibilityInfo();
+        Exit(0);
+    }
+}
+
+void TProgram::DoPrintCompatibilityInfo()
+{
+    THROW_ERROR_EXCEPTION("Compatibility info is not implemented for this program");
 }
 
 int TProgram::Run(int argc, const char** argv)
@@ -119,7 +131,7 @@ int TProgram::Run(int argc, const char** argv)
     OptsParseResult_ = std::make_unique<TOptsParseResult>(this, argc, argv);
 
     auto run = [&] {
-        HandleVersionAndBuild();
+        HandleProgramInfo();
         DoRun();
     };
 
--- a/yt/yt/library/program/program.h
+++ b/yt/yt/library/program/program.h
@@ -24,8 +24,8 @@ public:
     [[noreturn]]
     int Run(int argc, const char** argv);
 
-    //! Handles --version/--yt-version/--build [--yson] if they are present.
-    void HandleVersionAndBuild();
+    //! Handles --version/--yt-version/--build/--compatibility-info [--yson] if they are present.
+    void HandleProgramInfo();
 
     //! Nongracefully aborts the program.
     /*!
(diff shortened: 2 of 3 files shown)


yt/yt/library/program/program.cpp
```diff
@@ -111,6 +114,16 @@ void TProgram::HandleVersionAndBuild()
...
 }
 
+void TProgram::HandleCompatibilityInfo() {
+    if (PrintCompatibilityInfo_) {
+        PrintCompatibilityInfoAndExit();
+    }
+}
+
+void TProgram::PrintCompatibilityInfoAndExit() {
+    THROW_ERROR_EXCEPTION("PrintCompatibilityInfoAndExit is not implemented");
```
achulkov2: Let's write a less code-related error message, e.g. "Compatibility info is not implemented for this program".

yt/yt/library/program/program.h
```diff
@@ -27,6 +27,9 @@ class TProgram
     //! Handles --version/--yt-version/--build [--yson] if they are present.
     void HandleVersionAndBuild();
 
+    //! Handles --compatibility-info if it is present.
+    void HandleCompatibilityInfo();
```
achulkov2: Why not add it to the method above? It is already handling different options, seems like it just needs a more generic name and it can handle all of our options.

yt/yt/server/master/cell_master/program.cpp
```diff
@@ -177,19 +170,13 @@ class TCellMasterProgram
...
     void ValidateOpts() final
     {
         if (static_cast<int>(IsDumpSnapshotMode()) +
             static_cast<int>(IsValidateSnapshotMode()) +
-            static_cast<int>(IsExportSnapshotMode()) +
-            static_cast<int>(IsPrintCompatibilityInfoMode()) > 1)
+            static_cast<int>(IsExportSnapshotMode()) > 1)
         {
-            THROW_ERROR_EXCEPTION("Options 'dump-snapshot', 'validate-snapshot', 'export-snapshot', 'compatibility-info' are mutually exclusive");
+            THROW_ERROR_EXCEPTION("Options 'dump-snapshot', 'validate-snapshot', 'export-snapshot' are mutually exclusive");
```
achulkov2, line 179: Let's double check what happens if we pass it along with these options with which it is exclusive?
Probably we will just print compatibility info and ignore everything else, but not sure.

yt/yt/server/master/cell_master/program.cpp
```diff
@@ -296,15 +278,21 @@ class TCellMasterProgram
...
+        if (UseYson_) {
+            NYson::TYsonWriter writer(&Cout, NYson::EYsonFormat::Pretty);
+            auto info = NYTree::BuildYsonStringFluently()
+                .BeginMap()
+                    .Item("current_reign").Value(NCellMaster::GetCurrentReign())
+                .EndMap();
+            NYson::Serialize(info, &writer);
+            Cout << Endl;
+        } else {
+            Cout << "Current reign: " << NCellMaster::GetCurrentReign() << Endl;
```
achulkov2: I think the style for the other options is `Current Reign`.

yt/yt/server/master/cell_master/program.cpp
```diff
@@ -296,15 +278,21 @@ class TCellMasterProgram
         SleepForever();
     }
 
-    void DoPrintCompatibilityInfo()
+    void PrintCompatibilityInfoAndExit()
```
achulkov2: Let's only override the `Print` part and not the `Exit` part, since it is all the same everywhere?

Exit can be part of the base method implementation.

yt/yt/library/program/program.cpp
```diff
@@ -111,6 +114,16 @@ void TProgram::HandleVersionAndBuild()
     }
 }
 
+void TProgram::HandleCompatibilityInfo() {
```
achulkov2: This is our [styleguide](https://github.com/ytsaurus/ytsaurus/blob/main/yt/styleguide/cpp.md) for C++. There are some more kinks to it, but it is a good start. 

For example, `{` in functions go on the next line. Let's fix this and other places according to the styleguide.

achulkov2 (changes requested):

yt/yt/server/master/cell_master/program.cpp
temich-nosov: Yes, that exactly what will happen, I've checked. This validation won't happen when --compatibility-info flag is passed.

achulkov2 (approved):
LGTM, we can move it out of draft.

raundn (approved):
LGTM

PR #1584, opened 2026-02-04
[YT Flow] Fix serializing TInstant in states

TheLegionCrazy:
* Changelog entry
Type: fix
Component: yt/yt/flow/lib/serializer

YTree serializes TInstant as Iso8601 (string) but TInstant columns are generated in appropriate type that does not support string values. 
1. Write TInstant column schema as `timestamp` (to maintain accuracy)
2. Convert timestamp columns to microseconds during serialization


yt/yt/flow/lib/serializer/unittests/serializer_ut.cpp
```diff
@@ -77,6 +77,7 @@ struct TTestSchemaYsonStruct
     NProto::TTestMessage YsonProto;
     std::tuple<double, std::string> Tuple;
     std::optional<NYson::TYsonString> Yson;
+    TInstant Instant;
```
Mikari, line 80: I suggest to write similar test for TDuration.

yt/yt/flow/lib/serializer/serializer.cpp
```diff
@@ -66,7 +66,11 @@ class TStructSerializer final
     void OnStringScalar(TStringBuf value) override
     {
         if (StructLevel() && FieldId_) {
-            AddValue(MakeUnversionedStringValue(value, *FieldId_));
+            if (ESimpleLogicalValueType::Timestamp == Schema_->Columns().at(*FieldId_).CastToV1Type()) {
+                AddValue(MakeUnversionedUint64Value(TInstant::ParseIso8601(value).MicroSeconds(), *FieldId_));
```
Mikari, line 70: Looks like possible performance degradation: TInstant => string => ui64.

Mikari:
I suppose, this partially solves the problem https://github.com/ytsaurus/ytsaurus/commit/97fed55cca9765467ad2863d3aace16170d986a5
New types - NYT::NYTree::TPreciseInstant and TPreciseDuration always serialized as ui64 without any precision loss.

TheLegionCrazy:
> I suppose, this partially solves the problem [97fed55](https://github.com/ytsaurus/ytsaurus/commit/97fed55cca9765467ad2863d3aace16170d986a5) New types - NYT::NYTree::TPreciseInstant and TPreciseDuration always serialized as ui64 without any precision loss.

yup.

Issue #1586, opened 2026-02-05
Go SDK: wire encoder produces text YSON for columns of type any, causing replication failures


mtvch:
## Steps to Reproduce
1. Create a replicated table with column of type `{"type"="any";"type_v3"={"type_name"="optional";"item"="yson";}}`
2. Insert values into replicated table using `ytrpc` client:
```go
encoded, err := yson.MarshalFormat("[1;2];", yson.FormatBinary)
row["val"] = yson.RawValue(encoded)
ytClient.InsertRows(ctx, path, []any{row}, opts)
```
3. Write to replicated table returns success, but replication error happens:
`Cannot decode non-binary YSON string in \"any\"-typed column`
https://github.com/ytsaurus/ytsaurus/blob/ff8057d73b41af567aa89d2cd8d5fa749ad318ed/yt/yt/client/table_client/helpers.cpp#L1535

## More context
With http client simpler serialization works without errors:
```go
row["val"] = "[1;2];"
ytClient.InsertRows(ctx, path, []any{row}, opts)
```
But if you replace http with rpc client here, write to replicated table will return error: `invalid type: expected "any", actual "string"`. That's why I added more sofisticated serialization logic.

## Possible solution
Replace `yson.Marshal(value.Interface())`
https://github.com/ytsaurus/ytsaurus/blob/ff8057d73b41af567aa89d2cd8d5fa749ad318ed/yt/go/wire/encode.go#L334
with `yson.MarshalFormat(value.Interface(), yson.FormatBinary)`

But I'm still not sure if it's correct behaviour that replicated table accepts write and can't replicate it to it's replica. I would expect replicated table to return immediately in such scenario.

PR #1587, opened 2026-02-05
Go SDK: fix wire encoding yson text -> binary

mtvch:
Possible fix for https://github.com/ytsaurus/ytsaurus/issues/1586


tinarsky:
Hello, I'll check it within a week

PR #1591, opened 2026-02-08
yt/docs: add missing operation options to documentation

koct9i:
Two operation options defined in `yt/yt/ytlib/scheduler/config.cpp` were missing from user documentation.

* Added `omit_inaccessible_rows` (default: false) - Skips inaccessible rows instead of failing at startup
* Added `suspend_on_job_failure` (default: false) - Suspends operation on job failure for investigation instead of failing after reaching `max_failed_job_count`

Both options documented in English and Russian operation options reference.

Co-authored-by: Konstantin Khlebnikov <khlebnikov@nebius.com>


coteeq:
Shipped!

PR #1605, opened 2026-02-20
Move row_level_security into library/query to avoid llvm targeting

MrLolthe1st:
Made yt/yt/ytlib link without LLVM dependencies


yt/yt/library/query/row_level_security_api/row_level_security.h
```diff
@@ -58,14 +58,6 @@ void FormatValue(TStringBuilderBase* builder, const TRlsReadSpec& rlsReadSpec, T
 
 ////////////////////////////////////////////////////////////////////////////////
 
-struct IRlsCheckerFactory
-    : public TRefCounted
```
Krisha11, line 62: Why was it moved to another place?

yt/yt/library/query/row_level_security_api/row_level_security.h
```diff
@@ -1,13 +1,13 @@
 #pragma once
 
-#include "public.h"
+#include <yt/yt/client/table_client/public.h>
```
Krisha11: `yt/yt/library` should be before `yt/yt/core`

yt/yt/server/controller_agent/controllers/table.h
```diff
@@ -13,7 +13,7 @@
 #include <yt/yt/ytlib/cypress_client/public.h>
 
 #include <yt/yt/ytlib/table_client/config.h>
-#include <yt/yt/ytlib/table_client/row_level_security.h>
+#include <yt/yt/library/query/row_level_security_api/row_level_security.h>
```
Krisha11: `yt/yt/library` should be after `yt/yt/client`

yt/yt/server/controller_agent/helpers.h
```diff
@@ -12,7 +12,7 @@
 #include <yt/yt/ytlib/scheduler/config.h>
 #include <yt/yt/ytlib/scheduler/helpers.h>
 
-#include <yt/yt/ytlib/table_client/row_level_security.h>
+#include <yt/yt/library/query/row_level_security_api/row_level_security.h>
```
Krisha11: same

yt/yt/ytlib/chunk_client/data_source.h
```diff
@@ -7,7 +7,7 @@
 #include <yt/yt/ytlib/scheduler/cluster_name.h>
 #include <yt/yt/ytlib/scheduler/input_query.h>
 
-#include <yt/yt/ytlib/table_client/row_level_security.h>
+#include <yt/yt/library/query/row_level_security_api/row_level_security.h>
```
Krisha11: same

yt/yt/ytlib/table_client/schemaless_multi_chunk_reader.cpp
```diff
@@ -10,14 +10,15 @@
...
 #include "remote_dynamic_store_reader.h"
-#include "row_level_security.h"
 #include "schemaless_block_reader.h"
 #include "schemaless_multi_chunk_reader.h"
 #include "table_read_spec.h"
 #include "versioned_chunk_reader.h"
 #include "versioned_reader_adapter.h"
 #include "performance_counters.h"
 
+#include <yt/yt/library/query/row_level_security_api/row_level_security.h>
```
Krisha11: same

yt/yt/library/query/row_level_security_api/row_level_security.h
MrLolthe1st: Because yt/yt/ytlib peerdirs to yt/yt/library/query/row_level_security_api
ya.make doesn't work with circular dependencies

coteeq:
Let's do it a bit more elegant:
There is a `TCGInstanceHolder` in `row_level_security.cpp`. It would be much better if we only offloaded its creation and interface and left everything else as is.

Also, I don't think that a separate dir in `library/query` is necessary. We can just place `row_level_security.h` in `library/query/engine_api` (same for `_api` counterparts).

yt/yt/library/query/engine_api/row_level_security.cpp
```diff
@@ -0,0 +1,9 @@
+#include "row_level_security.h"
+
+namespace NYT::NQueryClient {
+
+Y_WEAK std::pair<ICGInstanceHolderPtr, NTableClient::TTableSchemaPtr> MakeRlsCGInstance(const NTableClient::TTableSchemaPtr& /*schema*/, const std::string& /*predicate*/) {
```
Krisha11: { next line please

MrLolthe1st:
> There is a TCGInstanceHolder in row_level_security.cpp. It would be much better if we only offloaded its creation and interface and left everything else as is.

I'm tried to move as less parts as possible, but: TCGInstanceHolder now implements ICGInstanceHolder interface, and it's creation and implementation moved to query/engine. Moreover, there was an query_engine_config, that lived inside query/engine, and client_dynamic_tables_impl.cpp used it. I'm moved it into query/engine_api, along with part of cg_cache (it was splitted into cg_cache_config and cg_cache).

Other parts remains untouched.

yt/yt/library/query/engine_api/row_level_security.h
```diff
@@ -0,0 +1,18 @@
...
+        NTableClient::TUnversionedValue* value,
+        NTableClient::TUnversionedRow schemafulRow,
+        const NTableClient::TRowBufferPtr& rowBuffer) = 0;
+};
+
+using ICGInstanceHolderPtr = TIntrusivePtr<ICGInstanceHolder>;
+
+std::pair<ICGInstanceHolderPtr, NTableClient::TTableSchemaPtr> MakeRlsCGInstance(const NTableClient::TTableSchemaPtr& schema, const std::string& predicate);
+
+}
```
coteeq, line 18: `// namespace NYT::NQueryClient`

yt/yt/library/query/engine_api/row_level_security.h
```diff
@@ -0,0 +1,18 @@
+#include <yt/yt/client/table_client/name_table.h>
+#include <yt/yt/client/table_client/row_buffer.h>
+#include <yt/yt/client/table_client/schema.h>
+
+namespace NYT::NQueryClient {
+
+struct ICGInstanceHolder : public virtual TRefCounted {
```
coteeq, line 7:
```
struct ICGInstanceHolder:
    public virtual TRefCounted
{
```

yt/yt/library/query/engine_api/row_level_security.h
```diff
@@ -0,0 +1,18 @@
...
+    virtual void Run(
+        NTableClient::TUnversionedValue* value,
+        NTableClient::TUnversionedRow schemafulRow,
+        const NTableClient::TRowBufferPtr& rowBuffer) = 0;
+};
+
+using ICGInstanceHolderPtr = TIntrusivePtr<ICGInstanceHolder>;
```
coteeq, line 14: Use `DECLARE_REFCOUNTED_TYPE`

yt/yt/library/query/engine_api/row_level_security.cpp
coteeq, line 5: `{` on new line

yt/yt/library/query/unittests/evaluate/test_evaluate.cpp
```diff
@@ -2,8 +2,7 @@
 
 #include <yt/yt/library/query/base/query_helpers.h>
 
-#include <yt/yt/library/query/engine/query_engine_config.h>
-
+#include <yt/yt/library/query/engine_api/query_engine_config.h>
```
coteeq, line 5: Sort includes please

yt/yt/server/controller_agent/ya.make
```diff
@@ -36,6 +36,7 @@ SRCS(
 PEERDIR(
     library/cpp/yt/phdr_cache
 
+    yt/yt/library/query/engine
```
coteeq, line 39: Sort peerdirs please

yt/yt/ytlib/api/native/client_dynamic_tables_impl.cpp
```diff
@@ -106,7 +106,7 @@
 #include <yt/yt/library/query/base/functions.h>
 #include <yt/yt/library/query/base/query_preparer.h>
 
-#include <yt/yt/library/query/engine/query_engine_config.h>
+#include <yt/yt/library/query/engine_api/query_engine_config.h>
```
coteeq, line 109: It should be grouped (and sorted) with other `engine_api` includes.

yt/yt/library/query/engine_api/row_level_security.h
```diff
@@ -0,0 +1,18 @@
...
+
+std::pair<ICGInstanceHolderPtr, NTableClient::TTableSchemaPtr> MakeRlsCGInstance(const NTableClient::TTableSchemaPtr& schema, const std::string& predicate);
```
coteeq, line 16: Let's adjust schema in ytlib so this function would only create the CG instance
Something like this is in my head:
```
ICGInstanceHolderPtr CreateCGInstance(
    const TConstExpressionPtr& expression,
    const TTableSchemaPtr& schema);
```

yt/yt/library/query/engine/row_level_security.cpp
```diff
@@ -0,0 +1,81 @@
...
+            value,
+            schemafulRow.Elements(),
+            rowBuffer);
+    }
+};
+
+////////////////////////////////////////////////////////////////////////////////
+
+}
+
```
coteeq, line 39: Too many newlines

yt/yt/library/query/engine/row_level_security.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include <yt/yt/library/query/engine/folding_profiler.h>
+
+#include <yt/yt/library/query/base/query.h>
+#include <yt/yt/library/query/base/query_preparer.h>
+
+namespace NYT::NQueryClient {
+
+using namespace NTableClient;
+namespace {
```
coteeq, line 11: Why anon namespace?

yt/yt/library/query/engine/row_level_security.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+
+struct TCGInstanceHolder final : public ICGInstanceHolder
```
coteeq, line 15: Same here - inherit on another line

yt/yt/library/query/engine_api/query_engine_config.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include <yt/yt/core/misc/configurable_singleton_def.h>
+
+namespace NYT::NQueryClient {
+
+using namespace NYson;
+using namespace NYTree;
+
+////////////////////////////////////////////////////////////////////////////////
+
+#ifdef YT_VERBOSE_CHANGING_QUERY_ENGINE_CONFIG
```
coteeq, line 14: This belongs to `engine/query_engine_config.cpp`

(UPD: Or here if we move reconfiguration stuff here)

yt/yt/library/query/engine/public.h
coteeq, line 1: This file is essentially empty. Let's delete it.

yt/yt/library/query/engine/query_engine_config.cpp
coteeq, line 48: I don't think you can DEFINE_RECONFIGURABLE_SINGLETON multiple times. Let's move singleton stuff to engine_api

yt/yt/library/query/engine_api/row_level_security.h
coteeq, line 7: I didn't notice it before, but we actually do not need an abstract interface here. Let's make this a concrete type instead, so we do not to pay for dynamic dispatch on every row.

yt/yt/library/query/engine_api/cg_cache_config.cpp
```diff
@@ -0,0 +1,18 @@
+#include "cg_cache_config.h"
+
+#include <library/cpp/yt/memory/leaky_ref_counted_singleton.h>
```
coteeq, line 3: Is it needed?

yt/yt/library/query/engine_api/cg_cache_config.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
+
+#include "public.h"
+
+#include <yt/yt/library/query/engine_api/evaluation_helpers.h>
```
coteeq, line 5: This and `configurable_singleton_decl` are not used

coteeq:
> I'm moved it into query/engine_api, along with part of cg_cache

It's time to change the title of PR then :)

coteeq:
I chatted a bit with colleagues from query engine's team and we think that we can remove the dependency ourselves. We can borrow the implementation from orm, which, I believe, is not exported to github.

Thank you for your patch anyway. It showed me that removing this dependency is not (at least shouldn't be) as hard as I imagined.

yt/yt/library/query/engine/query_engine_config.cpp
MrLolthe1st: DEFINE_RECONFIGURABLE_SINGLETON isn't needed here. I'll just remove it
weak functions in dummy implementation will be ignored

MrLolthe1st:
Discussed with YT team, that they will remove dependency themselves.

PR #1612, opened 2026-02-24
[query-tracker/0.1] Use get_config_from_env for YtClient in init scripts

koct9i:
- **init_operations_archive: use get_config_from_env**
- **init_queue_agent_state: use get_config_from_env**
- **init_query_tracker_state: use get_config_from_env**

---

Init scripts should respect configuration files and environment variables.
For example this is required for passing CA-bundle for HTTPS-only cluster.

There is simple way to fetch all options from environment and config files.
Unfortunately this is not default behaviour and contains some boilerplate.

update_config_from_env() configs applying order:
- config argument
- apply env YT_CONFIG_PATCHES
- apply config file YT_CONFIG_PATH (~/.yt/config)
- apply env YT_*

YtClient configs applying order:
- get_default_config()
- apply config argument
- apply proxy and token argument
- apply remote patch

---

Pull Request resolved: https://github.com/ytsaurus/ytsaurus/pull/1516

Co-authored-by: denvr <denvr@ytsaurus.tech>
commit_hash:e45f5f78cbd8cec8157a1a76f9b9dfbedf6aadf2

(cherry picked from commit 61fdad70be24b74e915e269546fa6d8d0a61462d)

### DIFF
--- a/yt/python/yt/environment/init_operations_archive.py
+++ b/yt/python/yt/environment/init_operations_archive.py
@@ -1,7 +1,8 @@
 #!/usr/bin/python3
 
 import yt.yson as yson
-from yt.wrapper import YtClient, config, ypath_split
+from yt.wrapper import YtClient, ypath_split
+from yt.wrapper.default_config import get_config_from_env
 
 from yt.environment.init_cluster import get_default_resource_limits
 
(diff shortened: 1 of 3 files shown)


dim-an:
Shipped!

PR #1619, opened 2026-02-26
Remove deprecated ClientMapTimeout

Tony-Romanov:
From log:
`WARN  ytserver-yql-agent(pid=71096, tid=0x00007F173A7306C0) [default] yql_dispatch.cpp:186: Pragma "ClientMapTimeout" is deprecated and has no effect	Yql:4`
Type: fix
Component: query-tracker

https://github.com/ytsaurus/ytsaurus/blob/main/yt/yql/providers/yt/common/yql_yt_settings.cpp#L152


antoniox:
Shipped!

PR #1627, opened 2026-03-03
Add repository conditions to prevent scheduled/push/PR workflows from running in forks

koct9i:
Scheduled, push, and PR workflows that require secrets fail silently or
error in forks where those secrets are unavailable.

Adds if: github.repository == 'ytsaurus/ytsaurus' guards to the affected jobs,
consistent with the existing pattern in cache-heater.yaml.

Co-authored-by: Konstantin Khlebnikov <khlebnikov@nebius.com>

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -8,6 +8,7 @@ on:
 jobs:
   check:
     name: Run checks
+    if: github.repository == 'ytsaurus/ytsaurus'
     runs-on: ubuntu-latest
     strategy:
       matrix:
@@ -58,7 +59,7 @@ jobs:
   send-alert:
     needs:
       - check
-    if: always()
+    if: always() && github.repository == 'ytsaurus/ytsaurus'
     uses: ./.github/workflows/check-status-and-send-alert.yaml
     with:
       is_failure: ${{ contains(join(needs.*.result, ','), 'failure') }}
--- a/.github/workflows/chyt_release.yaml
+++ b/.github/workflows/chyt_release.yaml
@@ -23,6 +23,7 @@ env:
 jobs:
   start-vm:
     name: Start VM
+    if: github.repository == 'ytsaurus/ytsaurus'
     runs-on: trampoline
     outputs:
       label: ${{ steps.start-yc-runner.outputs.label }}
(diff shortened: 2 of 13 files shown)


epsilond1 (approved):

ilyaibraev:
Why restrict WF from running in forks? And, as far as I remember, unless you explicitly allow WF to run in a new repository, they won't run. Moreover, some WF runs on GitHub's computing power and doesn't require any secrets (except for secrets for sending crash alerts, but those aren't that important).

koct9i:
> Why restrict WF from running in forks? And, as far as I remember, unless you explicitly allow WF to run in a new repository, they won't run. Moreover, some WF runs on GitHub's computing power and doesn't require any secrets (except for secrets for sending crash alerts, but those aren't that important).

None of these workflows are working in forks. They constantly spam errors.
The only option is patching all of them or disabling all workflows completely.

ilyaibraev:
> None of these workflows are working in forks. They constantly spam errors. The only option is patching all of them or disabling all workflows completely.

As far as I can see, Go&Java CI works: https://github.com/koct9i/ytsaurus/actions/runs/22633191740 https://github.com/koct9i/ytsaurus/actions/runs/21804143155

What if we disable only release and system WFs?

.github/workflows/go.yaml
```diff
@@ -8,6 +8,7 @@ on:
 jobs:
   check:
     name: Run checks
+    if: github.repository == 'ytsaurus/ytsaurus'
```
ilyaibraev, line 11: What if we disable only release and system WFs?

ilyaibraev (changes requested):
What if we disable only release and system WFs?

.github/workflows/go.yaml
koct9i: What are you trying to achieve? If something runs and does not fails this does not makes it useful by default. I don't find any of these workflow useful and worth running automatically.
Do you have real scenarios when they might be useful for you?

Useful atomatic per-commit workflow must run in minutes not hours.
Some kind of fast linter/syntax cehck is ok. Running all tests is not ok.

Workflow which runs test must be either filtered by changed files and/or fillter mught be provieded as parameter at running manually.

PR #1630, opened 2026-03-05
Add schema generator for polymorphic yson struct

perst20:
---

* Added polymorphic yson struct schema generator
Type: feature
Component: yt/yt/core/ytree


Mikari:
1) Could you explain motivation, please.
2) Is it compatible with type_v3? I'm not sure tagged<yson> is allowed by type_v3. If I remember correctly, yson always should be optional.
3) 'polymorphic' field looks like extension to type_v3. Is it really required?

PR #1637, opened 2026-03-11
Add new odin checks: cypress_commands, read_static_table_commands, write_static_table_commands

Krock21:
Currently, there is no good alternatives for these checks:
1. Cypress commands
2. Static table read
3. Static table write

Other checks can be polluted by other issues on the cluster (compute, connectivity, ...), which makes it hard to calculate SLA for the most basic things

This PR adds checks above

These checks are deployed to our clusters already

### DIFF
--- a/yt/odin/bin/yt_odin_generate_config/__main__.py
+++ b/yt/odin/bin/yt_odin_generate_config/__main__.py
@@ -331,13 +331,11 @@ def create_odin_checks_config():
                 },
             },
             "read_static_table_commands": {
-                "check_timeout": 60,
                 "options": {
                     "table_path": "//sys/admin/odin/read_static_table_commands/table",
                 },
             },
             "write_static_table_commands": {
-                "check_timeout": 60,
                 "options": {
                     "temp_tables_path": "//sys/admin/odin/write_static_table_commands",
                 },


yt/odin/checks/bin/read_static_table_commands/__main__.py
```diff
@@ -0,0 +1,51 @@
...
+    _ensure_table_exists(yt_client, logger, table_path)
+
+    logger.info("Checking existence of %s", table_path)
+    if not yt_client.exists(table_path):
+        message = "Static table {} does not exist after creation attempt".format(table_path)
+        logger.error(message)
+        return states.UNAVAILABLE_STATE, message
+
+    logger.info("Reading %s", table_path)
+    rows = _ensure_seed_row_present(yt_client, logger, table_path)
```
savnadya, line 40: `read_static_table_commands` and `write_static_table_commands` look almost the same. Wouldn't it be better to make a single check?

Krock21: We see that some users care about cluster being available for reading, while others care about cluster being available for writing
Merging these checks will mix these expectations together and make the signal too mixed and unusable by both types of users

I think splitting dynamic table commands check to read/write one may be useful as well, to represent real expectation from real users

savnadya:
Shipped!

PR #1640, opened 2026-03-13
Fix pending disk space accounting in job resource manager

Kontakter:
* Changelog entry
Type: fix
Component: map-reduce

Fix pending disk space accounting in job resource manager

### DIFF
--- a/yt/yt/server/node/job_agent/job_resource_manager.cpp
+++ b/yt/yt/server/node/job_agent/job_resource_manager.cpp
@@ -625,6 +625,28 @@ public:
             releasingResources);
     }
 
+    void OnResourceDemandUpdated(const TLogger& Logger, const TJobResources& oldDemand, const TJobResources& newDemand)
+    {
+        YT_ASSERT_THREAD_AFFINITY(JobThread);
+
+        auto delta = newDemand - oldDemand;
+
+        TJobResources pendingResources;
+        {
+            auto guard = WriterGuard(ResourcesLock_);
+            pendingResources =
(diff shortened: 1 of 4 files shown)


Kontakter:
This code fully written by AI, my session: https://gist.github.com/Kontakter/38cf80d0f99206571ac7d57072510cd4

k-pogorelov:
Sorry for the slow review, and thanks for waiting.

yt/yt/server/node/job_agent/job_resource_manager.cpp
```diff
@@ -2056,18 +2078,26 @@ void TResourceHolder::UpdateResourceDemand(
...
+            allocationAttributes.PortCount);
+
+        oldDemand = BaseResourceUsage_;
+        BaseResourceUsage_ = resources;
+        AllocationAttributes_ = allocationAttributes;
+    }
 
-    BaseResourceUsage_ = resources;
-    AllocationAttributes_ = allocationAttributes;
+    ResourceManagerImpl_->OnResourceDemandUpdated(Logger, oldDemand, resources);
```
k-pogorelov, line 2100: I think this should be fixed by construction instead.

UpdateResourceDemand() is a compat path, and it is exactly what creates the inconsistency here: the holder is registered in Pending with one demand, and then its resource usage is changed before start.

Since allocation_attributes are already known at allocation creation time, I would rather pass them into the ResourceHolder constructor together with the final patched resource demand, and then drop UpdateResourceDemand() entirely. That should remove the bug at the root instead of fixing its consequences.

k-pogorelov (changes requested):

PR #1641, opened 2026-03-14
Truncate job proxy stderr on start

Kontakter:


yt/yt/server/node/exec_node/job_environment.cpp
```diff
@@ -164,6 +164,19 @@ class TProcessJobEnvironmentBase
                 });
             }
 
+            if (Bootstrap_->GetConfig()->ExecNode->JobProxy->JobProxyLogging->TruncateJobProxyStderrOnStart) {
```
k-pogorelov, line 167: Tell me, please, why is this necessary? This may complicate the debugging process.

PR #1653, opened 2026-03-19
Logging: cleanup old timestamp-suffixed segments on rotation checks

Kontakter:
* Changelog entry
Type: fix
Component: map-reduce

Logging: cleanup old timestamp-suffixed segments on rotation checks, it fixes removing old job proxy log files.

### DIFF
--- a/yt/yt/core/logging/file_log_writer.cpp
+++ b/yt/yt/core/logging/file_log_writer.cpp
@@ -66,6 +66,11 @@ public:
 
     void MaybeRotate() override
     {
+        if (FirstMaybeRotate_) {
+            FirstMaybeRotate_ = false;
+            RemoveOldSegments();
+        }
+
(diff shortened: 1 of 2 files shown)


yt/yt/core/logging/file_log_writer.cpp
```diff
@@ -66,6 +66,11 @@ class TFileLogWriter
 
     void MaybeRotate() override
     {
+        if (FirstMaybeRotate_) {
+            FirstMaybeRotate_ = false;
```
epsilond1, line 70: Maybe `IsFirstMaybeRotate`?

`FirstMaybeRotate` -- calls as managed parameter.

epsilond1:
In general, LGTM.

epsilond1 (changes requested):

PR #1658, opened 2026-03-24
Test on remote input.

Tony-Romanov:
Type: fix
Component: controller-agent


ilyaibraev:
Hello!

Please provide more context about the problem you're solving. You wrote "fix" but you only wrote a test.

Tony-Romanov:
I saw only two options for "Component" by default for "fix" & "feature", but test isn't fix or feature.

This is how YQL works with remote clusters: starts tx on a remote server then start merge operation for read from the remote transaction.

This scenario works fine on main but crashes the controller agent on our old  24.2.

In main it's just yet another test. 
YQL's scenario without any yql's components.
I didn't found the similar case in YT tests.

k-pogorelov:
Hello!
We have a test TestSchedulerRemoteOperationCommands::test_merge_does_not_teleport, that tests in particular the same invariant.

24.2 is not supported already, you can use fresher version.

Tony-Romanov:
Yes, we know about 24.2 and how fix our problem.
 
No, test_merge_does_not_teleport isn't the same.

In my test, the most critical point is:
**start_transaction**
and
**;transaction_id="{}">** 

But I didn't see any "start_transaction" in TestSchedulerRemoteOperationCommands.
Test without tx works fine on 24.2 too.

I can move the test into TestSchedulerRemoteOperationCommands if you say that.

coteeq:
I cherry-picked commit that contains the fix https://github.com/ytsaurus/ytsaurus/commit/7368708d848c6a4056e18b2824bbe0fe328cfeaa

For the record: there already is a test with parent transactions. It just did not make it to 24.2 (though it has now with the mentioned cherry-pick) https://github.com/ytsaurus/ytsaurus/blob/da21892e8cd27d1d5e9a4e989e4327432a1b37ec/yt/yt/tests/integration/controller/test_remote_operation.py#L624

PR #1661, opened 2026-03-25
Fix NULL handling crash in scan queries

UgnineSirdis:
There can be situations when we get NULL in nonoptional values. It will be better to stop the query, but not stop the server in this case.

One example of this situation is current implementation of BulkUpsert API in YDB: it allows to insert NULLs into a column that is marked as NOT NULL.

Related PR: https://github.com/ydb-platform/ydb/pull/36482


yql/essentials/minikql/computation/mkql_computation_node_pack.cpp
vitstn, line 660: .Get<> will handle nulls

Let it crash

vitstn (changes requested):

PR #1664, opened 2026-03-26
Fix postprocess call for polymorphic yson structs

TheLegionCrazy:
---

* Changelog entry
Type: fix
Component: yt/core/ytree

I have tried a lot of stuff but the duck-type approach is the only working way


TheLegionCrazy:
https://github.com/ytsaurus/ytsaurus/commit/98a5efc95e9c312df3b05fc7de0006c36c416da6

PR #1674, opened 2026-04-06
Add HLL tests and activate HLL as aggregate column

blackHATred:
Ytsaurus already has functionality for implementing hyperloglog, but it has not been utilized for creating aggregate columns and corresponding tests. This PR changes that by allowing hyperloglog to be used as an aggregate column, which will be useful for implementing unique counters

---

Type: feature
Component: dynamic tables / cypress master

### DIFF
--- a/yt/yt/client/table_client/schema.cpp
+++ b/yt/yt/client/table_client/schema.cpp
@@ -1979,9 +1979,6 @@ void ValidateColumnSchema(
             "dict_sum",
         };
         for (int precision = 7; precision <= 14; ++precision) {
-            result.insert(Format("hll_%v", precision));
-            result.insert(Format("hll_%v_state", precision));
-            result.insert(Format("hll_%v_merge", precision));
             result.insert(Format("hll_%v_merge_state", precision));
         }
         return result;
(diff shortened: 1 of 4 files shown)


torilov:
Hi,

Thank you for your contribution!

Unfortunately, we are not ready to merge it as-is.

The key issue we see with the current approach is that a user may need to:
    1. Access finalized values.
    2. Access raw state values for further aggregation via the query language.

The current design does not allow us to express both desired options.

We would also like to move away from using the string type in the schema, but we do not have a solid design for that yet.

Overall, we would prefer to design the API more carefully here to avoid breaking the interface in the future.
The implementation is good for a simplified scenario, but we are not ready to take it upstream in its current form.

blackHATred:
> The key issue we see with the current approach is that a user may need to:
> 1. Access finalized values.
> 2. Access raw state values for further aggregation via the query language.

Currently, both options are implemented. You can access the raw state and perform further operations on it (in this example, compute an estimate):
https://github.com/ytsaurus/ytsaurus/pull/1674/changes#diff-753f41bcf8994391025d34a464ff62a2d7154b3e848acb9366695ef411eba476R641

You can also use the existing `cardinality_merge` to obtain a finalized estimate via `SELECT`, without touching the raw state:
https://github.com/ytsaurus/ytsaurus/pull/1674/changes#diff-753f41bcf8994391025d34a464ff62a2d7154b3e848acb9366695ef411eba476R702

What changes are planned in YTsaurus that my implementation conflicts with? What should I modify to make it possible to introduce HyperLogLog at the aggregation column?

PR #1675, opened 2026-04-07
Fix aggregate columns for GO SDK for RPC client

teaDrunk4rd:
Fix "Aggregate columns does not work as expected for GO SDK for RPC client"
https://github.com/ytsaurus/ytsaurus/issues/1648

### DIFF
--- a/yt/go/yson/struct_tag.go
+++ b/yt/go/yson/struct_tag.go
@@ -13,6 +13,7 @@ type Tag struct {
 	Attr      bool
 	Attrs     bool
 	Key       bool
+	Aggregate bool
 }
 
 // ParseTag parses yson annotation for struct field tag.
@@ -48,6 +49,8 @@ func ParseTag(fieldName string, fieldTag reflect.StructTag) (tag *Tag, skip bool
 		case "attrs":
 			tag.Attrs = true
 			tag.Name = ""
+		case "aggregate":
+			tag.Aggregate = true
 		}
 	}
 
--- a/yt/go/wire/decode.go
+++ b/yt/go/wire/decode.go
@@ -27,6 +27,8 @@ type field struct {
 	attrs     bool
 
 	required bool // todo needed?
+
+	aggregate bool
 }
 
 type structType struct {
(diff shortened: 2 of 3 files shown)


tinarsky:
Hello, I'll review it within a week

yt/go/yson/struct_tag.go
```diff
@@ -48,6 +49,8 @@ func ParseTag(fieldName string, fieldTag reflect.StructTag) (tag *Tag, skip bool
 		case "attrs":
 			tag.Attrs = true
 			tag.Name = ""
+		case "aggregate":
```
tinarsky, line 52: This change wires yson:"...,aggregate" into wire.Value.Aggregate for struct rows, which matches the RPC rowset path. The repro in #1648 uses map[string]any plus InsertRowsOptions.Aggregate; that flow still does not set per-cell Aggregate on the wire. Can you confirm that fixing the struct + yson tag case is what you intended for this PR, and that leaving the map-only scenario as-is (or for a follow-up) is acceptable?

yt/go/yson/struct_tag.go
```diff
@@ -13,6 +13,7 @@ type Tag struct {
 	Attr      bool
 	Attrs     bool
 	Key       bool
+	Aggregate bool
```
tinarsky, line 16: Please add a small regression test that builds a struct with an aggregate yson tag, runs wire.Encode, and asserts the aggregate bit is set on the corresponding value (e.g. by unmarshaling the rowset or by checking whatever helpers encode_test.go already uses). Good place:

https://github.com/ytsaurus/ytsaurus/blob/main/yt/go/wire/encode_test.go

tinarsky (changes requested):

yt/go/yson/struct_tag.go
tinarsky: I'll fix it myself

yt/go/yson/struct_tag.go
tinarsky: I'll fix it myself

tinarsky (approved):

Issue #1678, opened 2026-04-11
Using child fields of pointer value without handling nil case


Babushkin05:
yt/go/yt/internal/rpcclient/encoder.go:736

```go
func (e *Encoder) CreateQueueProducerSession(
	ctx context.Context,
	producerPath ypath.Path,
	queuePath ypath.Path,
	sessionID string,
	opts *yt.CreateQueueProducerSessionOptions,
) (result *yt.CreateQueueProducerSessionResult, err error) {
	req := &rpc_proxy.TReqCreateQueueProducerSession{
		ProducerPath:    []byte(producerPath.String()),
		QueuePath:       []byte(queuePath.String()),
		SessionId:       ptr.String(sessionID),
		MutatingOptions: convertMutatingOptions(opts.MutatingOptions),
	}
	if opts.UserMeta != nil {
		req.UserMeta, err = yson.Marshal(opts.UserMeta)
		if err != nil {
			return
		}
	}
...
}
```

opts may be nil

dim-an:
Closed with linked PR

PR #1679, opened 2026-04-11
Fix nil-pointer panic in CreateQueueProducerSession when opts is not provided

Babushkin05:
This PR fixes a potential nil-pointer dereference in the Go SDK.

In Encoder.CreateQueueProducerSession we access opts.MutatingOptions, but opts is allowed to be nil by the public API. When the caller passes nil, the code panics.

Change:
- Initialize opts with an empty yt.CreateQueueProducerSessionOptions when opts == nil, so subsequent field access is safe.

Refs: #1678

* Changelog entry
Type: fix
Component: yt/go/yt/internal/rpcclient/encoder.go:736

Prevent nil-pointer panic in CreateQueueProducerSession by handling nil options (opts == nil).


savnadya:
Shipped!

PR #1688, opened 2026-04-15
Introduce the OffshoreDataGateway component

pavel-bash:
# Description

This PR adds the `OffshoreDataGateway` component which is used when reading data from S3. This component is also integrated into the replication reader to actually perform the reads.

One of the most important topics to note in this work is the introduction of the new proto message `TChunkReplicaSpec` and its usage in the `OffshoreDataGateway`'s RPC requests. This structure allows us to pass the medium index which is required to find the specific offshore medium where the chunk is located. Maybe there are other solutions, I'll be happy to discuss that.

Another topic is the approach to handling different media in the replication reader. Right now the replication reader uses only addresses to uniquely identify different peers to read from. This works because we read from data nodes only, and we do not care about the medium there. Now the reads may also happen from `OffshoreDataGateway`-s, and the medium starts to matter.

Imagine a scenario when a chunk has two offshore replicas on two different mediums, one pointing to Google S3 storage, another to AWS S3. We must have a way to differentiate those two replicas, even though the address is the same - the sentinel `OffshoreNodeAddress`. This is why I introduce a structure called `TPeerId` which includes both the address and the medium index, and now replication reader works with it to differentiate different replicas.

Last topic is the testing. It's impossible to implement an "honest" integration test at the moment as writes to offshore replicas are not implemented, and also masters know nothing of them. I have implemented a C++ unit test to check the behaviour of replication reader only - `test_s3_data.cpp`.


pavel-bash:
Previous version of this PR (contains some amount of discussion): https://github.com/ytsaurus/ytsaurus/pull/1553.

yt/python/yt/environment/default_config.py
```diff
@@ -953,3 +951,12 @@ def get_replicated_table_tracker_config():
             "update_period": 100,
         },
     }
+
+def get_offshore_data_gateway_config():
+    return yson.loads(b"""
+{
```
cherepasshka: we use json-formatted descriptions of config, better to change to:
{
        "dynamic_config_manager": {
            "update_period": 100,
        },
    }

yt/yt/server/job_proxy/job_proxy.cpp
```diff
@@ -48,6 +48,7 @@
 #include <yt/yt/ytlib/chunk_client/helpers.h>
 #include <yt/yt/ytlib/chunk_client/job_spec_extensions.h>
 #include <yt/yt/ytlib/chunk_client/traffic_meter.h>
+#include <yt/yt/ytlib/chunk_client/medium_directory_synchronizer.h>
```
cherepasshka: Is it used?

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+        return CreateS3RegularChunkReader(
+            std::move(s3Client),
+            mediumDescriptor,
+            std::move(s3ReaderConfig),
+            chunkId);
+    }
+
+    template <typename TRequestsType>
+    THashMap<TChunkId, IChunkReaderPtr> CreateS3Readers(
```
cherepasshka: Isn't used

yt/yt/ytlib/chunk_client/replication_reader.cpp
```diff
@@ -94,6 +94,96 @@ static const double MaxBackoffMultiplier = 1000.0;
...
+        // Non-offshore peer requires correct address.
+        YT_VERIFY(IsOffshore || address != std::nullopt);
+
+        // Offshore peer requires correct medium index.
+        YT_VERIFY(!IsOffshore || MediumIndex != GenericMediumIndex);
+    }
+
+    std::string GetPrintableAddress() const
+    {
+        // TODO(discuss in PR): I think this approach is cleaner than making Address non-optional
```
AleksandraZh: you can leave address optional and make 2 constructor for TPeerId -- one for offshore media and one for domestic

yt/yt/ytlib/chunk_client/replication_reader.cpp
```diff
@@ -94,6 +94,96 @@ static const double MaxBackoffMultiplier = 1000.0;
...
+        // and assigning this value - like that we keep the invariant that address is null if the
+        // peer is offshore. What do you
+        if (Address) {
+            return *Address;
+        }
+        return "generic-peer-address";
+    }
+
+    bool operator == (const TPeerId& other) const noexcept = default;
```
AleksandraZh: operator==

yt/yt/ytlib/chunk_client/replication_reader.cpp
```diff
@@ -94,6 +94,96 @@ static const double MaxBackoffMultiplier = 1000.0;
...
+
+    bool IsOffshore;
```
AleksandraZh, line 123: I'd rather it be a struct if it has private fields, but I do not insist

yt/yt/ytlib/chunk_client/replication_reader.cpp
AleksandraZh: and maybe make 2 helpers -- CreateOffshorePeerId and CreateDomesticPeerId

yt/yt/ytlib/chunk_client/replication_reader.cpp
```diff
@@ -896,23 +1001,31 @@ class TSessionBase
...
 
-    TPeerList PickPeerCandidates(
+    //! Picks the #count peers from the peer queue; returns a pair of the selected
+    //! peers and a boolean showing if more calls to this function may happen during
+    //! the current pass.
+    //!
+    //! NB: The function may return fewer than #count peers even when PeerQueue_ is not empty,
+    //! and return false to signal that it must not be called again in this round. This
+    //! prevents mixing domestic and offshore peers in the same selection.
+    std::pair<TPeerList, bool> PickPeerCandidates(
```
AleksandraZh: I am not a fan of returning a pair here, maybe we can pass a parameter indicating if we need offshore peers, domestic, or any

yt/yt/ytlib/chunk_client/replication_reader.cpp
AleksandraZh, line 2778: or you can just check here that we found offshore candidate and remove that bool from PickPeerCandidates

yt/yt/ytlib/chunk_client/replication_reader.cpp
```diff
@@ -4610,6 +4784,27 @@ IChunkReaderAllowingRepairPtr CreateReplicationReader(
...
     TChunkId chunkId,
     TChunkReplicaList seedReplicas)
+{
+    TChunkReplicaWithMediumList replicaList;
+    replicaList.reserve(seedReplicas.size());
+    for (auto& seedReplica: seedReplicas) {
+        replicaList.push_back(TChunkReplicaWithMedium(std::move(seedReplica)));
+    }
+
+    return New<TReplicationReader>(
```
AleksandraZh: CreateReplicationReader?

yt/yt/ytlib/offshore_data_gateway/offshore_data_gateway_channel.cpp
```diff
@@ -0,0 +1,178 @@
...
+                // This channel already detects failures internally.
+                auto channel = CreateBalancingChannel(
+                    balancingChannelConfig,
+                    ChannelFactory_,
+                    EndpointDescription_,
+                    EndpointAttributes_);
+
+                {
+                    auto guard = Guard(SpinLock_);
+                    CachedChannel_ = channel;
```
AleksandraZh, line 145: Do we ever use this? We recreate TOffshoreDataGatewayChannelProvider in CreateOffshoreDataGatewayChannel and CreateOffshoreDataGatewayChannel is the only thing that is visible outside

I'm not saying we should remove CachedChannel_, maybe let TOffshoreDataGatewayChannelProvider linger somewhere for longer

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+                const auto& meta = resultsError.Value();
+
+                *response->mutable_chunk_meta() = static_cast<NChunkClient::NProto::TChunkMeta>(*meta);
+
+                context->SetResponseInfo("MetaSize: %v", response->chunk_meta().ByteSize());
+
+                context->Reply();
+            })
+                .Via(StorageInvoker_));
```
AleksandraZh, line 203: formatting looks off

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+        auto reader = CreateS3Reader(*request, New<TS3ReaderConfig>(), chunkId);
+        reader->GetMeta({})
+            .Subscribe(BIND([=, this_ = MakeStrong(this)] (const TErrorOr<NChunkClient::TRefCountedChunkMetaPtr>& resultsError) {
+                if (!resultsError.IsOK()) {
+                    context->Reply(resultsError);
+                    return;
+                }
+
+                if (context->IsCanceled()) {
```
AleksandraZh: maybe check this first

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
```diff
@@ -0,0 +1,253 @@
...
+    DECLARE_RPC_SERVICE_METHOD(NChunkClient::NProto, GetBlockSet)
+    {
+        auto chunkId = FromProto<TChunkId>(request->chunk_id());
+        auto blockIndexes = FromProto<std::vector<int>>(request->block_indexes());
+
+        context->SetRequestInfo("ChunkId: %v, BlockIndexes: %v",
+            chunkId,
+            blockIndexes);
+
+        auto reader = CreateS3Reader(*request, New<TS3ReaderConfig>(), chunkId);
```
AleksandraZh, line 147: let's cache it

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
```diff
@@ -0,0 +1,253 @@
...
+        reader->ReadBlocks({}, blockIndexes)
+            .Subscribe(BIND([=, this_ = MakeStrong(this)] (const TErrorOr<std::vector<TBlock>>& resultsError) {
+                if (!resultsError.IsOK()) {
+                    context->Reply(resultsError);
+                    return;
+                }
+
+                if (context->IsCanceled()) {
```
AleksandraZh: maybe check this first

yt/yt/server/offshore_data_gateway/bootstrap.cpp
```diff
@@ -0,0 +1,321 @@
...
+                .Finish(),
+        };
+
+        auto registrar = CreateCypressRegistrar(
+            std::move(options),
+            New<TCypressRegistrarConfig>(),
+            NativeClient_,
+            GetCurrentInvoker());
+
+        while (true) {
```
AleksandraZh: why not just call Start?

yt/yt/ytlib/chunk_client/replication_reader.h
```diff
@@ -23,6 +23,17 @@ IChunkReaderAllowingRepairPtr CreateReplicationReader(
     TChunkId chunkId,
     TChunkReplicaList seedReplicas);
 
+// TODO(discuss in PR): we need the last argument to be of type TChunkReplicaWithMedium,
+// as we need the medium index to read offshore data. 8 months ago there was a commit which stops
+// passing the medium index into chunk readers, so we need to understand how to handle this. Maybe
+// (partially) revert the commit?
```
AleksandraZh: Yes, I think we are okay with restoring them (at least for offshore media)

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
```diff
@@ -0,0 +1,253 @@
...
+            OffshoreDataGatewayLogger(),
+            TServiceOptions{
+                .Authenticator = std::move(authenticator),
+            })
+        , StorageInvoker_(std::move(storageInvoker))
+        , S3Poller_(std::move(s3Poller))
+        , MediumDirectory_(std::move(mediumDirectory))
+        , MediumDirectorySynchronizer_(std::move(mediumDirectorySynchronizer))
+    {
+        RegisterMethod(RPC_SERVICE_METHOD_DESC(GetBlockSet));
```
AleksandraZh, line 66: how would writes work? I don't like the idea of asymmetry here

yt/yt/ytlib/chunk_client/replication_reader.cpp
pavel-bash: This was my initial approach. However, I've changed it to the single ctor when I encountered the lines in the `AddPeer` function:
```
TPeerId peerId{
    replica.GetNodeId(),
    replica.GetMediumIndex(),
    std::move(address),
};
```

Here we don't know whether the peer is domestic or offshore without explicitly checking for the `replica.GetNodeId()`, and also the `address` is optional here, so we'll have to unwrap it if the peer is domestic.

Why I liked the idea with a single ctor - does this job automatically by checking all arguments against the invariants and creating the correct type of the `TPeerId`. If you insist, I will change it to the two ctors approach, but I think currently it's more convenient to use.

yt/yt/ytlib/offshore_data_gateway/offshore_data_gateway_channel.cpp
pavel-bash: Well, it is actually used. In the `CreateOffshoreDataGatewayChannel` we use the `CreateRoamingChannel` to get the channel, wrap it into two other channel additions and return it.

The `TRoamingChannel` uses `auto asyncChannel = Provider_->GetChannel(request);` in `Send` (thus, for every request), and the `Provider_` there is the `TOffshoreDataGatewayChannelProvider` itself. Thus, the `CachedChannel_` is reused every time this channel is used.

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
pavel-bash: I've searched the code for the `.Via` usages, and I found both variants: in some cases it's on the same indentation as the `Subcribe` (like [here](https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/ytlib/chunk_client/replication_reader.cpp#L719)), and sometimes it's moved to the right (same file, [here](https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/ytlib/chunk_client/replication_reader.cpp#L2334)).

I don't know whether there's an agreement how we should format this, if you tell me to move it to the left, I'll do it.

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
pavel-bash: Cache the S3 reader between different calls to the OffshoreDataGateway? To be honest, I'm not sure it's a good idea:
1. It depends not only on the chunk ID, but also on the medium, and both are necessary to take into account. For example, the chunk may be reachable via one medium, but the other medium's authentication is not good to read this chunk anymore
2. Also, the authentication credentials may have changed in the medium since the last call, and the cached version of the reader would not be able to handle this

Of course those are the problems which can be solved, but I am not sure we gain that much from caching to even think about them...

yt/yt/ytlib/chunk_client/replication_reader.h
pavel-bash: So, what should we do? Shall I remove this comment and leave this restoration work to you? Or shall I do that in a separate PR?

yt/yt/server/offshore_data_gateway/offshore_data_gateway_service.cpp
pavel-bash: Discussed this during the meeting and in the SIG chat.

cherepasshka:
Shipped!

PR #1689, opened 2026-04-16
Add register queue consumer vital

fojnk:
---
Changelog entry
Type: fix
Component: go-sdk
Added support for register_queue_consumer_vital permission checks in the Go SDK.
CheckPermission now normalizes register_queue_consumer_vital to register_queue_consumer and sets vital=true when needed, which prevents error parsing EPermission value "register_queue_consumer_vital" and makes vital queue-consumer permission checks work correctly via both HTTP and RPC clients.


yt/go/yt/internal/params_gen.go
```diff
@@ -1695,6 +1695,10 @@ func writeCheckPermissionOptions(w *yson.Writer, o *yt.CheckPermissionOptions) {
 		w.MapKeyString("columns")
 		w.Any(o.Columns)
 	}
+	if o.Vital != nil {
```
tinarsky, line 1698: Was params_gen.go regenerated with yt-gen-client? If not, please regenerate it from yt/go/yt with go generate . (see the //go:generate line in interface.go), or run:
go run ./internal/yt-gen-client -interface interface.go -interface internal/interface.go -output internal/params_gen.go

tinarsky (changes requested):

tinarsky (approved):

PR #1696, opened 2026-04-24
Add empty target to bring contrib/python/boto3 to stable/25.3

pavel-bash:
# Description

Based on this [PR](https://github.com/ytsaurus/ytsaurus/pull/1461) which brought the `boto3` dependency into `main`. This dependency is currently a part of `stable/25.4`, we would really like it to appear on other `stable/25.X` branches.


savnadya:
Shipped!

PR #1708, opened 2026-05-01
[queues] fix queue/consumer snapshot ordering

TheLegionCrazy:
If a consumer is reading a queue faster than the PassPeriod, we can sometimes end up in a situation where the queue snapshot is older than the consumer snapshot.
It leads to incorrect consumer disposition and may report wrong partition lag row counts (e.g. <0, even though the consumer is not actually ahead).

I could not think of a deterministic test to reproduce this.
Also, I believe that querying each consumer partition snapshot with the corresponding queue partition barrier timestamp would end up DDOSing the consumer tablets, but using the lowest barrier is fine.

---

* Changelog entry
Type: fix
Component: queue-agent

Use lowest barrier timestamp of queue partitions when querying consumer partition snapshot.

### DIFF
--- a/yt/yt/benchmarks/queues/run/main.cpp
+++ b/yt/yt/benchmarks/queues/run/main.cpp
@@ -311,7 +311,11 @@ private:
     void ReportLag()
     {
         i64 nextRowIndex = 0;
-        auto partitionInfos = WaitFor(ConsumerClient_->CollectPartitions(std::vector<int>{0}, /*withLastConsumeTime*/ false))
+        auto partitionInfos = WaitFor(
+            ConsumerClient_->CollectPartitions(
+                std::vector<int>{0},
+                /*timestamp*/ std::nullopt,
+                /*withLastConsumeTime*/ false))
             .ValueOrThrow();
         if (!partitionInfos.empty()) {
             YT_VERIFY(partitionInfos.size() == 1);
--- a/yt/yt/client/queue_client/consumer_client.cpp
+++ b/yt/yt/client/queue_client/consumer_client.cpp
@@ -247,6 +247,7 @@ public:
 
     TFuture<std::vector<TPartitionInfo>> CollectPartitions(
         int expectedPartitionCount,
+        std::optional<TTimestamp> timestamp = std::nullopt,
         bool withLastConsumeTime = false) const override
     {
         if (expectedPartitionCount <= 0) {
@@ -277,6 +278,7 @@ public:
             &TGenericConsumerClient::DoCollectPartitions,
             MakeStrong(this),
             selectQuery,
+            timestamp,
             withLastConsumeTime)
             .AsyncVia(GetCurrentInvoker())
             .Run()
@@ -296,6 +298,7 @@ public:
 
     TFuture<std::vector<TPartitionInfo>> CollectPartitions(
         const std::vector<int>& partitionIndexes,
+        std::optional<TTimestamp> timestamp,
         bool withLastConsumeTime) const override
     {
         auto selectQuery = Format(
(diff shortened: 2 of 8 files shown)


ItIsApachee:
I'll do my best to review this PR within a couple of weeks.

TheLegionCrazy:
@ItIsApachee hi, any news?

ItIsApachee:
@TheLegionCrazy read this PR, sorry for the delay.
Seems fine for single cluster setups, however in general queue and consumer are not located on the same cluster, so relying on barrier timestamp from queue's cluster on consumer's cluster might not work.
I don't really know that much about dynamic table timestamps myself, so I'll ask around, but this problem should probably be solved in some different way.

ItIsApachee:
@TheLegionCrazy, this issue is not a priority to us right now, so I'd suggest you try to come up with alternative solutions, which I think should exist. We can discuss them in this PR.
In the meantime I'll try to research this topic myself a bit and see if there are workarounds and whether it works in a single cluster setup as I can't guarantee that yet as well.

TheLegionCrazy:
@ItIsApachee

> so I'd suggest you try to come up with alternative solutions

I don't see the motivation for exploring alternatives yet - what's the concern with the current approach besides multicluster setups ?

ItIsApachee:
> @ItIsApachee
> 
> > so I'd suggest you try to come up with alternative solutions
> 
> I don't see the motivation for exploring alternatives yet - what's the concern with the current approach besides multicluster setups ?

It would be best to come up with a universal solution. I don't like the idea of single cluster only logic.

ItIsApachee:
@TheLegionCrazy, got some info:
- monotonicity (as described in dynamic table transactions guarantees) is only held if the same cluster clock is used
- table writes to non-replicated dynamic tables of some cluster use the clock of this cluster
- for replicated tables metacluster's clock is used
- for chaos the clock is configured in the proxy configuration

With this in mind using barrier timestamp mostly works in a single cluster setup, but it must not be applied in cases timestamp monotonicity is not guaranteed.

ItIsApachee:
I might've come up with a way to solve this problem generally. I'll provide details in a bit.

ItIsApachee:
@TheLegionCrazy so consumers not only store the offset, but some meta info as well.
Meta is calculated here: https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/client/queue_client/consumer_client.cpp#L517
Meta calculation also fetches some info about the queue tablet using CollectPartitionRowInfos: https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/client/queue_client/helpers.cpp#L20
This calculation for example includes `$cumulative_data_weight` column. However, queue can also be created with a `$timestamp` column, which is used in queue agent for metrics calculation.
I'd bet that you can extend consumer meta to select `$timestamp` column as well (only in the case it exists). With that we would have a way to detect if consumer snapshot is ahead of the queue snapshot. The tricky part is then to wait for queue snapshot fresh enough to use for the full consumer snapshot calculation, but it should be possible.
Why is it tricky:
- no such mechanism yet exists
- we can't trust consumer meta column value fully in queue agent (the user can write anything to it if they so desire), especially since I propose waiting for indefinite amount of time to get a fresh enough queue snapshot, so the waiting part should be written carefully
- I'm not sure how to cover this logic with tests carefully, it's certainly possible, but not trivial

ItIsApachee:
It's actually not really the consistency that we achieve, since queue might be even more ahead of consumer (compared to the observed queue snapshot), but at the very least the `consumer_partition_offset <= queue_partition_upper_row_index` inequality holds, which is the thing we actually want.

ItIsApachee:
I'll convert this PR to draft for now, so it does not ping me regularly

TheLegionCrazy:
Thanks for diving. I'll do my best to evaluate the option you suggested next week.

Issue #1711, opened 2026-05-03
[go] wire: support custom row-level marshaler interface to bypass reflection


orloffv:
### Context

When writing high-throughput rows from a Go service to YTsaurus, the choice between `ythttp.NewClient` and `ytrpc.NewClient` shows up as a real CPU asymmetry on the client because of how each transport handles row encoding:

- `ythttp` calls `yson.Writer.Any(row)`, which honors a row's `MarshalYSON(*yson.Writer) error` method when present. A hand-rolled marshaler removes reflection from the hot path entirely.
- `ytrpc` calls `wire.Encode(rows)`, which is reflection-only on the row struct. There is no equivalent extension point: no `WireMarshaler` interface check, and no public way to construct a pre-encoded `yt.RowBatch` from outside the `rpcclient` internal package.

The per-field `encoding.BinaryMarshaler` / `encoding.TextMarshaler` hooks in `convertValue` exist, but they always emit a `TypeBytes` value. They cannot be used to populate `Any`-typed columns, which is what one typically wants for a column holding nested attributes.

### Concrete impact

Benchmarked with a typical structured-log row schema — a handful of primitive fields (a few uint64 timestamps, a few short strings) plus one `map[string]string` field stored as `Any`, optionally a longer free-form string field — for batches of 1000 rows, on Apple M4 Pro:

| payload shape | yson (HTTP) | wire (RPC, reflection) | wire hand-rolled¹ |
|---|---:|---:|---:|
| small, no map | 1.26ms | 2.65ms | 1.44ms |
| many-key map | 10.4ms | 12.7ms | 11.3ms |
| small JSON-y blob | 3.0ms | 4.5ms | 3.4ms |
| large free-form string | 21.0ms | 24.0ms | 22.8ms |

¹ A hand-rolled function building `wire.Row` directly from struct fields, with manual binary-YSON encoding of the map.

The reflection-based `wire.Encode` is consistently the heaviest. The hand-rolled variant matches HTTP closely (within ~10% on time, lower allocations) — but it is unreachable from production code because `ytrpc.NewClient.InsertRows` always calls `wire.Encode`, with no hook to substitute custom logic.

### Suggested directions

Any of the following would let downstream services migrate from HTTP to RPC without taking a CPU regression on the client:

1. **A `WireMarshaler` interface checked at the row level** in `wire.Encode` / `encodeReflect`, mirroring how `yson.Writer.Any` honors `MarshalYSON`. Signature could be e.g.

[...]

PR #1712, opened 2026-05-04
[go] wire: add RowMarshaler interface to bypass row-level reflection

orloffv:
Discussion: #1711

### Summary

Adds a row-level `RowMarshaler` interface to `wire.Encode`, mirroring how `yson.Writer.Any` honors `MarshalYSON`. If a value passed to `Encode` (or its pointed-to value) implements `RowMarshaler`, `MarshalRow(Resolver) (Row, error)` is called instead of the reflection-based path. Column ids come through a `Resolver` that wraps the per-call `NameTable`, so hand-rolled rows compose cleanly with reflection-encoded ones in the same batch.

This lets services that maintain custom YSON marshalers for the HTTP transport keep equivalent client-side performance after switching to the RPC transport. Today, `ytrpc` always goes through reflection in `wire.Encode`, while `ythttp` honors per-row `MarshalYSON` — so a service that already optimised its YSON path takes a CPU regression on migration.

### Public API

```go
// Resolver translates a column name to its NameTable id, registering the
// name on first use. Implementations must be safe for concurrent use.
type Resolver interface {
    LookupOrAdd(name string) uint16
}

// RowMarshaler is implemented by types that supply their own wire row
// encoding. If a value passed to Encode implements RowMarshaler,
// MarshalRow is called instead of reflection-based encoding.
type RowMarshaler interface {
    MarshalRow(r Resolver) (Row, error)
}
```

The change is strictly additive: types that don't implement `RowMarshaler` use the existing reflection path with no behaviour change.

### Tests

- `TestEncodeRowMarshaler` — encoded `NameTable` and `[]Row` match the expected output for a batch of two rows that implement `RowMarshaler`.
- `TestEncodeRowMarshalerTakesPrecedenceOverReflect` — a type that has both yson-tagged fields and `MarshalRow` is encoded via `MarshalRow` (the reflection-derived `NameTable` entries do not appear).

### Benchmark

`BenchmarkEncode` (added) compares `wire.Encode` against a generic row shape (a few primitive scalars + one `map[string]string` serialised as `Any`), batch=1000, on Apple M4 Pro:

| mode | ns/op | B/op | allocs/op |
|---|---:|---:|---:|
| reflect | 1.69ms | 6.40 MB | 69k |
| row_marshaler | 1.24ms | 5.27 MB | 31k |

The hand-rolled marshaler saves ~27% time and ~55% allocations on this shape.

### Why this matters

For services migrating high-throughput dynamic-table writers from `ythttp` to `ytrpc`, `wire.Encode`'s reflection-only path is currently a significant client-side CPU and allocation tax versus the YSON path with hand-rolled `MarshalYSON`. With this hook, a service can supply a `MarshalRow` that builds the row directly and recover the gap (in our internal benchmarks against a real production-like row shape, the closure to within ~10% of `MarshalYSON`).

The per-field `encoding.BinaryMarshaler` / `encoding.TextMarshaler` hooks already in `convertValue` are not sufficient: they always emit `TypeBytes`, so they cannot populate `Any`-typed columns that hold nested data.

### DIFF
--- a/yt/go/wire/encode.go
+++ b/yt/go/wire/encode.go
@@ -17,6 +17,48 @@ type NameTableEntry struct {
 	Name string
 }
 
+// Resolver translates a column name to its NameTable id, registering the
+// name on first use. Implementations must be safe for concurrent use,
+// because Encode invokes row marshaling in parallel.
+type Resolver interface {
+	LookupOrAdd(name string) uint16
+}
+
+// RowMarshaler is implemented by types that supply their own wire row
+// encoding. If a value passed to Encode (or its pointed-to value, when
+// the value is a non-nil pointer) implements RowMarshaler, Encode calls
+// MarshalRow instead of using reflection. Column ids must be obtained
+// via the supplied Resolver so that all rows in the batch reference the
+// same NameTable.
+type RowMarshaler interface {
+	MarshalRow(r Resolver) (Row, error)
+}
+
+// nameTableResolver wraps Encode's per-call indexMap so RowMarshaler
+// implementations can register column names without touching internals.
+type nameTableResolver struct {
+	indexMap map[NameTableEntry]uint16
+	mapLock  *sync.RWMutex
+}
+
+func (r *nameTableResolver) LookupOrAdd(name string) uint16 {
+	k := NameTableEntry{Name: name}
+	r.mapLock.RLock()
+	id, ok := r.indexMap[k]
+	r.mapLock.RUnlock()
+	if ok {
+		return id
+	}
+	r.mapLock.Lock()
+	defer r.mapLock.Unlock()
+	if id, ok = r.indexMap[k]; ok {
+		return id
+	}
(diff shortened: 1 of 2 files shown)


tinarsky:
Hello, I'll check it within a week

yt/go/wire/encode.go
```diff
@@ -17,6 +17,48 @@ type NameTableEntry struct {
 	Name string
 }
 
+// Resolver translates a column name to its NameTable id, registering the
+// name on first use. Implementations must be safe for concurrent use,
+// because Encode invokes row marshaling in parallel.
+type Resolver interface {
```
tinarsky: `NameTableBuilder`?

tinarsky (changes requested):

yt/go/wire/encode.go
orloffv: fixed

tinarsky (approved):

PR #1714, opened 2026-05-05
Fix memory management Bugs for `PyList_SetItem`

wr-web:
https://github.com/ytsaurus/ytsaurus/issues/1713


savnadya:
Hello! Thank you for the PR.`contrib/libs/apache/arrow` is not used in ytsaurus actually, we will remove it soon. YTsaurus uses other version of this library: `contrib/libs/apache/arrow_next`.

PR #1726, opened 2026-05-13
Secure mTLS secrets passed to job proxy via environment variables

koct9i:
**Implement Secure Environment in undumpable memory with zeroing**

Singleton GetSecureEnvironment() implements interface IAttributeDictionary.

Function MoveToSecureEnvironment() moves environment variables with given
names or name prefixes into secure environment and hides original values.

**Secure mTLS secrets passed to job proxy via environment variables**

Protect environment variables which are not should be forwarded to user job.
For example mTLS certificates passed to job-proxy via environment.

This solution is better than nothing for case when execution environments
are not completely isolated and provides some protection for all cases.

I.e. certificates will not be included as plain text in code dumps and
no longer exposed in /proc/pid/environ.

---

* Changelog entry
Type: feature
Component: map-reduce

Secure mTLS secrets passed to job proxy via environment variables

### DIFF
--- a/yt/yt/core/crypto/secure_environment.cpp
+++ b/yt/yt/core/crypto/secure_environment.cpp
@@ -1,0 +1,187 @@
+#include "secure_environment.h"
+
+#include <yt/yt/core/ytree/helpers.h>
+
+#include <yt/yt/library/undumpable/undumpable.h>
+
+#include <library/cpp/yt/memory/chunked_memory_pool.h>
+#include <library/cpp/yt/memory/leaky_singleton.h>
+
+#include <util/system/env.h>
+
+namespace NYT::NCrypto {
+
+using namespace NYTree;
+using namespace NYson;
+
+////////////////////////////////////////////////////////////////////////////////
+
+struct TSecureMemoryPoolTag
+{ };
+
+////////////////////////////////////////////////////////////////////////////////
+
+class TSecureMemoryChunkProvider
+    : public IMemoryChunkProvider
+{
+public:
+    std::unique_ptr<TAllocationHolder> Allocate(size_t size, TRefCountedTypeCookie cookie) override
+    {
+        return std::unique_ptr<THolder>(TAllocationHolder::Allocate<THolder>(size, cookie));
+    }
+
+private:
+    struct THolder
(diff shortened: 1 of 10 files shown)


yt/yt/core/crypto/secure_environment.cpp
```diff
@@ -0,0 +1,187 @@
...
+{
+    return LeakySingleton<TSecureEnvironmentHolder>()->Dictionary;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+void MoveToSecureEnvironment(const std::vector<TStringBuf>& names, const std::vector<TStringBuf>& namePrefixes)
+{
+    auto secureEnv = GetSecureEnvironment();
+    IterateEnv([&](TStringBuf name, TStringBuf value) {
```
k-pogorelov, line 172: ] (

yt/yt/server/job_proxy/program.cpp
```diff
@@ -129,6 +129,7 @@ class TJobProxyProgram
         ConfigureSingletons(config);
 
         auto jobProxy = New<TJobProxy>(std::move(config), OperationId_, JobId_);
+        jobProxy->SecureEnvironmentVariables();
```
k-pogorelov, line 132: Some threads could have already been spawned (in ConfigureSingletons)

So it looks like a potential data race.

Let's call SecureEnvironmentVariables before any thread is started.

k-pogorelov (changes requested):

yt/yt/core/crypto/secure_environment.cpp
```diff
@@ -0,0 +1,187 @@
...
+using namespace NYson;
+
+////////////////////////////////////////////////////////////////////////////////
+
+struct TSecureMemoryPoolTag
+{ };
+
+////////////////////////////////////////////////////////////////////////////////
+
+class TSecureMemoryChunkProvider
```
p4vook, line 24: We have TSecretString for that, see https://github.com/ytsaurus/ytsaurus/blob/5d39168f3ef552a9d62c6322c7a1a056bc04e821/yt/yt/library/signature/common/key_pair.h#L39

PR #1735, opened 2026-05-27
Go SDK: partial result error handling

mtvch:
Hello! Was missing error feedback when using EnablePartialResult option, so I added it.

* Changelog entry
Type: feature
Component: go-sdk

Added PartialResultError that is returned when EnablePartialResult is used and some tablets are unavailable.


tinarsky:
Hello, I'll review it within a week

tinarsky (approved):

PR #1736, opened 2026-05-27
Add quiet mode to check-permission using CLI

koct9i:
- **Problem**
  `yt check-permission` lacked a machine-friendly mode for branchable shell usage; it always printed payload output instead of returning a clean status-only signal.

- **Changes**
  - Added `-q` / `--quiet` to `yt check-permission`.
  - In quiet mode, command suppresses stdout and exits immediately with:
    - `0` when permission action is `allow`
    - `1` otherwise
  - Updated CLI docs (EN/RU) for new flag.
  - Added CLI test coverage for both allow/deny quiet-mode paths.

- **Behavior example**
  ```bash
  yt check-permission user read //path -q
  # no stdout; exit code is 0 if allowed, 1 if denied
  ```

---

* Changelog entry
Type: feature
Component: python-sdk

Add quiet mode to check-permission using CLI in shell

### DIFF
--- a/yt/docs/en/_includes/api/cli/commands.md
+++ b/yt/docs/en/_includes/api/cli/commands.md
@@ -2531,7 +2531,7 @@ usage: yt add-member [-h] [--params PARAMS] [--member MEMBER] [--group GROUP] [m
 checks permission for Cypress node
 
 ```bash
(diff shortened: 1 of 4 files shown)


denvr:
lgtm

denvr (approved):

PR #1743, opened 2026-06-05
Add job profiler support to Go and Python MapReduce SDKs

desertfury:
Allow user jobs to request CPU/memory profiles via the operation spec.

- Go SDK: new JobProfiler spec type and UserScript.Profilers field; the job runner collects a pprof profile (runtime/pprof) when the job proxy selects the job for profiling, writing it to the profile pipe.
- Python SDK: profilers() spec option on UserJobSpecBuilder and a py-spy-based CPU profiler started from py_runner_helpers when YT_JOB_PROFILER_SPEC is set.

### DIFF
--- a/yt/python/yt/wrapper/py_runner_helpers.py
+++ b/yt/python/yt/wrapper/py_runner_helpers.py
@@ -268,6 +268,31 @@ def check_allowed_structured_skiff_attributes(params):
         raise YtError("@yt.wrapper.with_context decorator may be used only with mappers")
 
 
+def start_user_job_profiler():
+    raw_spec = os.environ.get("YT_JOB_PROFILER_SPEC")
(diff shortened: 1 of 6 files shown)


yt/python/yt/wrapper/py_runner_helpers.py
Kontakter, line 271: I am not sure that explicit dependency on `py-spy` is the right way. 

By the way, why not to run this collector in `start()` method of your python operation?

denvr, line 271: The same thought. At now it seems as specific user case.
f.e.:
- enabling/configuration - better in config, not direct access to envs in function, not in operation spec
- how profiling results should passed from job
- implicit fixed dependency for py-spy

savnadya (changes requested):
Waiting for a user response

PR #1747, opened 2026-06-08
S3 writes and some tests

AleksandraZh:


yt/yt/server/master/chunk_server/chunk_replica.h
```diff
@@ -237,6 +237,8 @@ struct TSequoiaChunkReplica
     NNodeTrackerClient::TChunkLocationIndex LocationIndex = NNodeTrackerClient::InvalidChunkLocationIndex;
     // Not persisted, used for getting StoredReplicas attribute.
     EChunkReplicaState ReplicaState = EChunkReplicaState::Generic;
+    // Not persisted (yet), used for master offshore replicas.
+    int MediumIndex = -1;
```
achulkov2, line 241: medium id?
no migration if we ever get rid of medium indices (at least for offshore media)

yt/yt/ytlib/chunk_client/replication_writer.cpp
```diff
@@ -913,43 +950,62 @@ class TReplicationWriter
...
+        bool isOffshore = (target.GetNodeId() == OffshoreNodeId);
+
+        // For offshore replicas there is no real data node; the request goes to
+        // the Offshore Data Gateway which proxies writes to S3.  We use a
+        // synthetic TNodeDescriptor so that the rest of the writer can treat
+        // this node uniformly (logging, traffic accounting, etc.).
+        TNodeDescriptor nodeDescriptor;
+        std::string address;
+        if (isOffshore) {
+            address = "offshore-gateway";
```
achulkov2, line 962: We have some similar address in replication_reader, values should probably be in sync.

yt/yt/server/master/chunk_server/chunk.cpp
```diff
@@ -383,13 +383,10 @@ void TChunk::AddReplica(
...
             return;
         }
-        // TODO(cherepashka): allow once offshore media are implemented.
-        YT_LOG_ALERT(
-            "Attempted to add offshore medium stored replica for chunk, ignored "
-            "(ChunkId: %v, ReplicaIndex: %v, MediumIndex: %v)",
-            GetId(),
-            replica.GetReplicaIndex(),
-            replica.GetEffectiveMediumIndex());
+        if (approved) {
```
achulkov2, line 386: alert/verify for non-approved offshore replicas?
shouldn't be any

yt/yt/tests/integration/s3/test_s3.py
```diff
@@ -0,0 +1,659 @@
...
+    # TODO(achulkov2): S3 medium as intermediate medium. Implement or disable for now.
+
+    # TODO(achulkov2): Test master interaction with tables with offshore replicas: concat.
+    # TODO(achulkov2): Test forbidden operations: erasure, journals, remote copy jobs (difficult to get creds from other cluster for now),
+    # master jobs: replication, merge?, repair (sanity), reincarnation?
+
+    # TODO(achulkov2): Test files.
+
+    # TODO(pavel-bash): Test ban mechanism in replication reader does not ban the offshore data gateway but only a replica
+    # when we support multiple replicas; example test as a git diff: https://paste.nebius.dev/paste/993b1203-e882-4256-ad57-762f42043e9c.html.
```
achulkov2, line 651: kek
@pavel-bash can you share this snippet in non-private paste tool?

yt/yt/server/master/chunk_server/chunk_manager.cpp
```diff
@@ -2336,6 +2358,10 @@ class TChunkManager
         std::optional<int> hintIndex,
         TObjectId hintId) override
     {
+        if (GetDynamicConfig()->AllowOffshoreMedia) {
```
achulkov2, line 2361: missing negation

yt/yt/ytlib/chunk_client/replication_writer.cpp
```diff
@@ -913,43 +950,62 @@ class TReplicationWriter
...
-                    this_->Nodes_.end(),
-                    [&] (const auto& node) {
-                        return node->GetDefaultAddress() == address && !node->IsAlive();
-                    }) == 0;
-
-                return needRetry;
-            }));
+        IChannelPtr channel;
+        if (isOffshore) {
+            channel = Client_->GetNativeConnection()->GetOffshoreDataGatewayChannel();
```
achulkov2, line 978: won't work when there is more than one odg instance -- balancing channel would break session locality

achulkov2, line 978: also this channel is retrying by default, not good for writes :)

yt/yt/tests/integration/s3/test_s3.py
```diff
@@ -0,0 +1,659 @@
...
+            "bucket": uuid.uuid4().hex,
+        },
+        {
+            "name": "s3_extra2",
+            "bucket": uuid.uuid4().hex,
+        },
+        {
+            "name": "s3_prefixed",
+            "bucket": uuid.uuid4().hex,
+            "prefix": "prefix/",
```
achulkov2, line 78: This option doesn't exist in upstream, as far as I can see.

It is actually best to add before any real production setup, to not have to do cleaning/migration of existing chunks.

@pavel-bash can you share our code here, it is something simple

achulkov2 (changes requested):

yt/yt/tests/integration/s3/test_s3.py
pavel-bash: Sure, here it is: https://gist.github.com/pavel-bash/194807e9058f87c5597f367d1a2ef821.

yt/yt/tests/integration/s3/test_s3.py
pavel-bash: The code is a little bit different in the upstream, so I've tried to converge the two versions. Here are my suggested additions: https://gist.github.com/pavel-bash/08a76a67df8a41fe4b735606a544d019.

yt/yt/ytlib/chunk_client/medium_directory_synchronizer.cpp
```diff
@@ -126,7 +126,10 @@ class TMediumDirectorySynchronizer::TImpl
             auto client = connection->CreateClient(TClientOptions::FromUser(NSecurityClient::RootUserName));
 
             TGetClusterMetaOptions options;
-            options.ReadFrom = NApi::EMasterChannelKind::MasterSideCache;
+            if (Config_->UseCache) {
+                // I'll fix that.
```
pavel-bash, line 130: A ToDo so that it's not forgotten?

yt/yt/server/master/chunk_server/chunk_replica.h
AleksandraZh: It is called mediumIndex in master chunk replicas

PR #1755, opened 2026-06-16
yt/yt/core/bus: test catching exception from TSslContext::ApplyConfig

koct9i:
Test for "YT-27641: Abort connection if SSL certificates/keys/CAs can not be read"

This code is executed inside RunNoExcept context and cannot throw.
There many reasons for errors within ApplyConfig, in test is used
incorrect SslConfigurationCommands is just for simplicity.


yalekseev:
Shipped!

PR #1758, opened 2026-06-19
[queues] Use response codecs for PullRows in native api

TheLegionCrazy:
---
* Changelog entry
Type: fix
Component: native-api / proxy

Native connection config has lookup/select RPC codecs but "native tablet node pull" never used them unlike legacy "select pull"

### DIFF
--- a/yt/yt/ytlib/api/native/client_dynamic_tables_impl.cpp
+++ b/yt/yt/ytlib/api/native/client_dynamic_tables_impl.cpp
@@ -3130,7 +3130,10 @@ IUnversionedRowsetPtr TClient::DoPullQueueViaTabletNodeApi(
     ValidateTabletMountedOrFrozen(tableInfo, tabletInfo);
     auto channel = GetReadCellChannelOrThrow(tabletInfo->CellId);
     TQueryServiceProxy proxy(channel);
-    proxy.SetDefaultTimeout(options.Timeout.value_or(Connection_->GetConfig()->DefaultFetchTableRowsTimeout));
+
+    const auto& connectionConfig = Connection_->GetConfig();
+    proxy.SetDefaultResponseCodec(connectionConfig->SelectRowsResponseCodec);
+    proxy.SetDefaultTimeout(options.Timeout.value_or(connectionConfig->DefaultFetchTableRowsTimeout));
 
     auto req = proxy.FetchTableRows();
     ToProto(req->mutable_tablet_id(), tabletInfo->TabletId);
--- a/yt/yt/ytlib/api/native/client_dynamic_tables_impl.cpp
+++ b/yt/yt/ytlib/api/native/client_dynamic_tables_impl.cpp
@@ -3131,10 +3131,6 @@ IUnversionedRowsetPtr TClient::DoPullQueueViaTabletNodeApi(
     auto channel = GetReadCellChannelOrThrow(tabletInfo->CellId);
     TQueryServiceProxy proxy(channel);
 
-    const auto& connectionConfig = Connection_->GetConfig();
-    proxy.SetDefaultResponseCodec(connectionConfig->SelectRowsResponseCodec);
-    proxy.SetDefaultTimeout(options.Timeout.value_or(connectionConfig->DefaultFetchTableRowsTimeout));
-
     auto req = proxy.FetchTableRows();
     ToProto(req->mutable_tablet_id(), tabletInfo->TabletId);
     ToProto(req->mutable_cell_id(), tabletInfo->CellId);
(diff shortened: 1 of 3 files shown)


yt/yt/ytlib/api/native/client_dynamic_tables_impl.cpp
```diff
@@ -3130,7 +3130,10 @@ IUnversionedRowsetPtr TClient::DoPullQueueViaTabletNodeApi(
     ValidateTabletMountedOrFrozen(tableInfo, tabletInfo);
     auto channel = GetReadCellChannelOrThrow(tabletInfo->CellId);
     TQueryServiceProxy proxy(channel);
-    proxy.SetDefaultTimeout(options.Timeout.value_or(Connection_->GetConfig()->DefaultFetchTableRowsTimeout));
+
+    const auto& connectionConfig = Connection_->GetConfig();
+    proxy.SetDefaultResponseCodec(connectionConfig->SelectRowsResponseCodec);
```
savnadya: Let's add "PullQueueResponseCodec"?
And set it via`req->set_response_codec(...)`

savnadya (dismissed):

yt/yt/ytlib/api/native/client_dynamic_tables_impl.cpp
TheLegionCrazy: Done.

I've set `None` codec by default for compatibility, so new release will not increase proxy/tabnode cpu usage. 
Is that okay?

TheLegionCrazy:
@savnadya hi, any news?

yt/yt/ytlib/api/native/client_dynamic_tables_impl.cpp
ItIsApachee: @TheLegionCrazy you didn't address "And set it via `req->set_response_codec(...)`" part of the issue.

ItIsApachee:
@TheLegionCrazy I've unresolved the conversation above, since it was not completely addressed. I'll take a look again once you fix this.

yt/yt/ytlib/api/native/client_dynamic_tables_impl.cpp
TheLegionCrazy: @ItIsApachee i did.
https://github.com/ytsaurus/ytsaurus/pull/1758/changes#diff-b921182110a6e142f884e4b408237cbe708a92f41018faa44cb00ece2c4ebd6eR3146

ItIsApachee: Can you explain to me why we use upper case version here? Select/lookup rows requests use lower case version. It might be that select/lookup rows add this field to protobuf and we don't, but I want to know the reason why.
Can you look into the difference here between pull queue/consumer and select/lookup rows? Maybe this change requires changes to tablet node code as well?

ItIsApachee: cc @TheLegionCrazy

ItIsApachee: It might be that what we use currently is the codec for the entire request and what lookup/select rows use is the codec for the wire format only.
I don't think we should deviate away from the established convention here.

TheLegionCrazy: lowercase `set_response_codec` is coming from protobuf request attributes: https://github.com/ytsaurus/ytsaurus/blob/2d08879faffc5dd96745dc0b80ec520ce33a24a1/yt/yt/library/query/proto/query_service.proto#L70-L75

This actually looks like a legacy workaround since these fields are marked with "TODO" and "SetResponseCodec" does compression/decompression on the bus level (also in rpc threads), there is no need to do this in a boilerplate way.

ItIsApachee: Might be true, I've asked the dynamic table team for now, I'd like to wait a bit if someone responds with more details.
One thing to mention is that the codec feature is rarely used in our installation, what's your goal with this? By the way, also consider that lookup/select rows response codecs are used for requests to tablet nodes and might not be used for proxy response (I don't know if that's the case or not), which might defeat the purpose.

TheLegionCrazy: >  what's your goal with this?

To exchange CPU for NET in tabnode ↔ proxy requests.
We've rolled out this patch to one of our clusters and now see pretty good results in pulling queues.

> for requests to tablet nodes and might not be used for proxy response

it is true, these codecs are only used for tabnode ↔ proxy requests.
But codecs between client and proxy are configured in client config.

yt/yt/ytlib/api/native/config.h
```diff
@@ -310,6 +310,8 @@ struct TConnectionDynamicConfig
     std::optional<TDuration> LookupRowsInMemoryLoggingSuppressionTimeout;
     std::optional<TDuration> LookupRowsExtMemoryLoggingSuppressionTimeout;
 
+    NCompression::ECodec PullQueueResponseCodec;
+
```
ItIsApachee: A slight nitpick. I feel like this feature should be opt-in, i.e. `std::optional<NCompression::ECodec>` instead and conditional `SetResponseCodec`.
Setting codec to none might in theory disable compression codec set at the channel proxy level.
It would be best to leave the behavior as is in the default case, even in case of slight changes in other parts, e.g. if someone were to add default proxy level codec.

ItIsApachee (changes requested):
Still waiting to hear from anyone that knows anything about codecs, but for now here is a small issue I found, but otherwise LGTM.
Let's fix the changes I requested and if nobody replies in a week or so, I'll merge it.

ItIsApachee (approved):
LGTM

Issue #846, opened 2024-09-13
python CLI client ignores system certs


faucct:
The problem is that the `requests` library only uses certificates from `certifi`. Doing `truststore.inject_into_ssl()` in executable files is a solution recommended by `requests` authors:
https://github.com/psf/requests/issues/2966#issuecomment-2119324292

PR #1767, opened 2026-06-29
Trust system certificate store in yt (#846)

desertfury:
The yt Python client makes HTTPS requests through the vendored requests, which verifies TLS certificates only against the bundled certifi CA list and ignores the operating system trust store (e.g. corporate root certificates). See #846.

  When proxy/use_system_ca is enabled (default), the client now builds an ssl.create_default_context() — verifying against the OS trust store and honoring SSL_CERT_FILE/SSL_CERT_DIR, REQUESTS_CA_BUNDLE/CURL_CA_BUNDLE, and an explicit proxy/ca_bundle_path — and binds it to its own requests session adapter. The vendored certifi bundle is kept only as a fallback when no system store is available. Trust is assembled per session instead of mutating the process-global REQUESTS_CA_BUNDLE, so there is no cross-client side effect, and it applies to any yt client, not just the CLI. No new dependency is required.

  Add an integration test (test_https.py) on a local TLS cluster with a self-signed CA: it reproduces #846, verifies the fix, checks that a CA provided only via REQUESTS_CA_BUNDLE is honored, that an untrusted cert is still rejected, and that ca_bundle_path keeps working.

  Changelog entry
  Type: fix
  Component: python-sdk
  The yt HTTP client now trusts the operating system certificate store (certifi as fallback), configurable via proxy/use_system_ca.

### DIFF
--- a/yt/python/yt/wrapper/http_helpers.py
+++ b/yt/python/yt/wrapper/http_helpers.py
@@ -18,6 +18,7 @@ from yt.common import _pretty_format_for_logging, get_fqdn as _get_fqdn
 import io
 import os
 import re
+import ssl
 import sys
 import time
 import types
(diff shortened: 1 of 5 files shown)


koct9i:
I like this. 

But I would prefer to use vendored CA bundle only as failback when system bundle isn't found.

Internal CA for non-opensouece use should be placed into separate file, which should be empty/missing in opensource version.

desertfury:
wow. there's a preexisting bug

http_helpers.py:144 os.environ["REQUESTS_CA_BUNDLE"] = ca_bundle_path

When multiple clients using different ca_bundle_path, the last one wins for the entire process and leaks to non-ytsaurus code in the same interpreter.

yt/python/yt/wrapper/http_helpers.py
denvr, line 124: Bug. In case of internal Arcaida builds this function (`certs.where()`) returns cert content, not path (but path to cert in internal whl package)

denvr, line 174: Design.
As i can see - system CA presents almost all nix systems and vendored CA will almost never used.
For internal usage vendored CA should always used (with system CA)
(vendored CA presents only in internal builds)

yt/python/yt/wrapper/tests/test_https.py
denvr, line 75: minor
Please, mark author as opensource (prefix)

yt/python/yt/wrapper/http_helpers.py
denvr, line 129: Feel free with type hints anywhere

denvr, line 268: minor bug
It seems we starts silently ignore potential usage of `verify` in code (`verify=False`, `verify=/some/path.pem`)

denvr (changes requested):

PR #1772, opened 2026-07-02
Fix yql_complete --pos option .

Tony-Romanov:
Type: fix
Component: yql

### DIFF
--- a/yql/essentials/tools/yql_complete/yql_complete.cpp
+++ b/yql/essentials/tools/yql_complete/yql_complete.cpp
@@ -57,7 +57,7 @@ int Run(int argc, char** argv) {
     TString inQueryText;
     TString freqFileName;
     TString schemaFileName;
-    TMaybe<ui64> pos;
+    std::optional<ui64> pos;
     opts.AddLongOption('i', "input", "input file").RequiredArgument("input").StoreResult(&inFileName);
     opts.AddLongOption('q', "query", "input query text").RequiredArgument("query").StoreResult(&inQueryText);
     opts.AddLongOption('f', "freq", "frequences file").StoreResult(&freqFileName);
@@ -120,7 +120,8 @@ int Run(int argc, char** argv) {
         MakePureLexerSupplier(),
         NSQLComplete::MakeUnionNameService(std::move(services), ranking));
 
-    if (!NYql::IsUtf8(queryString)) {
+    size_t size = 0U;
+    if (!(NYql::IsUtf8(queryString) && GetNumberOfUTF8Chars(queryString.data(), queryString.size(), size))) {
         ythrow yexception() << "provided input is not UTF encoded";
     }
 
@@ -129,6 +130,7 @@ int Run(int argc, char** argv) {
     }
 
     NSQLComplete::TCompletionInput input = NSQLComplete::SharpedInput(queryString);
+    input.CursorPosition = std::min(size, pos.value_or(size));
 
     auto output = engine->CompleteAsync(input).ExtractValueSync();
     for (const auto& c : output.Candidates) {


vityaman (changes requested):
1. Why TMaybe is replaced to std::optional?
2. Could you, please, add a reproducer test?

Tony-Romanov:
> 1. Why TMaybe is replaced to std::optional?

More common with the same functionality.  

> 2. Could you, please, add a reproducer test?

No, I didn't see any tests for this tool, or for any other tools in folder.

vityaman:
@Tony-Romanov, 

1. We in YQL prefer `util` primitives over `stdlib`. I agree with you and using stdlib is better, but this is a rule. Maybe somewhen we would find-and-replace `util` primitives.

2. Ah, I thought that the change goes to `yql/essentials/sql/v1/ide`. Okay, no need for test.

3. Why this change is needed? The tool terminates? I thought it should throw an error from a completion engine internals, that is caught at the tool `main` and `what` is printed.

4. Are you interested in using an autocompletion engine? If so, feel free to ask for some features, requirements and share ideas.

Tony-Romanov:
@vityaman 

>     1. We in YQL prefer `util` primitives over `stdlib`. I agree with you and using stdlib is better, but this is a rule. Maybe somewhen we would find-and-replace `util` primitives.

When I worked at YQL three years ago, there were no strict rules prohibiting the use of std.
On the contrary, I always preferred using std for new code unless it required interoperability with the old code at the interface level. 
Has YQL really gotten that much worse since then? 
YTsaurus is moving in the right direction right [now](https://github.com/ytsaurus/ytsaurus/commit/d465c17a385e7a7dee88f76baa8e5097132b4c40). 

>     3. Why this change is needed? The tool terminates? I thought it should throw an error from a completion engine internals, that is caught at the tool `main` and `what` is printed.

>     4. Are you interested in using an autocompletion engine? If so, feel free to ask for some features, requirements and share ideas.


Yesterday, we talked with our analysts about what they'd like to see from YQL, among other things, and one of the points was that your autocomplete leaves a lot to be desired.

Today, I came in for the first time to see what it is and how it works. I started with the simplest utility and accidentally noticed a glaring bug, fixed it, threw it in PR, and moved on.

And then I created a tool that can really quickly and easily test autocomplete, [here](https://github.com/ytsaurus/ytsaurus/pull/1773).

As for this PR, you can commit it, or leave the option missfunctional, or make changes to this PR yourself, I don't care anymore.

yql/essentials/tools/yql_complete/yql_complete.cpp
```diff
@@ -129,6 +130,7 @@ int Run(int argc, char** argv) {
     }
 
     NSQLComplete::TCompletionInput input = NSQLComplete::SharpedInput(queryString);
+    input.CursorPosition = std::min(size, pos.value_or(size));
```
vityaman, line 134: I do not think this is a correct logic, because `CursorPosition` is in bytes on the API level and in UTF-8 characters in an analysis level. Also it seems to overwrite a `CursorPosition` set by `NSQLComplete::SharpedInput`.

vityaman (changes requested):

vityaman:
@Tony-Romanov, could you, please, assist me to find a string, on which a bug happens? What should be a content of file `a.txt` and value of `-p` so it raises? `cat a.txt | ./yql_complete -p <pos>`?

vityaman:
@Tony-Romanov, are you going to implement an autocompletion API for a YTsaurus OSS Web UI? If so, we need to discuss it, because I planned to make a common LSP-like library for YQL this year and then to integrate it everywhere. I dream that there should be a unified Rest API described via OpenAPI and an TypeScript SDK (based on generated `TypeScript` client).

yql/essentials/tools/yql_complete/yql_complete.cpp
vityaman: Maybe the decision to make it in bytes is a mistake and I need to change it, because it seems that almost all clients (except YDB CLI https://github.com/ydb-platform/ydb/blob/a52ffb856a584a5099e37b9c060da3ba4a3db5fb/ydb/public/lib/ydb_cli/commands/interactive/complete/yql_completer.cpp#L64) expect it to be in characters.

Tony-Romanov:
> @Tony-Romanov, could you, please, assist me to find a string, on which a bug happens? What should be a content of file `a.txt` and value of `-p` so it raises? `cat a.txt | ./yql_complete -p <pos>`?

Before fix:
```
Tony@fedora:~/Projects/ytsaurus/yql/essentials/tools/yql_complete$ ./yql_complete --query="sEleCt Sta from AS_TABLE([<||>]) A"
[Keyword] AS
[Keyword] ASSUME
Tony@fedora:~/Projects/ytsaurus/yql/essentials/tools/yql_complete$ ./yql_complete --query="sEleCt Sta from AS_TABLE([<||>]) A" --pos=3 
[Keyword] AS
[Keyword] ASSUME
Tony@fedora:~/Projects/ytsaurus/yql/essentials/tools/yql_complete$ ./yql_complete --query="sEleCt Sta from AS_TABLE([<||>]) A" --pos=10 
[Keyword] AS
[Keyword] ASSUME
Tony@fedora:~/Projects/ytsaurus/yql/essentials/tools/yql_complete$ ./yql_complete --query="sEleCt Sta from AS_TABLE([<||>]) A"
[Keyword] AS
[Keyword] ASSUME
```

After fix:
```
Tony@fedora:~/Projects/ytsaurus/yql/essentials/tools/yql_complete$ ./yql_complete --query="sEleCt Sta from AS_TABLE([<||>]) A" --pos=3 
[Keyword] SELECT
Tony@fedora:~/Projects/ytsaurus/yql/essentials/tools/yql_complete$ ./yql_complete --query="sEleCt Sta from AS_TABLE([<||>]) A" --pos=10 
[FunctionName] StartsWith()
[FunctionName] StaticMap()
[FunctionName] StablePickle()
[FunctionName] StaticZip()
[FunctionName] StaticFold()
[FunctionName] StaticFold1()
Tony@fedora:~/Projects/ytsaurus/yql/essentials/tools/yql_complete$ ./yql_complete --query="sEleCt Sta from AS_TABLE([<||>]) A"
[Keyword] AS
[Keyword] ASSUME

```

Tony-Romanov:
> @Tony-Romanov, are you going to implement an autocompletion API for a YTsaurus OSS Web UI? If so, we need to discuss it, because I planned to make a common LSP-like library for YQL this year and then to integrate it everywhere. I dream that there should be a unified Rest API described via OpenAPI and an TypeScript SDK (based on generated `TypeScript` client).


No, that's not true. Autocompletion in YT OSS UI is already implemented and working.
But users complain that it works poorly and ask for it to be fixed. This isn't about the API, but about the logic of the autocompletion library itself.
And at this point, I'm not ready to offer any specific improvements to its logic, although I understand that users don't like it.
So I started by creating (fixing) a couple of convenient tools that significantly speed up the process: change, build, and check, both manually and using an AI agent.

vityaman:
@Tony-Romanov, as I know, an autocompletion in YT Web UI now is implemented fully in `TypeScript` and based on `ydb-platform-monaco-yql-languages` and `gravity-ui/websql-autocomplete`. It does not use the `yql/essentials/sql/v1/ide` yet.

The C++ library is not compatible with a web browser runtime and is used in `YDB CLI` and also in some other one service. 

So this is why I am going
> to make a common LSP-like library for YQL this year and then to integrate it everywhere

Tony-Romanov:
> The C++ library is not compatible with a web browser runtime and is used in `YDB CLI` and also in some other one service.

Interesting. But some of the things users complained about, as far as I can see, work just as well with the C++ library. But some work better, like UDFs.

And I'm not responsible for YT UI; I can only create tickets for them.

vityaman:
Will approve to import the PR, then replace `TMaybe` with `std::optional`, then will merge. And also I will debug your cases, because I am unable to understand statically.

vityaman (approved):

vityaman:
@Tony-Romanov, let's discuss, what problems are present in a current completion engine and how to you would like to report them. 

I propose to make a GitHub Issue, to which we both will be subscribed. You will send test cases via comments there and I will fix them. And we will have discussion there, because it is more suitable, than PR comments. 

What do you think about it?

nepal (approved):

Tony-Romanov:
> I propose to make a GitHub Issue, to which we both will be subscribed. You will send test cases via comments there and I will fix them. And we will have discussion there, because it is more suitable, than PR comments.
> 
> What do you think about it?

Ok. Let's commit yqlrun first—it's easy to try everything.

But some things are immediately much better after switching to the new API in the UI, like UDFs or pragmas.

And some are still bad. For example, the user didn't clearly articulate this, but I tried it and realized what the problem was. The problem is that the lack of a predefined column separator ',' or between explicit expressions ';' almost fully breaks autocompletion. With library lookup points, it's clear—invalid SQL. But a human normally enters, for example, the name of a column or function first, and only then a comma. I'm sure this can be improved.

vityaman:
The problem was that I forgot to implement a logic to handle `--pos` argv. I changed your solution a little. Thanks! I'll create a GitHub issue for a future discussion.

vityaman:

PR #1773, opened 2026-07-02
Add autocompletion to yqlrun's SQL editor.

Tony-Romanov:
Type: yql
Component: feature

### DIFF
--- a/yql/tools/yqlrun/http/sql_complete_servlet.cpp
+++ b/yql/tools/yqlrun/http/sql_complete_servlet.cpp
@@ -1,0 +1,284 @@
+#include "sql_complete_servlet.h"
+
+#include <yt/yql/providers/yt/common/yql_names.h>
+
+#include <yql/essentials/sql/v1/ide/completion/name/cluster/static/discovery.h>
+#include <yql/essentials/sql/v1/ide/completion/name/object/simple/schema.h>
+#include <yql/essentials/sql/v1/ide/completion/name/object/simple/static/schema.h>
+#include <yql/essentials/sql/v1/ide/completion/name/service/cluster/name_service.h>
+#include <yql/essentials/sql/v1/ide/completion/name/service/impatient/name_service.h>
+#include <yql/essentials/sql/v1/ide/completion/name/service/schema/name_service.h>
+#include <yql/essentials/sql/v1/ide/completion/name/service/static/name_service.h>
+#include <yql/essentials/sql/v1/ide/completion/name/service/static/name_set.h>
+#include <yql/essentials/sql/v1/ide/completion/name/service/union/name_service.h>
+#include <yql/essentials/sql/v1/ide/completion/sql_complete.h>
+
+#include <yql/essentials/sql/v1/lexer/antlr4_pure/lexer.h>
+#include <yql/essentials/sql/v1/lexer/antlr4_pure_ansi/lexer.h>
+
+#include <library/cpp/json/json_reader.h>
+#include <library/cpp/json/json_writer.h>
+#include <library/cpp/yson/node/node_io.h>
+
+#include <util/stream/buffer.h>
+
+namespace {
+
+using namespace NSQLComplete;
+
+TString CandidateKindToString(ECandidateKind kind) {
+    switch (kind) {
+        case ECandidateKind::Keyword:
+            return "keyword";
+        case ECandidateKind::PragmaName:
+            return "pragma";
+        case ECandidateKind::TypeName:
+            return "type";
+        case ECandidateKind::FunctionName:
+            return "function";
+        case ECandidateKind::HintName:
+            return "hint";
+        case ECandidateKind::FolderName:
+            return "folder";
+        case ECandidateKind::TableName:
+            return "table";
+        case ECandidateKind::ClusterName:
+            return "cluster";
+        case ECandidateKind::ColumnName:
+            return "column";
+        case ECandidateKind::BindingName:
+            return "binding";
+        case ECandidateKind::UnknownName:
+            return "unknown";
+    }
+    return "unknown";
+}
+
+TLexerSupplier MakeLexerSupplier(NSQLTranslationV1::TLexers lexers) {
+    return [lexers = std::move(lexers)](bool ansi) {
+        return NSQLTranslationV1::MakeLexer(
+            lexers,
+            ansi,
+            NSQLTranslationV1::ELexerFlavor::Pure);
+    };
+}
+
+TVector<TString> ExtractColumnsFromTableAttr(const TString& tableAttr) try {
+    TVector<TString> columns;
+
+    if (tableAttr.empty()) {
+        return columns;
+    }
+
+    const auto root = NYT::NodeFromYsonString(tableAttr);
+    if (!root.IsMap() || !root.HasKey(NYql::YqlRowSpecAttribute)) {
+        return columns;
+    }
+
+    const auto& rowSpec = root[NYql::YqlRowSpecAttribute];
+    if (!rowSpec.IsMap() || !rowSpec.HasKey("Type")) {
+        return columns;
+    }
+
+    const auto& type = rowSpec["Type"];
+    if (!type.IsList() || type.AsList().size() < 2U) {
+        return columns;
+    }
+
+    const auto& typeList = type.AsList();
+    if (!typeList[0].IsString() || typeList[0].AsString() != "StructType" || !typeList[1].IsList()) {
+        return columns;
+    }
+
+    for (const auto& member : typeList[1].AsList()) {
+        if (member.IsList() && !member.AsList().empty() && member[0].IsString()) {
+            columns.emplace_back(member[0].AsString());
+        }
+    }
+
+    return columns;
+} catch (...) {
+    return {};
+}
+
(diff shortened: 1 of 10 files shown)


yql/tools/yqlrun/http/sql_complete_servlet.cpp
```diff
@@ -0,0 +1,284 @@
...
+            return "cluster";
+        case ECandidateKind::ColumnName:
+            return "column";
+        case ECandidateKind::BindingName:
+            return "binding";
+        case ECandidateKind::UnknownName:
+            return "unknown";
+    }
+    return "unknown";
+}
```
vityaman: Let's just use default `ToString`: https://github.com/ytsaurus/ytsaurus/blob/main/yql/essentials/sql/v1/ide/completion/sql_complete.cpp#L259. It seems a representation a text for these internal UI is not very important, but the code amount will be smaller.

yql/tools/yqlrun/http/sql_complete_servlet.cpp
```diff
@@ -0,0 +1,284 @@
...
+    for (const auto& member : typeList[1].AsList()) {
+        if (member.IsList() && !member.AsList().empty() && member[0].IsString()) {
+            columns.emplace_back(member[0].AsString());
+        }
+    }
+
+    return columns;
+} catch (...) {
+    return {};
+}
```
vityaman: Is it possible to at least log an error here?

yql/tools/yqlrun/http/sql_complete_servlet.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+    for (const TString& cluster : clusters) {
+        AddTable(data, cluster, "Input", inputColumns);
+        AddTable(data, cluster, "Output", {});
+        AddTable(data, cluster, outputTable, {});
+    }
+
+    TVector<INameService::TPtr> children = {
+        std::move(staticNames),
+        MakeImpatientNameService(
```
vityaman: It the `ImpatientNameService` really needed? It is needed to deal with async requests, that might be long to wait. The idea is that `ImpatientNameService` decorates `CachedNameService` with `MakeLocalCache`. `CachedNameService` blocks if there is no entry in RAM. `ImpatientNameService` forgets about an async request and returns immediately, so user does not wait, and in a background the cache will be filled later, so on a next request an answer will be immediate.

yql/tools/yqlrun/http/sql_complete_servlet.cpp
```diff
@@ -0,0 +1,284 @@
...
+            MakeSchemaNameService(
+                MakeSimpleSchema(
+                    MakeStaticSimpleSchema(std::move(data))))),
+        std::move(clusterNames),
+    };
+
+    return MakeUnionNameService(std::move(children), std::move(ranking));
+}
+
+void WriteCompletedToken(NJson::TJsonWriter& writer, const TCompletedToken& token, const TString& program) {
```
vityaman: It is better to prefer `TStringBuf` over `const TString&`, when a null-termination is not required.

```suggestion
void WriteCompletedToken(NJson::TJsonWriter& writer, const TCompletedToken& token, TStringBuf program) {
```

yql/tools/yqlrun/http/sql_complete_servlet.cpp
```diff
@@ -0,0 +1,284 @@
...
+    TSchemaData data;
+
+    const TVector<TString> inputColumns = ExtractColumnsFromTableAttr(tableAttr);
+    const TVector<TString> clusters = {"", "plato"};
+
+    for (const TString& cluster : clusters) {
+        AddTable(data, cluster, "Input", inputColumns);
+        AddTable(data, cluster, "Output", {});
+        AddTable(data, cluster, outputTable, {});
+    }
```
vityaman: It is better to make a separate factory function `MakeRequestSimpleSchema` for `MakeStaticSimpleSchema`.

Then make a `MakeRequestNameService` a private method of `TSqlCompleteServlet`, so it will have less args.

yql/tools/yqlrun/http/sql_complete_servlet.cpp
```diff
@@ -0,0 +1,284 @@
...
+namespace NHttp {
+
+TSqlCompleteServlet::TSqlCompleteServlet() {
+    Lexers_.Antlr4Pure = NSQLTranslationV1::MakeAntlr4PureLexerFactory();
+    Lexers_.Antlr4PureAnsi = NSQLTranslationV1::MakeAntlr4PureAnsiLexerFactory();
+
+    TFrequencyData frequency = LoadFrequencyData();
+    Ranking_ = MakeDefaultRanking(frequency);
+    StaticNameService_ = MakeStaticNameService(LoadDefaultNameSet(), frequency);
+    ClusterNameService_ = MakeImpatientNameService(
```
vityaman: Here an impatience is redundant too.

yql/tools/yqlrun/http/sql_complete_servlet.cpp
```diff
@@ -0,0 +1,284 @@
...
+    const bool parsed = NJson::ReadJsonFastTree(body, &value, true);
+    Y_ENSURE_EX(parsed, THttpError(HTTP_BAD_REQUEST) << "can't parse json");
+
+    const auto program = value["program"].GetStringSafe();
+    const auto cursorPosition = value["cursorPosition"].GetUIntegerSafe();
+    Y_ENSURE_EX(cursorPosition <= program.size(), THttpError(HTTP_BAD_REQUEST) << "invalid cursor position");
+
+    const auto tableAttr = value["tableAttr"].GetStringSafe();
+    const auto parameters = value["parameters"].GetStringSafe();
+    const auto outputTable = value["outputTable"].GetStringSafe();
```
vityaman, line 217: I suggest to make a struct like `TCompletionRequestBody` with a static method `Parse`.

yql/tools/yqlrun/http/sql_complete_servlet.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+    return MakeUnionNameService(std::move(children), std::move(ranking));
+}
+
+void WriteCompletedToken(NJson::TJsonWriter& writer, const TCompletedToken& token, const TString& program) {
+    const auto length = token.Content.size();
+
+    writer.Write("completedToken"sv);
+    writer.OpenMap();
+    writer.Write("content"sv, program.substr(token.SourcePosition, length));
```
vityaman: Why not just `token.Content`? It is different from a program substr? It is defined as https://github.com/ytsaurus/ytsaurus/blob/main/yql/essentials/sql/v1/ide/completion/sql_complete.cpp#L107-112

yql/tools/yqlrun/http/yql_server.cpp
```diff
@@ -657,6 +658,7 @@ TAutoPtr<TYqlServer> CreateYqlServer(
 
     server->RegisterServlet("/js/yql-functions.js", new TYqlFunctoinsServlet());
     server->RegisterServlet("/js/sql-tokens.js", new TSqlTokensServlet());
+    server->RegisterServlet("/api/sql/complete", new TSqlCompleteServlet());
```
vityaman:
```suggestion
    server->RegisterServlet("/api/sql/completion", new TSqlCompleteServlet());
```

yql/tools/yqlrun/http/sql_complete_servlet.cpp
```diff
@@ -0,0 +1,284 @@
...
+            env.Parameters[name] = value["Data"];
+        } else {
+            env.Parameters[name] = value;
+        }
+    }
+
+    return env;
+} catch (...) {
+    return {};
+}
```
vityaman: Need to log an error too. The perfect solution could be to send it over http back to Web UI, but it seems not needed for an internal devtool.

yql/tools/yqlrun/http/sql_complete_servlet.cpp
```diff
@@ -0,0 +1,284 @@
...
+    TEnvironment env;
+
+    if (parameters.empty()) {
+        return env;
+    }
+
+    const auto root = NYT::NodeFromYsonString(parameters);
+    if (!root.IsMap()) {
+        return env;
+    }
```
vityaman, line 91: I think it is better to let it throw, it will be caught anyway.

vityaman (changes requested):
Thank you. It is a partial review. I need to test it locally and review a frontend better. Then I will return back.

Sorry for comments. I do not want to be too pedantic, but just trying to keep the code clean, when I can, because otherwise working on it can be frustrating.

yql/tools/yqlrun/http/sql_complete_servlet.cpp
Tony-Romanov: I don't think it's worth it. The code now looks simple and concise, and that structure has no chance of being reused elsewhere.

vityaman:
I did not forget, just need more time for a review.

nepal (approved):

vityaman:
Looks good. With some small sharp moments, but for an internal developer tool is not critical. Now the UX is much more better in yqlrun and thank you for this!

vityaman:

Issue #853, opened 2024-09-16
Add YT_OPERATION_TRANSACTION_ID to job environment


gritukan:
I have an operation that acquires a lock during start to ensure that no two operations are running concurrently. After the lock is acquired, transaction that was used for locking is used as a prerequisite.

Now I have to start such a per-operation transaction manually, however there are already operation-specific transactions that are managed by scheduler. We can pass id of such transaction to job environment and use it.

PR #1778, opened 2026-07-09
Add operation transaction id env

crystarm:
Related to #853.

The scheduler already maintains operation-scoped transactions, but their IDs weren't exposed to jobs. This made it impossible to use them as prerequisites when acquiring Cypress locks from within a job.

`YT_OPERATION_TRANSACTION_ID` is now set in every user job's environment. It contains the `output_transaction_id` for operations with output tables, or the `async_scheduler_transaction_id` otherwise. The variable is omitted if neither transaction exists.

The variable is set in `TOperationControllerBase::InitUserJobSpecTemplate` and is identical for all jobs within one operation.


coteeq:
Hi.

> This variable can be used, for example, as a prerequisite when acquiring Cypress locks from within a job to ensure mutual exclusion across concurrent operations.

What's wrong with using user transaction for this purpose?

I'm struggling to understand what useful guarantees do these transactions have (and user tx does not).

crystarm:
@coteeq 
Hi!! (˵ •̀ ᴗ - ˵ ) ✧
The use case from #853: the user may not have a user transaction at all. `YT_TRANSACTION_ID` might simply be absent, so they'd have to create and manage a separate transaction manually just to get operation-lifetime semantics. Even though the scheduler already has one.

crystarm:
Do not mind the force push. This PR got a commit from another branch, since I started the branch for this PR from the branch for another PR. I've fixed everything now.

PR #1762, opened 2026-06-23
Preparing for LLVM18. The first step.

Tony-Romanov:
Type: refactoring
Component: query-tracker

PR #1781, opened 2026-07-10
Set explicit allign for i128 in YQL IR generation.

Tony-Romanov:
Currently, when generating IR in YQL for the i128 type, alignment 8 is used,
and this is exactly what is required for correct generation. 

But this is achieved thanks to this code [here](https://github.com/ytsaurus/ytsaurus/blob/main/contrib/libs/llvm16/lib/IR/DataLayout.cpp#L637-L641).
It's better to directly set the explicit layout for the i128 type.

Type: fix
Component: query-tracker

### DIFF
--- a/yql/essentials/minikql/codegen/codegen.cpp
+++ b/yql/essentials/minikql/codegen/codegen.cpp
@@ -667,6 +667,7 @@ public:
             error.print("error after ParseIR()", os);
             ythrow yexception() << what;
         }
+
         module->setTargetTriple(Triple_);
         module->setDataLayout(Engine_->getDataLayout().getStringRepresentation());
         if (uniqId) {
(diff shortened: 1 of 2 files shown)


ilyaibraev:
assigning @igormunkin

nepal (approved):

igormunkin:
@Tony-Romanov, looks like this is not patch doesn't follow the plan we discussed [here](https://github.com/ytsaurus/ytsaurus/pull/1762#issuecomment-4902979205). AFAIU, we agreed to adjust `TUnboxedValue`/`TUnboxedValuePod` usage in LLVM part instead of alignment tweaks.

Otherwise, I might be missing something in this patchset.

Tony-Romanov:
> looks like this is not patch doesn't follow the plan we discussed [here](https://github.com/ytsaurus/ytsaurus/pull/1762#issuecomment-4902979205). AFAIU, we agreed to adjust `TUnboxedValue`/`TUnboxedValuePod` usage in LLVM part instead of alignment tweaks.

@igormunkin 
No, this doesn't prevent you from refactoring the use of TUnboxedValuePod in LLVM if you want to. 

It's just that instead of implicitly inheriting the alignment value from i64, the explicit setting of the actual value is used.

This won't change anything in the actual code generation, but it will make the source of the alignment value clearer without looking into the LLVM library code.

PR #1788, opened 2026-07-27
Log welcome debug info in Python SDK

### LINKED ISSUE #1443: logging. Log welcome debug info in Python SDK
- The problem:
  The user's debug logs do not contain information about the client version, python, bindings, which can be extremely useful for debugging.
- TODO:
  When using (not importing) some method from the YtClient client for the first time, write a message to the debug log once (same data as in `yt --version` outputs).
  - package version
  - version and path to python
  - availability of bindings (yson, rpc)
- Where - TBD

a-pap:
Closes #1443.

**Motivation.** Users' debug logs carry no information about the client version, python, or bindings availability, which makes debugging their environments harder.

**What changed.** On the first actual command going through `yt.wrapper` (top of `driver.make_request` — the common funnel for the http, native/rpc and batch backends, covering both module-level functions and `YtClient` methods), the SDK logs a single DEBUG line with the same data as `yt --version`:

```
YT wrapper version: 0.13.30; python: 3.11.9 (/usr/bin/python3); yson bindings: 0.4.10; rpc bindings: found
```

- once per process: module-level flag + lock, emitted via the existing `yt.logger.log_once`;
- nothing is logged at import time; steady-state overhead is one boolean check per request;
- rpc bindings availability is detected with `importlib.util.find_spec`, so the heavy bindings module is never imported by this code;
- the whole path is wrapped in `try/except`, so a failure of welcome logging can never break a client call.

**How verified.**
- new no-cluster unit test `tests/serverless/test_misc.py::test_welcome_debug_info` (follows the `test_log_once` pattern);
- standalone check with a DEBUG capture handler and two failing requests against a fake proxy: the welcome line appears exactly once, before the first HTTP attempt; also exercised with yson bindings installed and with an rpc-bindings stub that raises on import (reported "found" while the stub was not imported);
- 16 threads racing into the helper produce exactly one record;
- `flake8 --max-line-length=100` shows no new findings vs baseline.

This change was developed with AI assistance (Claude Code); I reviewed and verified all of it locally.

* Changelog entry
Type: feature
Component: python-sdk

Python SDK now logs a one-time debug message on the first executed command with the client version, python version and executable path, and yson/rpc bindings availability (the same data as `yt --version`) to simplify debugging from user debug logs.

### DIFF
--- a/yt/python/yt/wrapper/driver.py
+++ b/yt/python/yt/wrapper/driver.py
@@ -2,7 +2,7 @@ from . import http_driver
 
 from . import native_driver
 from .batch_response import apply_function_to_result
-from .common import YtError, update, simplify_structure, set_param, hide_secure_vault
+from .common import YtError, update, simplify_structure, set_param, hide_secure_vault, get_version
 from .config import get_option, set_option, get_config, get_backend_type
 from .format import create_format, JsonFormat, YsonFormat, YtFormatError
 from .http_helpers import get_http_api_version, get_http_api_commands
@@ -14,7 +14,10 @@ import yt.yson as yson
 
 from copy import deepcopy
 
+import importlib.util
+import logging
 import os
+import sys
 import threading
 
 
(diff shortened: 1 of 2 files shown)


denvr (approved):
LGTM

Issue #1776, opened 2026-07-08
Improve YQL completion engine


vityaman:
There was a discussion in the PR #1772 comments. @Tony-Romanov came and asked to improve YQL completion in YTsaurus Web UI. While now it uses `gravity-ui/websql-autocomplete` to provide `textDocument/completion`, in future it is possible to replace this library with in-house better completion engine at `yql/essentials/sql/v1/ide/completion`. It was noted, that it has some advantages, but there is still a room for an improvement. @Tony-Romanov integrated a new completion engine to the YQLRun Web UI, so it can be tested comfortably now.

In this issue I propose to have a discussion about current completion engine. You can send query snippets with `#` as a cursor position, expected and actual completion candidates.

Clone: `YTSAURUSSUP-2844`

vityaman:
From https://github.com/ytsaurus/ytsaurus/pull/1772#issuecomment-4901512175:

> The problem is that the lack of a predefined column separator ',' or between explicit expressions ';' almost fully breaks autocompletion. With library lookup points, it's clear—invalid SQL. But a human normally enters, for example, the name of a column or function first, and only then a comma. I'm sure this can be improved.

So a better error recovery is needed.

PR #1792, opened 2026-07-28
Implement streaming mode.

Tony-Romanov:
Type: feature
Component: query-tracker

### DIFF
--- a/yql/essentials/tools/yql_complete/yql_complete.cpp
+++ b/yql/essentials/tools/yql_complete/yql_complete.cpp
@@ -235,7 +235,7 @@ void LogStreamError(const TString& message) {
 }
 
 void RunStream(const TCompletionResources& resources) {
-    while (true) {
+    for (;;) {
         TString document;
         const auto readResult = ReadJsonDocument(Cin, document);
         if (readResult == EReadDocumentResult::EndOfStream) {
--- a/yql/essentials/tools/yql_complete/yql_complete.cpp
+++ b/yql/essentials/tools/yql_complete/yql_complete.cpp
@@ -230,7 +230,7 @@ void WriteEmptyStreamResponse() {
     WriteStreamResponse({});
 }
 
-void LogStreamError(const TString& message) {
+void LogStreamError(TStringBuf message) {
     Cerr << "Failed to process stream request: " << message << Endl;
 }
 
--- a/yql/essentials/tools/yql_complete/yql_complete.cpp
+++ b/yql/essentials/tools/yql_complete/yql_complete.cpp
@@ -157,8 +157,8 @@ EReadDocumentResult ReadJsonDocument(IInputStream& input, TString& document) {
         return EReadDocumentResult::Complete;
     }
 
-    bool inString = first == '"';
-    bool escaped = false;
+    bool isInString = (first == '"');
+    bool isEscaped = false;
     TVector<char> closingCharacters;
     if (first == '{') {
         closingCharacters.push_back('}');
--- a/yql/essentials/tools/yql_complete/yql_complete.cpp
+++ b/yql/essentials/tools/yql_complete/yql_complete.cpp
@@ -15,11 +15,16 @@
 
 #include <library/cpp/getopt/last_getopt.h>
 #include <library/cpp/json/json_reader.h>
+#include <library/cpp/json/json_writer.h>
 #include <library/cpp/iterator/iterate_keys.h>
 #include <library/cpp/iterator/functools.h>
 
+#include <util/generic/vector.h>
 #include <util/charset/utf8.h>
 #include <util/stream/file.h>
+#include <util/stream/str.h>
+
+#include <cctype>
 
 NSQLComplete::TFrequencyData LoadFrequencyDataFromFile(TString filepath) {
     TString text = TUnbufferedFileInput(filepath).ReadAll();
@@ -61,7 +66,7 @@ NSQLComplete::TCompletionInput MakeCompletionInput(TString& text, TMaybe<ui64> p
         return NSQLComplete::SharpedInput(text);
     }
 
-    if (lengthUtf8 <= *pos) {
+    if (lengthUtf8 < *pos) {
         ythrow yexception() << "provided position " << *pos << " is out of range " << lengthUtf8;
     }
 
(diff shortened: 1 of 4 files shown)


vityaman:
Hello! Interesting feature. Why is it needed?

vityaman:
Streaming JSON parsing seems to be a useful feature and maybe deserves to be placed at `yql/essentials/utils/json`. 

Not far from today, I actually needed it, because I wanted to process JSON Lines (.jsonl) like input stream, but with items formatted, not only on a single line. I needed it to make a testing framework based on protocol messages traces, so a server streams these traces somewhere and then this traces can be replayed. At the end I just decided to write JSON array to stream and then to cope with possibly invalid (not closed) JSON (after a server crash, for example).

vityaman:
Also I can't accept it without a great amount of unit tests.

yql/essentials/tools/yql_complete/yql_complete.cpp
```diff
@@ -71,6 +76,221 @@ NSQLComplete::TCompletionInput MakeCompletionInput(TString& text, TMaybe<ui64> p
     };
 }
 
+class TCompletionResources {
```
vityaman: I propose to move this to a separate translation unit. Let's normalize a domain model a little, bot putting everything in a one file (somewhy people in YQL like doing that and it is not really maintainable). Let's try to keep to Single Responsibility Principle.

yql/essentials/tools/yql_complete/yql_complete.cpp
```diff
@@ -71,6 +76,221 @@ NSQLComplete::TCompletionInput MakeCompletionInput(TString& text, TMaybe<ui64> p
...
+            if (std::isspace(static_cast<unsigned char>(current))) {
+                break;
+            }
+            document.push_back(current);
+        }
+        return EReadDocumentResult::Complete;
+    }
+
+    bool inString = first == '"';
+    bool escaped = false;
```
vityaman:
```suggestion
    bool isInString = (first == '"');
    bool isEscaped = false;
```

yql/essentials/tools/yql_complete/yql_complete.cpp
```diff
@@ -71,6 +76,221 @@ NSQLComplete::TCompletionInput MakeCompletionInput(TString& text, TMaybe<ui64> p
...
+    Complete,
+    EndOfStream,
+    Incomplete,
+};
+
+EReadDocumentResult ReadJsonDocument(IInputStream& input, TString& document) {
+    document.clear();
+
+    char current;
+    while (input.ReadChar(current)) {
```
vityaman:
```suggestion
    for (char current; input.ReadChar(current);) {
```

yql/essentials/tools/yql_complete/yql_complete.cpp
```diff
@@ -71,6 +76,221 @@ NSQLComplete::TCompletionInput MakeCompletionInput(TString& text, TMaybe<ui64> p
...
+            if (!closingCharacters.empty() && closingCharacters.back() == current) {
+                closingCharacters.pop_back();
+                if (closingCharacters.empty()) {
+                    return EReadDocumentResult::Complete;
+                }
+            } else if (!closingCharacters.empty() && closingCharacters.front() == current) {
+                // Let the JSON parser report mismatched nested brackets, but
+                // terminate the malformed top-level document for recovery.
+                return EReadDocumentResult::Complete;
+            }
```
vityaman, line 131: I think the recovery should be tested as well, because now it is unclear to me, what is expected recovery behavior.

yql/essentials/tools/yql_complete/yql_complete.cpp
```diff
@@ -71,6 +76,221 @@ NSQLComplete::TCompletionInput MakeCompletionInput(TString& text, TMaybe<ui64> p
...
+        }
+    }
+
+    return EReadDocumentResult::Incomplete;
+}
+
+TString CandidateKindToString(NSQLComplete::ECandidateKind kind) {
```
vityaman: It seems to me that just `ToString` should work, because it should be derived from an `Out` implementation.

yql/essentials/tools/yql_complete/yql_complete.cpp
```diff
@@ -71,6 +76,221 @@ NSQLComplete::TCompletionInput MakeCompletionInput(TString& text, TMaybe<ui64> p
...
+    NJson::WriteJson(&Cout, &response, false);
+    Cout << Endl;
+    Cout.Flush();
+}
+
+void WriteEmptyStreamResponse() {
+    WriteStreamResponse({});
+}
+
+void LogStreamError(const TString& message) {
```
vityaman:
```suggestion
void LogStreamError(TStringBuf message) {
```

yql/essentials/tools/yql_complete/yql_complete.cpp
```diff
@@ -71,6 +76,221 @@ NSQLComplete::TCompletionInput MakeCompletionInput(TString& text, TMaybe<ui64> p
...
+    for (const auto& candidate : candidates) {
+        response.AppendValue(NJson::TJsonMap{
+            {"word", candidate.Content},
+            {"type", CandidateKindToString(candidate.Kind)},
+        });
+    }
+
+    NJson::WriteJson(&Cout, &response, false);
+    Cout << Endl;
+    Cout.Flush();
```
vityaman: `Endl` should flush, doesn't it?

```suggestion
    Cout << Endl;
```

yql/essentials/tools/yql_complete/yql_complete.cpp
```diff
@@ -71,6 +76,221 @@ NSQLComplete::TCompletionInput MakeCompletionInput(TString& text, TMaybe<ui64> p
...
+            auto input = MakeCompletionInput(query, position);
+            auto output = engine->CompleteAsync(input).ExtractValueSync();
+            WriteStreamResponse(output.Candidates);
+        } catch (const std::exception& error) {
+            LogStreamError(error.what());
+            WriteEmptyStreamResponse();
+        } catch (...) {
+            LogStreamError(CurrentExceptionMessage());
+            WriteEmptyStreamResponse();
+        }
```
vityaman: Why can't we let an internal dev-only tool to crash on such an unexpected error? It should be caught at `Main` once and terminate input stream reading.

yql/essentials/tools/yql_complete/yql_complete.cpp
```diff
@@ -71,6 +76,221 @@ NSQLComplete::TCompletionInput MakeCompletionInput(TString& text, TMaybe<ui64> p
...
+                if (!request["position"].IsUInteger()) {
+                    ythrow yexception() << "position must be a non-negative integer";
+                }
+                position = Min<ui64>(request["position"].GetUIntegerSafe(), queryLength);
+            }
+
+            const NJson::TJsonValue* schema = nullptr;
+            if (request.Has("schema")) {
+                schema = &request["schema"];
+            }
```
vityaman: Parsing a request should be extracted to a separate function. SRP.

yql/essentials/tools/yql_complete/yql_complete.cpp
```diff
@@ -71,6 +76,221 @@ NSQLComplete::TCompletionInput MakeCompletionInput(TString& text, TMaybe<ui64> p
...
+void WriteEmptyStreamResponse() {
+    WriteStreamResponse({});
+}
+
+void LogStreamError(const TString& message) {
+    Cerr << "Failed to process stream request: " << message << Endl;
+}
+
+void RunStream(const TCompletionResources& resources) {
+    while (true) {
```
vityaman: That is more idiomatic in C++.

```suggestion
    for (;;) {
```

yql/essentials/tools/yql_complete/yql_complete.cpp
vityaman: I think a Factory is a better name. It is kinda unexpected, that just resources can produce a completion engine.

vityaman (changes requested):
Besides stylistic comments, I have main 2 points to say:

1. Why this feature is needed?

2. Streaming JSON parsing is cool, should be in `utils`, well tested, reliable.

vityaman:
Let's link a PR to the issue https://github.com/ytsaurus/ytsaurus/issues/1776 to keep our history.

vityaman:
It seems to me that we are building the same thing in the same codebase with the same goals, but do not share our intentions and so can't help each other in an optimal way.

I am ready to be open to you (as my company lets me :>), let's speak a little bit more. For example, you can express your wishes and plans in the issue comments as well as describing a PRs including information, why the patch is needed, part of which project is it, what is done, why is it done they way it is. 

I appreciate your work and glad to receive patches from you, but I still otherwise would anyway ask you for it in comments. This is because accepting a code implies taking responsibilities of it, and one of responsibilities is keeping a component minimal and elegant.

Tony-Romanov:
> * Why this feature is needed?

I want to use the yql autocompletion library in our QT UI.

And I already have a working test cluster with a UI code modification that can optionally use this utility instead of the native UI implementation. While it's far from ideal, it already makes autocompletion much better.

And the most important point: I want to see this functionality on our production clusters in a month, or a month and a half at most.
Then, when you come up with and implement a dedicated, reliable, fault-tolerant, high-performance, beautiful, magnificent, etc. yql autocompletion service, we'll take a look and most likely switch to it with pleasure.
But I need a solution right now.

I already have the solution. I don't need to commit anything to upstream; we can keep all the modifications in our repository. But since I see a desire to collaborate on your part, especially in contrast to the rest of the YQL team, for which I'm especially grateful, I'm trying to introduce them.

And a week and a half ago, on a sync with the right YT/QT team, I asked for a meeting to discuss plans about autocompletion, and so far, nothing.
That's about the timeline.

> * Streaming JSON parsing is cool, should be in `utils`, well tested, reliable.

That's great, but you can get it there yourself. 
For us, committing anything to the shared library is a huge pain.

vityaman:
@Tony-Romanov, 

> And I already have a working test cluster with a UI code modification that can optionally use this utility instead of the native UI implementation. While it's far from ideal, it already makes autocompletion much better.

Does the Web UI call this utility locally? Do you invoke this utility from a web-browser as a subprocess? How it works?

I imagined the YQL autocompletion implementation in YT Web UI as a HTTP endpoint on the server-side, maybe at the query tracker, I do not know (this is how it approximately works in our internal service).

vityaman:
About autocompletion plans. 

Now I am working on an open source LSP-server for YQL with an autocompletion, of course, included. Not sure that an autocompletion will be improved during the development, because  would like to focus on new features such as code navigation (rename, go to definition, find refences, symbols). But of course, old features such as documentation on hover, autocompletion, partial typecheck will be there too.

So in case you are doing an autocompletion integration into the YT Web UI via some local tool, I suggest to make its API as close as possible to the LSP protocol, so the future LSP server could be a replacement.

It seems not about YT Web UI, because the server will be integrated to the VSCode to begin with. I do not notice a desire of IDE-features integration from the YT team, but I see it as a set of HTTP API endpoints with a schema very close to the LSP, except that methods receives a full query text rather than a URI to keep the protocol stateless.

Tony-Romanov:
> Does the Web UI call this utility locally? Do you invoke this utility from a web-browser as a subprocess? How it works?

On UI server, implemented in Node.js, the call to the native JavaScript autocompletion implementation is replaced with a call to a local microservice that includes this utility and FIFO queue. This is implemented as simply as possible, as an add-on, and does not offer any further extensions.

> I imagined the YQL autocompletion implementation in YT Web UI as a HTTP endpoint on the server-side, maybe at the query tracker, I do not know (this is how it approximately works in our internal service).

You imagined everything correctly, and so did I. That's what needs to be done, but later. 
This solution definitely won't appear in a month.

vityaman:
@Tony-Romanov, it was decided that we do not need this feature in the `yql/essentials/tools/yql_complete`. I am sorry.

I also want to say, that I do not fully understand the approach you are taking with this tool and `Node.js`. As far I as I know, YTsaurus has a service called Query Tracker. I guess there should be a HTTP API to get some information for the UI, for example, cypress nodes, operation status and so on. 

Why not just adding an other fresh `POST /ide/v0/completion` endpoint there? It should be written in C++ and so an autocompletion engine should be easily pluggable. It seems to me that the patch there will be accepted more likely, that to this tool source code. 

Having a code there should be more comfortable for you and useful not only for you, but other YTsaurus clients. Even if you'll get rejected (e.g. because of not production readiness), you anyway have your own fork. 

So I suggest to contact YTsaurus backend maintainers. We in YQL mostly develop libraries, but not UI integrations (IDEs such as VSCode is an exception).

Tony-Romanov:
> it was decided that we do not need this feature in the `yql/essentials/tools/yql_complete`. I am sorry.

Looks like I was wrong and collaboration won't work out.

Even though, we're just adding an optional feature to a utility you don't use. It costs you nothing. It was basically broken until I fixed it.

Too bad.

vityaman:
@Tony-Romanov, you have found a bug in this utility and it was fixed, and we are thankful for it. But exactly this patch introduces functionality only you need (it is a deadcode for us), because we use it to quick check an autocompletion locally. The streaming is definitely not needed. A query input in a JSON is also is not needed, because it is not readable by a human (escape sequences). 

I don't get your disappointment, because I don't understand, what is the problem of having a separate build target (application) in your fork — it seems to have a low cost for you, not really higher than patching our dev-only tool. 

From my point of you, you are trying to push your application code into the YQL infrastructure (library) codebase, that nobody except you will use, so I can't accept it, because the YQL is a generic framework.

You are welcome with high quality patches on an infrastructure level or a long-living application useful for a YTsaurus users or development team. For example, we merged an autocompletion in YQLRun Web UI and with pleasure tried it.

I understand your hurry and a current approach on autocompletion development. I want to say, that even in case you can't provide an acceptable by us solution, you can provide us with ideas, issues, features and so in the issue we agreed to communicate in. You even can use it as notes, to make it not effort-consuming for you. I will definitely look at them and give a feedback. I am not sure, that I can improve an autocompletion with in short time, but maybe in 2-4 weeks (not an SLA, but I hope to manage it). For example, I found, that you have a branch with some syntax analysis recovery in a SELECT projection, that could share with me explicitly in that issue, so I could know about that problem existence.

Tony-Romanov:
> But exactly this patch introduces functionality only you need (it is a deadcode for us), because we use it to quick check an autocompletion locally.

The core concept of "collaboration" is not just thinking about your own needs, but also considering the interests of those you collaborate with.
And, at the very least, not prohibiting something that costs you nothing.
All that was required was for the utility to be in the shared codebase and compile, and this costs you nothing and happens automatically.

vityaman:

PR #1798, opened 2026-08-10
Fix response format in exists command documentation

faic:


panesher:
I think we're not returning dict in exists command:
```bash
$ yt exists //sys
true
```

Can you provide your example so I can take a look on it?

faic:
```
 ~ % curl --silent --show-error --get \
  "${YT_URL}/api/v4/exists" \
  --header "Authorization: OAuth ${YT_TOKEN}" \
  --data-urlencode "path=${YT_PATH}" |
jq .
{
  "value": false
}
```
It seems for v4 version it started returning dict https://github.com/ytsaurus/ytsaurus/blob/86bee79db92507b741b56e9039386533092d2767/yt/yt/client/driver/cypress_commands.cpp#L652-L664 , https://github.com/ytsaurus/ytsaurus/blob/86bee79db92507b741b56e9039386533092d2767/yt/yt/client/driver/command.h#L53-L67

panesher:
Can you add that in v3 it would return string value and in v4 map value?

Thanks

faic:
> Can you add that in v3 it would return string value and in v4 map value?
> 
> Thanks

done

panesher:
LGTM

PR #1801, opened 2026-08-12
Attach table from S3 storages

pavel-bash:
## Summary

Adds initial support for attaching S3-hosted Parquet files as YT tables via `attach_table`:
- Adds native/RPC APIs, Python SDK, and CLI commands.
- Generates YT chunk metadata from Parquet row groups using Arrow, and reads attached chunks through the offshore data gateway.
- Ports required master/replica source-URI handling for externally attached chunks.
- Adds `TestAttachTable` S3 integration coverage.

Main changes:
- `yt/yt/ytlib/table_client/attach_table.cpp`: attach workflow, validation, and master confirmation.
- `yt/yt/ytlib/chunk_client/external_parquet.*` and Arrow reader changes: Parquet-to-chunk metadata and row-group decoding.
- Master/ODG/replica changes: propagation of external S3 source URIs.
- Public API/CLI/Python bindings and integration tests.

For the simplicity of the review process this PR is split into 3 commits:
1. Adds source URI and format metadata to offshore chunk replicas.
2. Adds Arrow/Parquet parsing and serves virtual external chunks through the gateway.
3. Exposes the command through native/RPC/driver/Python CLI and adds Parquet S3 integration coverage.

### DIFF
--- a/yt/yt/client/chunk_client/chunk_replica.h
+++ b/yt/yt/client/chunk_client/chunk_replica.h
@@ -11,6 +11,8 @@
 
 #include <yt/yt/core/misc/protobuf_helpers.h>
 
+#include <string>
+
 namespace NYT::NChunkClient {
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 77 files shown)


yt/yt/client/chunk_client/chunk_replica.h
koct9i, line 66: This grows size of TChunkReplicaWithMedium from 8 to 32 bytes. LOL.

va-kuznecov: and that is important because on large clusters with high chunk count per data node ally replica announcement consumes x4 amount of memory during startup. That can cause OOM or other unexpected problems.

PR #1806, opened 2026-08-19
Wait for open sessions futures in case prefetching enabled

sagishev:
* Changelog entry
Type: fix
Component: dynamic-tables

Fix a race condition that occurs when `partition_reader_prefetch_key_limit` is enabled. Previously, futures created in `MaybePrefetchPartitionSessions` were not awaited, resulting in tablet-node crashes.
Bug can be reproduced consistently on a dynamic table with 100 chunks and the following @mount_config
```
  {
      "enable_data_node_lookup" = %true;
      "enable_key_filter_for_lookup" = %true;
      "partition_reader_prefetch_key_limit" = 50;
  }
```

### DIFF
--- a/yt/yt/server/node/tablet_node/lookup.cpp
+++ b/yt/yt/server/node/tablet_node/lookup.cpp
@@ -1165,6 +1165,8 @@ struct TPartitionSession
     // TODO(akozhikhov): Proper block fetcher: Create all partition sessions at the beginning of the lookup session.
     // Right know we cannot do that because chunk reader may call Open in ctor and start reading blocks.
     bool SessionStarted = false;
+    // Prevents reading the partition until all its store sessions are open.
+    TFuture<void> OpenStoreSessionsFuture = VoidFuture;
 
     TStoreSessionList StoreSessions;
     bool StoreSessionsPrepared = false;
@@ -1362,6 +1364,7 @@ private:
         const TSharedRange<TLegacyKey>& keys);
 
     std::vector<TFuture<void>> OpenStoreSessions(const TStoreSessionList& sessions);
+    void StartPartitionSession(TPartitionSession* partitionSession);
 
     void LookupInPartitions(const TError& error);
 
@@ -2089,11 +2092,10 @@ auto TTabletLookupSession<TPipeline>::Run() -> TFuture<typename decltype(TPipeli
     openStoreSessions(ChunkEdenSessions_);
 
     YT_VERIFY(!PartitionSessions_.empty());
-    PartitionSessions_[0].SessionStarted = true;
-    PartitionSessions_[0].StoreSessions = CreateStoreSessions(
-        PartitionSessions_[0].PartitionSnapshot->Stores,
-        PartitionSessions_[0].ChunkLookupKeys);
-    openStoreSessions(PartitionSessions_[0].StoreSessions);
+    StartPartitionSession(&PartitionSessions_[0]);
+    if (PartitionSessions_[0].OpenStoreSessionsFuture != VoidFuture) {
+        openFutures.push_back(PartitionSessions_[0].OpenStoreSessionsFuture);
+    }
 
     InflightKeyLookupCount_ += std::ssize(PartitionSessions_[0].ChunkLookupKeys);
     ++OpenedPartitionSessionCount_;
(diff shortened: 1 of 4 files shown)


yt/yt/server/node/tablet_node/lookup.cpp
```diff
@@ -2089,11 +2092,10 @@ auto TTabletLookupSession<TPipeline>::Run() -> TFuture<typename decltype(TPipeli
     openStoreSessions(ChunkEdenSessions_);
 
     YT_VERIFY(!PartitionSessions_.empty());
-    PartitionSessions_[0].SessionStarted = true;
-    PartitionSessions_[0].StoreSessions = CreateStoreSessions(
-        PartitionSessions_[0].PartitionSnapshot->Stores,
-        PartitionSessions_[0].ChunkLookupKeys);
-    openStoreSessions(PartitionSessions_[0].StoreSessions);
+    StartPartitionSession(&PartitionSessions_[0]);
+    if (PartitionSessions_[0].OpenStoreSessionsFuture != VoidFuture) {
```
Quid37: In such checks we omit explicitly writing `!= VoidFuture`

yt/yt/server/node/tablet_node/lookup.cpp
```diff
@@ -2173,6 +2175,21 @@ std::vector<TFuture<void>> TTabletLookupSession<TPipeline>::OpenStoreSessions(
...
+void TTabletLookupSession<TPipeline>::StartPartitionSession(TPartitionSession* partitionSession)
+{
+    YT_VERIFY(!std::exchange(partitionSession->SessionStarted, true));
+
+    partitionSession->StoreSessions = CreateStoreSessions(
+        partitionSession->PartitionSnapshot->Stores,
+        partitionSession->ChunkLookupKeys);
+
+    auto openFutures = OpenStoreSessions(partitionSession->StoreSessions);
+    partitionSession->OpenStoreSessionsFuture = openFutures.empty()
```
Quid37: Let's not set anything in empty case

yt/yt/server/node/tablet_node/lookup.cpp
```diff
@@ -1167,6 +1167,8 @@ struct TPartitionSession
     // TODO(akozhikhov): Proper block fetcher: Create all partition sessions at the beginning of the lookup session.
     // Right know we cannot do that because chunk reader may call Open in ctor and start reading blocks.
     bool SessionStarted = false;
+    // Prevents reading the partition until all its store sessions are open.
+    TFuture<void> OpenStoreSessionsFuture = VoidFuture;
```
Quid37: That's called OKFuture now.
However I would not set anything so it is null future, so we can later check it was actually set.

yt/yt/server/node/tablet_node/lookup.cpp
Quid37: OKFuture

yt/yt/server/node/tablet_node/lookup.cpp
```diff
@@ -2308,24 +2321,22 @@ bool TTabletLookupSession<TPipeline>::LookupInCurrentPartition()
...
-        if (!openFutures.empty()) {
-            auto sessionFuture = AllSucceeded(std::move(openFutures));
-            sessionFuture.Subscribe(BIND(
-                &TTabletLookupSession::LookupInPartitions,
-                MakeStrong(this))
-                .Via(Invoker_));
-            SetSessionFuture(std::move(sessionFuture));
-            return true;
-        }
+    if (const auto& maybeError = partitionSession.OpenStoreSessionsFuture.TryGet()) {
```
Quid37, line 2330: Let's add a verify that OpenStoreSessionsFuture is not null here

Quid37 (changes requested):
Can you also add an integration test (to test_lookup.py) that would crash without the fix?

Krock21:
We deployed the fix to our cluster and enabled the feature. It seems to be working, so this is the last bug to fix

Quid37:
Great! Can you please fix the issues so we can merge it

sagishev:
@Quid37 sorry for the delay, updated.
Can you please take a look? I've added some python test but it was required to tweak the config settings a little to make bug reproducible.

Quid37:
Thanks for the PR!

Quid37:
Shipped!

Quid37:
Shipped!

PR #1817, opened 2026-08-28
blockcodecs: switch the snappy codec to klauspost s2

orloffv:
s2.EncodeSnappy/Decode emit and read the standard snappy block format (same codec ID 50986, interoperable with the reference implementation) while encoding ~15-50% faster with slightly better ratio, depending on the platform. klauspost/compress is already a dependency via blockzstd.

The new test round-trips the codec against github.com/golang/snappy in both directions.


tinarsky:
Hello, thanks for the patch! Unfortunately we aren't able to accept external contributions to `library/go/` at the moment, so we can't merge this.

The change looks reasonable, though — we'll consider doing it on our side, with your PR as the reference. No timeline to promise at this point

Issue #1812, opened 2026-08-24
Improve downloads of truncated query results


ulya-sidorina:
### Problem description

Query Tracker stores only a preview of each query result, limited by row count (10k by default) and serialized size (~16 MB).
The UI download action exports this stored preview. Therefore, users may download an incomplete result while expecting the full query output.

### Expected behavior

Query Tracker already returns `is_truncated` flag and `full_result` metadata, but not all engines provide this info:
Engine |	`full_result` |	`is_truncated`
-- | -- | --
YQL	| Yes | Yes
CHYT | No | No
SPYT | No | Yes
YT QL | No | No
 
The download behavior should depend on `is_truncated` and `full_result`:
`is_truncated` | `full_result` | Behavior
-- | -- | --
true | available | Download from the full-result table
true | unavailable | Warn that the downloaded result is truncated
false | any | Keep the current behavior

### Required changes

- Provide `is_truncated` for CHYT and YT QL 
- Download from full result table (if available)
- Warn users in case of downloading truncated result (for ex. rename `Download` button to `Download truncated result`)

PR #1818, opened 2026-08-28
ISSUE-1812: add is_truncated support for CHYT query result

ulya-sidorina:
* Changelog entry
Type: feature
Component: chyt, query-tracker

Add `is_truncated` support for CHYT query results.

### DIFF
--- a/yt/chyt/client/protos/query_service.proto
+++ b/yt/chyt/client/protos/query_service.proto
@@ -25,6 +25,7 @@ message TRspExecuteQuery
 {
     optional NYT.NProto.TGuid query_id = 1;
     optional NYT.NProto.TError error = 2;
+    repeated bool is_truncated = 3;
     // Attachments contain wire-encoded data.
 }
 
--- a/yt/chyt/tests/server/test_query_tracker.py
+++ b/yt/chyt/tests/server/test_query_tracker.py
@@ -491,6 +491,27 @@ class TestQueriesChyt(ClickHouseTestBase):
             assert all(row["type"] == "ExceptionWhileProcessing" for row in rows)
 
 
+class TestQueriesChytWithRowLimit(ClickHouseTestBase):
+    QUERY_TRACKER_DYNAMIC_CONFIG = {
+        "chyt_engine": {
+            "row_count_limit": 100,
+        },
+    }
+
+    @authors("ulya-sidorina")
+    def test_result_truncation(self, query_tracker):
+        with Clique(1, alias="*ch_alias"):
+            settings = {"clique": "ch_alias", "cluster": "primary"}
+
+            query = start_query("chyt", "select number from numbers(100)", settings=settings)
+            query.track()
+            assert not query.get_result(0)["is_truncated"]
+
+            query = start_query("chyt", "select number from numbers(101)", settings=settings)
+            query.track()
+            assert query.get_result(0)["is_truncated"]
+
+
 class TestChytEngineProgress(ClickHouseTestBase):
     def setup_method(self, method):
         super().setup_method(method)
--- a/yt/yt/server/query_tracker/chyt_engine.cpp
+++ b/yt/yt/server/query_tracker/chyt_engine.cpp
@@ -424,9 +424,13 @@ private:
 
         std::vector<TErrorOr<TWireRowset>> wireRowsetOrErrors;
         wireRowsetOrErrors.reserve(rsp->Attachments().size());
-        for (const auto& ref : rsp->Attachments()) {
+        for (int index = 0; index < std::ssize(rsp->Attachments()); ++index) {
+            const auto& ref = rsp->Attachments()[index];
             if (!ref.Empty()) {
-                wireRowsetOrErrors.emplace_back(TWireRowset{.Rowset = ref});
+                wireRowsetOrErrors.emplace_back(TWireRowset{
+                    .Rowset = ref,
+                    .IsTruncated = index < rsp->incomplete_size() && rsp->incomplete(index),
+                });
             }
         }
         OnQueryCompletedWire(wireRowsetOrErrors);
--- a/yt/chyt/server/query_service.cpp
+++ b/yt/chyt/server/query_service.cpp
@@ -66,6 +66,7 @@ struct TRowset
 {
     TSharedRef Rowset;
     i64 TotalRowCount;
+    bool IsTruncated;
 };
 
 template <class TRequest>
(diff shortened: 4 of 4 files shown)


koct9i:
Issue: #1812

yt/chyt/client/protos/query_service.proto
```diff
@@ -25,6 +25,7 @@ message TRspExecuteQuery
 {
     optional NYT.NProto.TGuid query_id = 1;
     optional NYT.NProto.TError error = 2;
+    repeated bool incomplete = 3;
```
buyval01: The final destination of this field is the `IsTruncated` struct field:
https://github.com/ytsaurus/ytsaurus/pull/1818/changes#diff-575366785d981bdf110737c4111e5bf147b69fbd6178e6025324c212503d6d12R69

Why not name it is_truncated here as well?

yt/chyt/tests/server/test_query_tracker.py
```diff
@@ -57,6 +57,19 @@ def test_simple_query_not_indexed(self, query_tracker):
...
+            settings = {"clique": "ch_alias", "cluster": "primary"}
+
+            query = start_query("chyt", "select number from numbers(10000)", settings=settings)
+            query.track()
+            assert not query.get_result(0)["is_truncated"]
+
+            query = start_query("chyt", "select number from numbers(10001)", settings=settings)
+            query.track()
+            assert query.get_result(0)["is_truncated"]
+
```
buyval01: Here, we implicitly rely on the config’s default value for `row_count_limit`.

Let’s make this behavior explicit by specifying it in the test suite:
```
class TestQueriesChytWithRowLimit(ClickHouseTestBase):
      QUERY_TRACKER_DYNAMIC_CONFIG = {
          "chyt_engine": {
              "row_count_limit": 10000,
          },
      }
```

yt/yt/server/query_tracker/chyt_engine.cpp
```diff
@@ -424,9 +424,13 @@ class TChytQueryHandler
...
         std::vector<TErrorOr<TWireRowset>> wireRowsetOrErrors;
         wireRowsetOrErrors.reserve(rsp->Attachments().size());
-        for (const auto& ref : rsp->Attachments()) {
+        for (int index = 0; index < std::ssize(rsp->Attachments()); ++index) {
+            const auto& ref = rsp->Attachments()[index];
             if (!ref.Empty()) {
-                wireRowsetOrErrors.emplace_back(TWireRowset{.Rowset = ref});
+                wireRowsetOrErrors.emplace_back(TWireRowset{
+                    .Rowset = ref,
+                    .IsTruncated = index < rsp->incomplete_size() && rsp->incomplete(index),
```
buyval01, line 432: I’m not sure that we won’t change the current default value of `IsTruncated` later (it is currently `false`). However, such a change would cause every result produced by an older CHYT version to be treated as truncated because the new field is absent from the proto message.

ulya-sidorina: Sorry, I don't get it, do you mean we should explicitly set `IsTruncated` to false when the proto field is absent? The current expression already does that

yt/chyt/tests/server/test_query_tracker.py
ulya-sidorina: Done

yt/chyt/client/protos/query_service.proto
ulya-sidorina: There's no reason, renamed to `is_truncated`

yt/yt/server/query_tracker/chyt_engine.cpp
buyval01: I meant that, from my POV, it's better to respect the defined default value of the struct here. But I don't have a strong opinion, so LGTM! :)

yt/chyt/tests/server/test_query_tracker.py
```diff
@@ -491,6 +491,33 @@ def match(row):
             assert all(row["type"] == "ExceptionWhileProcessing" for row in rows)
 
 
+class TestQueriesChytWithRowLimit(ClickHouseTestBase):
+    DELTA_LOCAL_YT_CONFIG = {
+        "address_resolver_config": {
+            "enable_ipv4": True,
+        },
+    }
```
buyval01, line 499: Why do we need it?

buyval01 (approved):

yt/chyt/tests/server/test_query_tracker.py
ulya-sidorina: it's needed for my local tests, removed

buyval01 (approved):

yt/yt/server/query_tracker/chyt_engine.cpp
Artyomkun: To follow up on @buyval01's point about the default value — let me try to make the concrete failure mode explicit.

Suppose later someone changes the proto default of `is_truncated` from `false` to `true` (or the struct default, or the reader default). What happens:

- Old CHYT versions don't send the field at all.
- The new Query Tracker reads a missing field and takes the **new default**.
- Every result from an old CHYT version is now marked as truncated.
- The UI shows "Download truncated result" (or blocks download), even though the result is complete.

That's a silent data-loss-shaped bug for users: they either get a warning that isn't true, or they can't get the data they used to get.

Two things that would make this robust regardless of the default:

1. **Initialize `IsTruncated` explicitly in the struct**

```cpp
struct TRowset
{
    TSharedRef Rowset;
    i64 TotalRowCount;
    bool IsTruncated = false;  // explicit, not UB, not proto-dependent
};
```

Right now `bool IsTruncated;` is uninitialized — in C++ that's UB, and it also makes the behavior depend on the proto default.

2. **Treat "field absent" as a distinct case from "field present and false"**

If you can distinguish "old CHYT, no field" from "new CHYT, not truncated", then a future default change can't retroactively mark old results as truncated. In proto3 this usually means `optional bool` (which you already have) plus a presence check:

```cpp
.IsTruncated = rsp->has_incomplete() && index < rsp->incomplete_size() && rsp->incomplete(index),
```

(or the equivalent with `optional` presence). Then missing = "unknown", not "truncated".

**Also worth deciding before merge:**

- `full_result` is not added in this PR. So even with `is_truncated = true`, the UI can't download the full result for CHYT — it can only warn. Is that acceptable for this iteration, or should `full_result` land together?
- `incomplete` vs `is_truncated` — two names for the same concept. Consistency would help whoever reads this next.

Not blocking — just making the tradeoff explicit so it's a deliberate choice, not an accident.

PR #1820, opened 2026-08-28
ISSUE-1812: add is_truncated support for ql engine

ulya-sidorina:
* Changelog entry
Type: feature
Component: query-tracker

Add `is_truncated` support for YT QL query results.

### DIFF
--- a/yt/yt/server/query_tracker/ql_engine.cpp
+++ b/yt/yt/server/query_tracker/ql_engine.cpp
@@ -9,6 +9,8 @@
 
 #include <yt/yt/ytlib/api/native/client.h>
 
+#include <yt/yt/client/tablet_client/public.h>
+
 #include <yt/yt/core/ytree/convert.h>
 #include <yt/yt/core/ytree/attributes.h>
 
@@ -37,12 +39,14 @@ public:
         const TEngineConfigBasePtr& config,
         const NRecords::TActiveQuery& activeQuery,
         const TSelectRowsOptions& options,
+        bool failOnIncompleteInput,
         const IClientPtr& queryClient,
         const IInvokerPtr& controlInvoker)
         : TQueryHandlerBase(stateClient, stateRoot, controlInvoker, config, activeQuery)
         , Query_(activeQuery.Query)
         , QueryClient_(queryClient)
         , Options_(options)
+        , FailOnIncompleteInput_(failOnIncompleteInput)
     { }
 
     void Start() override
@@ -69,6 +73,7 @@ private:
     const std::string Query_;
     const IClientPtr QueryClient_;
     const TSelectRowsOptions Options_;
+    const bool FailOnIncompleteInput_;
 
     TFuture<TSelectRowsResult> AsyncQueryResult_;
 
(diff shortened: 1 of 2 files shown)


yt/yt/server/query_tracker/ql_engine.cpp
```diff
@@ -81,7 +81,11 @@ class TQLQueryHandler
             OnQueryFailed(queryResultOrError);
             return;
         }
-        OnQueryCompleted({TRowset{.Rowset = queryResultOrError.Value().Rowset}});
+        const auto& queryResult = queryResultOrError.Value();
```
Krisha11, line 89: yt ql also should be limited by RowCountLimit
https://github.com/ulya-sidorina/ytsaurus/blob/6ff6d897e8c95894dceebb597e1266a6994b86bf/yt/yt/server/query_tracker/config.cpp#L22

ulya-sidorina: Done. QL now applies QT `RowCountLimit`. When this limit is reached, QL returns a partial result instead of failing, so QT can mark the result with `is_truncated = true`

yt/yt/server/query_tracker/ql_engine.cpp
Krisha11, line 95: What will be returned in case of `queryResult.Statistics.IncompleteInput = 1` and `FailOnIncompleteInput_ = 0`?

ulya-sidorina: Successful (but incomplete) result, this behavior is the same as before

Krisha11: Is the incompeteoutput also set in this case? Or is the result succesful but incorrect and IsTruncated is not set?

ulya-sidorina: No, `IncompleteInput` and `IncompleteOutput` are independent flags, 
`IncompleteInput`  - QL stopped reading/scanning input rows
`IncompleteOutput` - QL stopped writing output rows

So in this case the result will be successful but incorrect and `IsTruncated` is not set

ulya-sidorina: I think `IsTruncated` shouldn't depend on `IncompleteInput`, because `IncompleteInput = true`  means something different: QL stopped reading source rows before finishing the logical query

ulya-sidorina: `IsTruncated` currently means the result was cut by an output/result row limit

Krisha11: Well, I hope that users who use `FailOnIncompleteResult = fasle` will know what they are doing.

Krisha11 (approved):

Krisha11:
```
yt/yt/tests/integration/queries/test_ql.py:118: in test_result_truncation_by_query_tracker_row_count_limit
    q.track()
yt/yt/tests/library/yt_queries.py:42: in track
    raise YtError.from_dict(query["error"])
E   yt.common.YtError: Query 3dbffcab-e414fa0e-ec7ba0e-4d88ced9 failed
E       ORDER BY used without LIMIT
```

Krisha11:
_By AI_
Neither new test covers the FailOnIncompleteInput_ branch, which is the trickiest part of this change: with fail_on_incomplete_result left at its default, the Query Tracker row count limit now disables FailOnIncompleteResult in SelectRows, and the input failure has to be re-raised by hand. Please add a case with a small input_row_limit (and no fail_on_incomplete_result override) asserting the query fails with QueryInputRowCountLimitExceeded.

PR #1824, opened 2026-09-02
Fix crash at replicating s3_medium, config is mandatory

koct9i:


AleksandraZh:
Could you please add a test for that (somewhere around test_master_cell_addition)

koct9i:
> Could you please add a test for that (somewhere around test_master_cell_addition)

done

PR #1839, opened 2026-09-14
Validate DQ settings and add more logs.

Tony-Romanov:
Type: fix
Component: query-tracker

### DIFF
--- a/yt/yql/plugin/config.cpp
+++ b/yt/yql/plugin/config.cpp
@@ -3,8 +3,10 @@
 #include <yt/yt/client/security_client/public.h>
 
 #include <yt/yt/core/logging/config.h>
+#include <yt/yt/core/misc/error.h>
 #include <yt/yt/core/ytree/fluent.h>
 
+#include <util/generic/hash_set.h>
 #include <util/string/vector.h>
 
 namespace NYT::NYqlPlugin {
@@ -241,6 +243,76 @@ void TDQYTBackend::Register(TRegistrar registrar)
         .Default(false);
     registrar.Parameter("scheduling_tag_filter", &TThis::SchedulingTagFilter)
         .Default({});
+
+    registrar.Postprocessor([] (TThis* config) {
+        if (config->ClusterName.empty()) {
+            THROW_ERROR_EXCEPTION("DQ backend cluster_name must not be empty");
+        }
+        if (config->JobsPerOperation == 0) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: jobs_per_operation must be positive", config->ClusterName);
+        }
+        if (config->MaxJobs == 0) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: max_jobs must be positive", config->ClusterName);
+        }
+        if (config->MaxJobs < config->JobsPerOperation) {
+            THROW_ERROR_EXCEPTION(
+                "DQ backend %Qv: max_jobs (%v) must not be less than jobs_per_operation (%v)",
+                config->ClusterName,
+                config->MaxJobs,
+                config->JobsPerOperation);
+        }
+        if (config->MaxJobs % config->JobsPerOperation != 0) {
+            THROW_ERROR_EXCEPTION(
+                "DQ backend %Qv: max_jobs (%v) must be divisible by jobs_per_operation (%v)",
+                config->ClusterName,
+                config->MaxJobs,
+                config->JobsPerOperation);
+        }
+        if (config->VanillaJobLite.empty()) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: vanilla_job_lite must not be empty", config->ClusterName);
+        }
+        if (config->VanillaJobCommand.empty()) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: vanilla_job_command must not be empty", config->ClusterName);
+        }
+        if (config->Prefix.empty()) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: prefix must not be empty", config->ClusterName);
+        }
+        if (config->TokenFile.empty()) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: token_file must not be empty", config->ClusterName);
+        }
+        if (config->UploadReplicationFactor == 0) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: upload_replication_factor must be positive", config->ClusterName);
+        }
+        if (config->CpuLimit <= 0) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: cpu_limit must be positive", config->ClusterName);
+        }
+        if (config->MemoryLimit <= 0) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: memory_limit must be positive", config->ClusterName);
+        }
+        if (config->CacheSize < 0) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: cache_size must not be negative", config->ClusterName);
+        }
+        if (config->UseTmpFs && config->CacheSize > config->MemoryLimit) {
+            THROW_ERROR_EXCEPTION(
+                "DQ backend %Qv: cache_size (%v) must not exceed memory_limit (%v) when use_tmp_fs is enabled",
+                config->ClusterName,
+                config->CacheSize,
+                config->MemoryLimit);
+        }
+        if (config->WorkerCapacity <= 0) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: worker_capacity must be positive", config->ClusterName);
+        }
+
+        THashSet<TString> fileNames;
+        for (const auto& file : config->VanillaJobFiles) {
+            if (!fileNames.insert(file->Name).second) {
+                THROW_ERROR_EXCEPTION(
+                    "DQ backend %Qv: duplicate vanilla_job_file name %Qv",
+                    config->ClusterName,
+                    file->Name);
+            }
+        }
+    });
 }
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -259,6 +331,7 @@ void TDQYTCoordinator::Register(TRegistrar registrar)
         .Default();
     registrar.Parameter("debug_log_file", &TThis::DebugLogFile)
         .Default();
+
 }
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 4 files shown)


yt/yql/plugin/config.cpp
```diff
@@ -241,6 +243,76 @@ void TDQYTBackend::Register(TRegistrar registrar)
...
+
+        THashSet<TString> fileNames;
+        for (const auto& file : config->VanillaJobFiles) {
+            if (!fileNames.insert(file->Name).second) {
+                THROW_ERROR_EXCEPTION(
+                    "DQ backend %Qv: duplicate vanilla_job_file name %Qv",
+                    config->ClusterName,
+                    file->Name);
+            }
+        }
```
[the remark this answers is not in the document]
Tony-Romanov: Having the same file under different names is strange, but it is not an error.

uzhastik:
LGTM

yt/yql/plugin/config.cpp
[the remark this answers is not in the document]
lukyanovx: [P1] This is a real collision and also breaks `additional_system_libs`

The collision described above is not the same local file configured under different names. It is two different local files with the same basename. `TDqManager::Start` discards the configured `Name` and constructs each resource from `LocalPath`, while `TResourceFile::GetRemoteFileName` falls back to that path's basename:
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yql/plugin/native/dq_manager.cpp#L262-L265
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yql/providers/dq/actors/yt/resource_manager.h#L46-L52

There is also a deterministic process-plugin failure: the top-level postprocessor appends every `additional_system_libs` entry to `VanillaJobFiles` on every load:
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yql/plugin/config.cpp#L564-L584

`TProcessYqlPlugin::BuildProcessConfig` clones the template through `CloneYsonStruct`, whose default is to deserialize with postprocessing enabled, so those files are appended a second time:
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yql/plugin/process/plugin.cpp#L605-L608
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yt/core/ytree/yson_struct-inl.h#L378-L389

The generated child config is then written and passed back to the agent executable, where the duplicate validator rejects it during config loading:
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yql/plugin/process/plugin.cpp#L578-L590

Please preserve the configured remote name or validate effective remote names after the final file list is assembled, make the append idempotent, and add a regression test that clones and reloads a process-plugin config containing `additional_system_libs`.

yt/yql/plugin/config.cpp
```diff
@@ -385,6 +525,10 @@ void TYqlPluginConfig::Register(TRegistrar registrar)
         .DefaultNew();
 
     registrar.Postprocessor([=] (TThis* config) {
+        if (config->EnableDQ && config->DQManagerConfig->YTBackends.empty()) {
```
lukyanovx, line 528: [P1] Do not require native backends for QtWorker mode

This rejects `enable_dq = true` whenever the native backend list is empty, before the agent can select the plugin implementation. However, the public QtWorker startup path deliberately passes `startDqManager = EnableDQ && !UseQtWorkerYqlPlugin`, so QtWorker mode does not start the native DQ manager and does not need its backend list:
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yql/agent/yql_agent.cpp#L357-L365

The native plugin already checks for missing backends when `StartDqManager` is true:
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yql/plugin/native/plugin.cpp#L388-L399

Since `TYqlPluginConfig` cannot see `UseQtWorkerYqlPlugin`, this validation turns otherwise supported QtWorker configurations into startup failures. This configuration shape is used by existing QtWorker deployments, so the failure is not hypothetical. Please remove this top-level check, keep the requirement at the point where the native manager is actually enabled, and add a regression test that loads the QtWorker configuration shape with DQ enabled and no native backends.

yt/yql/plugin/config.cpp
```diff
@@ -241,6 +243,76 @@ void TDQYTBackend::Register(TRegistrar registrar)
...
+        if (config->Prefix.empty()) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: prefix must not be empty", config->ClusterName);
+        }
+        if (config->TokenFile.empty()) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: token_file must not be empty", config->ClusterName);
+        }
+        if (config->UploadReplicationFactor == 0) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: upload_replication_factor must be positive", config->ClusterName);
+        }
+        if (config->CpuLimit <= 0) {
```
lukyanovx, line 286: [P1] Preserve `cpu_limit = 0` compatibility

Before this change, `TDQYTBackend` accepted zero and the DQ resource manager only forwarded it to the operation spec:
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yql/providers/dq/actors/yt/yt_resource_manager.cpp#L1281-L1282

The downstream user-job validator explicitly accepts `cpu_limit >= 0`:
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yt/ytlib/scheduler/config.cpp#L1316-L1318

Zero is used by existing deployments, and nothing in this DQ path requires a strictly positive value. Thus `<= 0` turns working configurations into startup failures. Please reject only negative values and add a config-loading regression test for zero.

yt/yql/plugin/config.cpp
```diff
@@ -241,6 +243,76 @@ void TDQYTBackend::Register(TRegistrar registrar)
...
+        if (config->VanillaJobLite.empty()) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: vanilla_job_lite must not be empty", config->ClusterName);
+        }
+        if (config->VanillaJobCommand.empty()) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: vanilla_job_command must not be empty", config->ClusterName);
+        }
+        if (config->Prefix.empty()) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: prefix must not be empty", config->ClusterName);
+        }
+        if (config->TokenFile.empty()) {
```
lukyanovx, line 280: [P2] Preserve inline `token` as an alternative to `token_file`

Both public protobuf configs expose an inline token:
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yql/providers/dq/config/config.proto#L93-L103
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yql/providers/dq/config/config.proto#L141-L142

`TDqManager::Start` checks `HasToken()` before falling back to `TokenFile` for both the coordinator and backends:
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yql/plugin/native/dq_manager.cpp#L132-L139
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yql/plugin/native/dq_manager.cpp#L251-L260

Program configs preserve unrecognized fields recursively, so the proto-only `token` field reaches this conversion path today:
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yt/library/program/program_config_mixin.h#L177
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yt/library/program/program_config_mixin.h#L209-L218

The subsequent native-plugin conversion does not discard it: the native `TDqManagerConfig` stores the backends and coordinator as raw `INodePtr` values, so their complete YSON reaches protobuf conversion:
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yql/plugin/native/dq_manager.h#L29-L34
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yql/plugin/native/plugin.cpp#L387-L389

The new typed postprocessor only checks `TokenFile` and rejects the config before conversion. Please retain the existing `token`/`token_file` alternative, either by representing both in the typed config or by validating after protobuf conversion. The coordinator check at line 387 needs the same treatment. Please cover both backend and coordinator inline-token configurations with regression tests.

yt/yql/plugin/config.cpp
```diff
@@ -282,6 +355,73 @@ void TDQManagerConfig::Register(TRegistrar registrar)
...
+                WorkerNodeIdCount);
+        }
+        const ui32 nodesPerBackend = config->YTBackends.empty()
+            ? 0
+            : WorkerNodeIdCount / config->YTBackends.size();
+        for (const auto& backend : config->YTBackends) {
+            if (!clusterNames.insert(backend->ClusterName).second) {
+                THROW_ERROR_EXCEPTION("Duplicate DQ backend cluster_name %Qv", backend->ClusterName);
+            }
+            if (backend->MaxJobs > nodesPerBackend) {
```
lukyanovx, line 407: [P2] Validate `max_jobs` against an explicit node-ID range

The protobuf supports explicit `MinNodeId` and `MaxNodeId` fields:
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yql/providers/dq/config/config.proto#L110-L111

At runtime the range is split evenly only when either bound is absent; an explicit range is otherwise preserved:
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yql/plugin/native/dq_manager.cpp#L202-L229

These proto-only bounds are preserved by the program config loader:
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yt/library/program/program_config_mixin.h#L177-L218

The subsequent native-plugin conversion does not discard them: the native `TDqManagerConfig` stores each backend as a raw `INodePtr`, so its complete YSON reaches protobuf conversion:
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yql/plugin/native/dq_manager.h#L29-L34
https://github.com/ytsaurus/ytsaurus/blob/4aaaa93fb8f1ebb4ff53bd59d92dbcd2e8dde5b9/yt/yql/plugin/native/plugin.cpp#L387-L389

This check always compares `max_jobs` with the equal-split width, so with explicit bounds it can reject a valid configuration or accept one that does not fit the actual range. Please use the explicit width when both bounds are present, or validate after the effective range has been assigned. Add regression cases where the explicit width is both smaller and larger than the equal split.

yt/yql/plugin/config.cpp
```diff
@@ -241,6 +243,76 @@ void TDQYTBackend::Register(TRegistrar registrar)
...
+        if (config->CpuLimit <= 0) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: cpu_limit must be positive", config->ClusterName);
+        }
+        if (config->MemoryLimit <= 0) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: memory_limit must be positive", config->ClusterName);
+        }
+        if (config->CacheSize < 0) {
+            THROW_ERROR_EXCEPTION("DQ backend %Qv: cache_size must not be negative", config->ClusterName);
+        }
+        if (config->UseTmpFs && config->CacheSize > config->MemoryLimit) {
```
lukyanovx, line 295: [P2] Keep public operator-generated defaults compatible with a small memory limit

The public Kubernetes operator emits `memory_limit`, `cache_size`, and `use_tmp_fs` with `omitempty` semantics:
https://github.com/ytsaurus/ytsaurus-k8s-operator/blob/fbb8855aed57b1fb8290681fd0536abfe30d8bf7/api/v1/ytsaurus_types.go#L876-L898
https://github.com/ytsaurus/ytsaurus-k8s-operator/blob/fbb8855aed57b1fb8290681fd0536abfe30d8bf7/pkg/ytconfig/yql_agent.go#L99-L108
https://github.com/ytsaurus/ytsaurus-k8s-operator/blob/fbb8855aed57b1fb8290681fd0536abfe30d8bf7/pkg/ytconfig/yql_agent.go#L466-L511

If a CR sets only a memory limit below 6 GB, the generated YSON contains that smaller `memory_limit` but omits zero-valued `cache_size` and unset `use_tmp_fs`. This config then applies the plugin defaults (`cache_size = 6000000000`, `use_tmp_fs = true`) and is rejected here.

This validation therefore breaks a configuration shape currently generated by the public operator. Please preserve compatibility with it in this PR, for example by applying defaults consistent with the configured memory limit.

lukyanovx (changes requested):

yt/yql/plugin/config.cpp
Tony-Romanov: To me, this is very strange logic: instead of fixing the generation of a potentially dangerous configuration, we are preserving the ability to launch using it.

lukyanovx: You are right - fixing the operator so it does not generate this configuration is the proper long-term solution. My concern was only about breaking existing installations during the transition. Keeping the warning for compatibility and fixing the operator separately sounds reasonable to me. Thanks!

lukyanovx:
LGTM

uzhastik:
LGTM

yt/yql/plugin/config.cpp
```diff
@@ -282,6 +357,84 @@ void TDQManagerConfig::Register(TRegistrar registrar)
...
+        if (config->InterconnectPort == config->GrpcPort) {
+            THROW_ERROR_EXCEPTION("DQ manager interconnect_port and grpc_port must be different");
+        }
+        if (config->ActorThreads == 0) {
+            THROW_ERROR_EXCEPTION("DQ manager actor_threads must be positive");
+        }
+        if (!config->YTCoordinator) {
+            THROW_ERROR_EXCEPTION("DQ coordinator config must be specified");
+        }
+        if (config->YTCoordinator->ClusterName.empty() && config->YTCoordinator->ProxyAddress.empty()) {
```
uzhastik: у нас не внедрен ProxyAddress в конфигах. лучше эту проверку ослабить

yt/yql/plugin/config.cpp
```diff
@@ -210,8 +215,10 @@ void TDQYTBackend::Register(TRegistrar registrar)
         .Default("//sys/yql_agent/dq/data");
     registrar.Parameter("upload_replication_factor", &TThis::UploadReplicationFactor)
         .Default(7);
+    registrar.Parameter("token", &TThis::Token)
```
uzhastik, line 218: для чего вы используете поле token? обычно токены через файлы и env передаются

Tony-Romanov: Мы, конечно, его так не используем, а это просто фикc замечания от @lukyanovx выше.

lukyanovx: Я попросил сохранить его, чтобы не ломать обратную совместимость: раньше `token` не был typed-параметром, но неявно прокидывался дальше через `KeepRecursive`. Мы тоже inline token не используем, поэтому, думаю, его можно удалить и требовать `token_file`.

При этом одного удаления `registrar.Parameter("token", ...)` недостаточно: из-за `KeepRecursive` неизвестный `token` продолжит передаваться дальше, а при наличии обоих полей runtime сейчас выберет именно inline `token`. Поэтому, если хотим действительно убрать его поддержку, нужно также изменить `TDqManager::Start()` для backend и coordinator, чтобы он всегда загружал токен из `token_file`.

PR #1847, opened 2026-09-21
[yt/java] Fix HttpClient selector thread leak

### LINKED ISSUE #1845: `YTsaurusClient.close()` does not terminate `HttpClient-*-SelectorManager` threads created for RPC proxy discovery
## Problem

`YTsaurusClient.close()` does not release the JDK `HttpClient` used for RPC proxy discovery.

Creating and closing clients repeatedly causes one new live `HttpClient-*-SelectorManager` thread per client. These threads are retained as JVM GC roots, so they cannot be collected by GC and accumulate until the process is restarted.

This is especially problematic for applications that create short-lived `YTsaurusClient` instances, for example in a polling task.

## Environment

- Java: 21.0.3
- `tech.ytsaurus:ytsaurus-client:1.2.7`
- same implementation is still present in `1.2.17`

## Reproducer

The cluster must use HTTP-based RPC proxy discovery.

```kotlin
import tech.ytsaurus.client.YTsaurusClient

fun main() {
    val before = selectorManagerCount()
    println("Selector managers before: $before")

    repeat(20) { index ->
        YTsaurusClient.builder()
            .setCluster("<YT HTTP proxy or cluster name>")
            // .setAuth(...)
            .build()
            .use { client ->
                // Forces HTTP /api/v4/discover_proxies request.
                client.waitProxies().handle { _, _ -> null }.join()
            }

        println("After close #${index + 1}: ${selectorManagerCount()}")
    }
}

private fun selectorManagerCount(): Int =
    Thread.getAllStackTraces().keys.count {
        it.name.startsWith("HttpClient-") &&
            it.name.endsWith("-SelectorManager")
    }
```
Observed output:
```text
Selector managers before: 0
After close #1: 1
After close #2: 2
...
After close #20: 20
```

## Expected behaviour
After `YTsaurusClient.close()` , the HTTP proxy-discovery client should be shut down and its `HttpClient-*-SelectorManager` thread should terminate. The selector-manager count should return to its baseline instead of growing with every closed YTsaurus client.

## Implementation findings
`YTsaurusClient.close()` reaches `ClientPoolService.close()` through `ClientPoolProvider.close()`.
However, `ClientPoolService.close()` only:
1. marks the pool as stopped;
2. cancels the next scheduled proxy update;
3. closes objects in `toClose`.   

`toClose` is declared but is not populated. `HttpProxyGetter` holds the `HttpClient`, does not implement `AutoCloseable`, and is not closed by `ClientPoolService.close()`.
As a result, cancelling periodic discovery does not close the JDK `HttpClient` or terminate its selector-manager thread.

mmazurkevich:
Fixes #1845.

Close the HTTP client used for proxy discovery together with `ClientPoolService`.

On Java 21 and newer, the client is closed explicitly. On Java 11–20, the strong reference to the internal HTTP client facade is released, allowing the selector thread to terminate after garbage collection.

Tested on Java 17 and Java 21.


mmazurkevich:
@tinarsky Hi! I’ve opened this PR with a fix for #1845.
Could you please take a look when you have time?

